PR #2, opened 2021-04-16
Add the linkchecker from sphinx

lornajane:
Check the links. Uses the built-in link checker in the Sphinx makefile.


lornajane:
Failing because sphinx isn't installed or set up, maybe another action could be added? Either python + dependencies, or https://github.com/marketplace/actions/sphinx-build

PR #11, opened 2021-05-18
First release of basic Create/migrate/extensions doc for pg

ftisiot:
I created draft pages for PG
* Creation
* Migration
* Extensions

As of now those are taking the existing help article content and displaying a bit more nicely, would like to know if this is the good direction


ftisiot:
Currently this build is failing due a broken link pointing to https://aiven.github.io/terraform-provider-aiven/README . Should it point to https://github.com/aiven/terraform-provider-aiven/blob/master/README.md instead?

lornajane:
This is off to a good start! The postgres extensions page is especially useful, that's something we've long needed. Do you think we should link out to more information about each one or write a bit about what they are/do, in case someone doesn't know their extension by name?

The other material looks like it needs a bit more work before it's ready so I'm just commenting for now (and I'm writing up more instructions on linking, this is completely confusing I know!), let me know when I should take another look!

ftisiot:
I removed the rest of the content for now, focusing on the extension page. I agree we should link each of the extensions, will work on it now

lornajane (approved):
Thanks for kickstarting the postgres section! I moved a bunch of stuff around since I am still deciding on the structure but I think this gets us up and going :)

PR #13, opened 2021-05-25
Pin packages version on requirements.txt

### LINKED ISSUE #9: sphinxcontrib-mermaid broken on Sphinx >= 4.0
Hi there!

sphinxcontrib-mermaid is currently broken for Sphinx >= 4.0, [here is the PR fixing it](https://github.com/mgaitan/sphinxcontrib-mermaid/pull/71).

For now, trying to run `make livehtml` returns the following error:
```bash
Extension error:
Could not import extension sphinxcontrib.mermaid (exception: cannot import name 'ENOENT' from 'sphinx.util.osutil' (/Users/jonatasbaldin/aiven/devportal/venv/lib/python3.8/site-packages/sphinx/util/osutil.py))
Command exited with exit code: 2
The server will continue serving the build folder, but the contents being served are no longer in sync with the documentation sources. Please fix the cause of the error above or press Ctrl+C to stop the server.
```

And I got `404` on the index page on the live server,

Disabling out the `sphinxcontrib.mermaid` from `conf.py:extensions` fixes the issue temporarily.

We have a couple of options:
1. Disable it for now until the PR is merged.
2. Use an earlier version of Sphinx.

What do you think?

jonatasbaldin:
The title says it all ✨ 

Also fixes #9.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,6 @@
-furo
-sphinx
-sphinx-autobuild
-sphinx-panels
-sphinxcontrib-mermaid
-sphinx-external-toc
+furo==2021.4.11b34
+sphinx==3.5.4
+sphinx-autobuild==2021.3.14
+sphinx-panels==0.5.2
+sphinxcontrib-mermaid==0.6.3
+sphinx-external-toc==0.1.0


lornajane (approved):
Thanks! We could probably look at the sphinx upgrade at some point as well

jonatasbaldin:
Yes! Once things are fixed with mermaid, we can upgrade!

PR #14, opened 2021-06-01
feat: add dark mode styling

turq84:


gskhan:
@turq84 the theme check for the system preferences to load the light or dark mode theme, you already created the light mode styles, we now just need to find this setting to hook it up.

lornajane:
The theme itself will pick it up https://pradyunsg.me/furo/customisation/colors/?highlight=dark#how-light-and-dark-mode-work . The basic idea is that you set the CSS for light styles, and then override what's needed for dark. Most of this can be configured using the theme settings in `conf.py` rather than overriding in CSS.

From the original request (I'm not sure if this was passed through) "We're using the "Furo" theme which as you've already observed has support for dark mode built in.

The theme aims to be completely configurable from conf.py and allows loads of variables for light and dark modes (the light ones are always loaded, the dark ones are then conditionally loaded later so they overwrite). However I suspected that we'd want to add some custom CSS as well so I've created and configured the include for /static/css/aiven.css - including an example of the dark mode thing. I found a bunch of variables in https://github.com/pradyunsg/furo/blob/main/src/furo/assets/styles/variables/_index.scss

The little icons I'm using are "replacements" in the text, you can find them at the end of conf.py."

turq84:
I have updated the `conf.py` file with the color variables. Thank you, Lorna, for your help!

gskhan:
great job @turq84 
@lornajane please check and comment

lornajane:
Ohhh, this looks really good. The only thing I find strange is the highlighting on the fixed-width/code fonts in dark mode. For example this page https://deploy-preview-14--developer-aiven-io-preview.netlify.app/docs/tools/cli.html. If it's a design feature, that's fine :laughing:

turq84:
Thanks for your feedback @lornajane! Regarding the highlighting, do you mean the background color? The font colors? I took the background color from the Figma design I was given. For the fonts/font colors, I don't recall changing anything.

I assume that you're talking about the attaching coding block, for example:
Screenshot 2021-06-07 at 12 09 45

Would you like for me to change anything?

lornajane:
Ah, sorry, I wasn't clear. Not the code samples, but the literal text in the copy, when in dark mode. I've attached a screenshot, you can see where something is `fixed width` it's got a background colour on it that looks strange to me.
Screenshot_2021-06-07 Aiven CLI - Aiven Developer documentation

turq84:
@lornajane Thanks for your help and patience! I agree with you and I have updated the background color for the code snippets (dark mode). Here's an image of the update. Any thoughts?

Screenshot 2021-06-07 at 14 52 41

lornajane (approved):
This looks great! Thanks so much!

PR #15, opened 2021-06-01
devportal: Initial draft content for PostgreSQL

diawarr:
Content converted from existing help articles.


lornajane:
First impressions: this is a lot of content! Even with the restructure, I don't think that the content on the old site is very well suited to the intended audience of this portal and we will need to build quite a bit on this initial installment to create developer-centric resources - but the updates and especially content conversion is valuable!

As a first step, I'm going to mark this PR as draft, I think we'll need to build on this content but it may involve moving small sections into our desired content structure on an article-by-article basis so that the pull requests can be reviewed in a meaningful way and we can better address the needs of our technical audience.

lornajane:
This PR very much got us off the mark but it won't be merged, so I'll close it now.

PR #19, opened 2021-06-07
Added PG "getting started" page in line with M3 one

ftisiot:
Added a simple "getting started" page for PG


lornajane (approved):
Good starting point but let's keep building on this, I'll open some issues with suggestions.

PR #21, opened 2021-06-07
Connect psql, pgadmin, python pages

ftisiot:
Added pages for psql, pgadmin and python connections to PG


ftisiot:
Also added Go, based on https://github.com/aiven/aiven-examples/blob/master/postgresql/go/main.go

lornajane (approved):
This is in good shape, the code all works well (I did rename some stuff).

PR #38, opened 2021-06-17
Enhancements to PG Getting Started Page

ftisiot:
Initial Enhancement, this is just a draft, creating the PR for Anna Demo

### DIFF
--- a/docs/products/postgresql/getting-started.rst
+++ b/docs/products/postgresql/getting-started.rst
@@ -1,7 +1,7 @@
 Getting Started
 ===============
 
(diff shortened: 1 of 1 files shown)


lornajane:
Can you take a look at the build failure?

ftisiot:
looking at it now

lornajane (approved):
Nice! Especially the sample dataset. I think this should get people started pretty quickly.

lornajane:
@hab-anka Can you take a look at this change too please? In particular, would you add something after step 6? The user has a database, with data in it ... should there be a call to action there to get them to play with it or do something else if they want to?

PR #43, opened 2021-06-21
Update index.rst

hab-anka:
Rewriting the overview M3 page to align it more to the company website, and to propose a new general structure of the product pages.

### DIFF
--- a/docs/products/m3db/index.rst
+++ b/docs/products/m3db/index.rst
@@ -1,18 +1,27 @@
-M3DB Welcome
+Aiven for M3
 ============
 
-M3DB is an open source, **distributed time series database**. It's ideal for organisations that have a very large volume of metrics to handle. This platform was originally developed at Uber, so that gives an idea of the scale of metrics that it can handle.
+Aiven for M3 is a fully managed **distributed time series database**, deployable in the cloud of your choice which can bring unlimited scalability and high-availability to your Prometheus monitoring environment and other time series applications.
 
-M3DB plays nicely with other tools that might be in the stack already:
+Why M3?
+-------
+M3 is a great choice if your organization has a very large volume of metrics to handle, and it can be used as part of your observability solution. 
 
-* Collect metrics with Prometheus; M3DB is designed as a scalable storage backend for Prometheus.
-* Write metrics direct from your applications using the InfluxDB-compatible wire protocol.
-* Dashboards and query features available using Grafana and PromQL.
+M3 is highly compatible with other Aiven products for the following tasks:
+
+* To collect metrics with Prometheus: M3DB is designed as a scalable storage backend for Prometheus.
+* To write metrics directly from your applications using the InfluxDB-compatible wire protocol.
+* To create dashboards and query available features using Grafana and PromQL.
+
+Check out all the features on our `M3 product page <https://aiven.io/m3#full-feature-list/>`_. 
+
+.. tip::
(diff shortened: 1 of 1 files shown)


docs/products/m3db/index.rst
```diff
@@ -1,18 +1,27 @@
...
-* Collect metrics with Prometheus; M3DB is designed as a scalable storage backend for Prometheus.
-* Write metrics direct from your applications using the InfluxDB-compatible wire protocol.
-* Dashboards and query features available using Grafana and PromQL.
+M3 is highly compatible with other Aiven products for the following tasks:
+
+* To collect metrics with Prometheus: M3DB is designed as a scalable storage backend for Prometheus.
+* To write metrics directly from your applications using the InfluxDB-compatible wire protocol.
+* To create dashboards and query available features using Grafana and PromQL.
+
+Check out all the features on our `M3 product page <https://aiven.io/m3#full-feature-list/>`_.
```
lornajane, line 16: The trailing slash at the end of the link is causing the linkcheck task to fail. It's only that there shouldn't be a `/` after `list` in the link there. From the screenshot you sent me, you can scroll up and see the one labelled "broken" - but these are tools that we can also set up on your laptop to check things as you go along and hopefully make things easier.

I'm not sure about linking out to a marketing page here, it would be less disruptive to list the M3 features instead. The "Platform features" aren't specific to M3 and aren't relevant on this page.

docs/products/m3db/index.rst
```diff
@@ -1,18 +1,27 @@
...
+Check out all the features on our `M3 product page <https://aiven.io/m3#full-feature-list/>`_. 
+
+.. tip::
+   Aiven offers M3 because we ourselves needed a solution that would work with the size of our own metrics - and we love it!
 
 First Steps
 -----------
 
-Check out our :doc:`getting-started` guide for M3DB and take your first steps right away.
+Check out our :doc:`getting-started` guide for M3 and take your first steps right away.
```
lornajane, line 24: This made me realise that my use of M3 vs M3DB is all over the place! M3 is the overall project, M3DB is the actual database. There's a bit more on the components on their own docs: https://m3db.io/docs/overview/components/ which might or might not help work out what to do here.

lornajane (commented):
Just a comment for now as I'm not quite ready to merge. Take a look at the small notes I added, as I think both need more discussion ( you can reply to the comments so we keep the info together).

Also important for future pull requests:
 * Please use a meaningful branch name that describes what is in the branch.
 * Commit messages should give an indication of what the change achieves, and should not be a duplicate of an existing commit message.

docs/products/m3db/index.rst
hab-anka: I linked to the full features for M3 - as listed now on the site. I guess this is now the only place where we can link.

docs/products/m3db/index.rst
hab-anka: I think we should specify what the main product name actually is, and in which cases should we use what. I see M3 and M3DB used interchangeably.

docs/products/m3db/index.rst
hab-anka: Fixed the link (hopefully).

hab-anka:
Hi @lornajane I have now fixed the link, and included the official name of Aiven for M3.

lornajane (commented):
Thanks!

PR #45, opened 2021-06-21
Update migrate-pg-dump-restore.rst

hab-anka:
Rewriting the migrating topic to make it consistent with the other alternative.


lornajane (approved):
This is good, thanks! I tweaked the titles a bit to keep them matching the others but we can revisit those if you'd like to.

PR #46, opened 2021-06-21
Remove mis-merged system files

lornajane:
There are some files that should not be here! Remove them and update the gitignore file.


ftisiot (approved):
Good to be merged, probably my fault!

lornajane:
nope, it was actually me merging the style change and not checking the diff closely enough!!

PR #48, opened 2021-06-23
High Availability and ugrades

ftisiot:
Added few pages about:
* Terminology
* High Availability for PostgreSQL
* HowTo perform and Upgrade
* Upgrade and failover procedures
(I added 4 files instead of the usual 3 bc the terminology is a small one :D )


lornajane (approved):
Thanks, this is really key info :)

PR #49, opened 2021-06-24
DBA type of tasks with Aiven for PostgreSQL

ftisiot:
This branch contains two articles:
- concept: Perform DBA-type tasks in Aiven for PostgreSQL
- howto: Claim public schema ownership


lornajane (approved):
The coolest thing about this is how much I learn when I have to read all the docs :)

PR #52, opened 2021-06-25
Remove placeholder pages, tidy main page

lornajane:
Removed the pages that don't need to be there before launch. Also try to do something more useful with the landing page.


lornajane:
This HowTo page is a great improvement. Can it work on the other table of contents pages too, like the Concepts pages? https://deploy-preview-52--developer-aiven-io-preview.netlify.app/docs/products/postgresql/concepts.html There might also be other places where the table of contents is used.

docs/tools/index.rst
```diff
@@ -1,12 +1,13 @@
-Tools for Aiven
-===============
+Aiven Tools
```
ftisiot, line 1: This is fine for the beta, however I would add some "squares" like the home page to make options more visible

ftisiot: One square for each tool

index.rst
```diff
@@ -1,70 +1,87 @@
...
+Explore
+-------
 
-Relational databases that you know and love ❤️
+Learn about the Aiven platform
 
 .. panels::
+    :card: shadow
+
+    |icon-postgres| **PostgreSQL** Powerful relational database platform. We have the latest versions, and all the extensions you need.
```
ftisiot, line 28: "all extensions you need" <- isn't this a bit over-promising?

index.rst
```diff
@@ -1,70 +1,87 @@
...
 
-    |icon-kafka| **Apache Kafka** For data streaming services, look no further than Apache Kafka.
+    ---
+
+    |icon-kafka| **Apache Kafka** Streaming data pipelines for the modern enterprise. Apache MirrorMaker2 and Kafka Connect also available.
 
     ---
 
-    |icon-kafka-connect| **Kafka Connect** To connect all the things into (source) and out of (sink) Apache Kafka.
+    |icon-elasticsearch| **ElasticSearch** Document database with specialist search features, bring your freeform ddocuments, logs or metrics, and make sense of them here.
```
ftisiot, line 54: ddocuments <- bug

index.rst
```diff
@@ -1,70 +1,87 @@
...
 
     ---
 
-    |icon-kafka-mirrormaker| **MirrorMaker2** Everything from one Apache Kafka to another.
+    |icon-cassandra| **Cassandra** High performance storage solution for large data quantities. This specialist data solution is a partitioned row store.
```
ftisiot, line 58: "partitioned row store" I believe could be confusing? should we add NoSQL somewhere?

ftisiot (changes requested):
Good shape, minor bugs to solve and I would add the various Aiven tools as squares like we do in the main home page for products

index.rst
lornajane: it isn't NoSQL, and its own website describes it that way. I thought it was a columnar database but apparently not!

index.rst
lornajane: good spot

index.rst
lornajane: I liked it! But you are probably correct, I'll tone it down.

docs/tools/index.rst
lornajane: I like this idea but I'm reluctant to delay this PR for it. Let's add it separately!

ftisiot (approved):
All good now

PR #55, opened 2021-06-28
PG Advanced Parameters

ftisiot:
Initial Draft of the advanced parameters page.  **Not meant to be complete yet**
Would like some input on the [parameters page](https://github.com/aiven/devportal/blob/pg_parameters/docs/products/postgresql/reference/list-of-advanced-params.rst)
a) general structure of the page
b) columns added on tables
c) Sections order

Also, we should start a discussion on how to keep this in sync... I amended several parameter descriptions since they were on different styles.

### DIFF
--- a/_toc.yml
+++ b/_toc.yml
@@ -29,6 +29,7 @@ entries:
         title: HowTo
         entries:
           - glob: docs/products/postgresql/howto/*
+      - file: docs/products/postgresql/reference/list-of-advanced-params
       - file: docs/products/postgresql/reference/list-of-extensions
       - file: docs/products/postgresql/reference/pg-connection-limits
       - file: docs/products/postgresql/reference/terminology
--- a/docs/products/postgresql/reference/list-of-advanced-params.rst
+++ b/docs/products/postgresql/reference/list-of-advanced-params.rst
@@ -1,0 +1,500 @@
+List of Advanced Parameters
+============================
+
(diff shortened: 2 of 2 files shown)


lornajane:
@ftisiot Should we suggest the updated parameter descriptions be passed back upstream? I can probably help make that happen if you think it would be valuable.

lornajane:
I don't think the table format works well, because the description is quite long and it ends up in a narrow column. We probably need something more like how the API documentation displays things https://api.aiven.io/doc/#tag/Project but for a static site I'm not sure of the best way to handle that!

ftisiot:
Lorna, yes that's my idea, let's fix descriptions and then pass them back to the `space invaders` team. As of now descriptions are not consistent. 
Once we solve this issue, we can tackle the next of not having consistency across field naming in the JSON export

hab-anka:
> I don't think the table format works well, because the description is quite long and it ends up in a narrow column. We probably need something more like how the API documentation displays things https://api.aiven.io/doc/#tag/Project but for a static site I'm not sure of the best way to handle that!

I think the table works well here - this is the classic display for such type of content. 
We can think of a list, but that would mean, we need to work on better formatting and organization.

hab-anka (approved):
Looks OK Francesco. Thanks for helping on the formatting in the mean time :)

PR #58, opened 2021-06-30
Add templates for pull requests and issues

lornajane:
Let's make things clearer for contributors


.github/ISSUE_TEMPLATE/question.md
```diff
@@ -0,0 +1,11 @@
+---
+name: ❓ Ask a question
+about: Got stuck or missing something from the docs? Ask away!
+---
+
+# What can we help you with?
+
+(for urgent or production platform issues, please contact the support team as normal)
```
ftisiot, line 8: Should we add "via aiven.io" with a link to send people in the right direction?

ftisiot (changes requested):
Adding a link in case of needing support could be useful.
Are the `request.md` and `review.md` swapped? 
e.g. shouldn't I use the `request` to ask for some content and the `review` to tell something is wrong?

lornajane:
Some excellent points! Can you take another look please?

ftisiot (approved):
Seems all correct

PR #60, opened 2021-06-30
Moved default value, valid range to main description

hab-anka:
And changed the capitalization of the title.


ftisiot (approved):
Good work!

PR #63, opened 2021-07-01
Changed to sentence style

hab-anka:


lornajane (approved):
Thanks for taking the time to revisit these, Going to do an eleventh-hour merge because it makes the nav entries cleaner.

Pull request format feedback: since this is a single task (correcting title case), I would recommend that you make all the changes, and then stage all the changes and commit once. This also avoids the need for repeated commit messages.

PR #66, opened 2021-07-09
Move postgres reference items into a group, fix CSS to support this

lornajane:

Now that we have a number of PostgreSQL reference articles, the nav is getting harder to manage. Nest the reference in its own section, and make the CSS styling on the listing pages (Concept, HowTo) more generic to cover this page too.


gskhan (approved):
👍

PR #67, opened 2021-07-09
Fix README link documentation syntax error

lornajane:

When following the instructions in the README for internal linking, I found an error. This correctly adds the colon after the label name.


hab-anka (approved):
Thanks for spotting.

PR #69, opened 2021-07-11
Add a rule to check headings

### LINKED ISSUE #62: Add a Vale rule to check sentence case on titles
Ensure that new content uses sentence case for titles by adding a vale configuration

lornajane:

We agreed to standardise on sentence case for headings, so let's enforce this in our automated style rules. Fixes #62.


lornajane:
Not ready for review because sentence case doesn't do what I expected it to do. For example, it produces an error for "PostgreSQL resources" which I thought was correct sentence case.

lornajane:
Sentence case means "first letter is upper case, everything else is lowercase" but it does support words that are specifically in the accept list. However this does mean that all the proper nouns (e.g. Python, Terraform) and the things that are spelt with capital letters at random points (PostgreSQL, pgAdmin) also need to be accepted words.

hab-anka (approved):
Looks good!

PR #72, opened 2021-07-19
Tiny fix for a formatting problem on M3 main page

lornajane:

I realised we had mis-formatted the links in the M3 landing page.


hab-anka (approved):
Thanks for spotting.

PR #123, opened 2021-07-22
Fix a label that somehow went missing in an earlier refactor

lornajane:

It's only generating a warning but I think it's worth fixing.


ftisiot (approved):
I was looking at the same now! Glad you already fixed it!

lornajane:
If you approve, and would feel confident to merge, please go ahead and merge.

PR #124, opened 2021-07-22
Move to a nested structure for PostgreSQL HowTo content

lornajane:

We have tooooooooo many items in the postgres/howto section of the navigation. This PR adds some structure to this and splits things up a bit - also adds some additional code samples (these are in their own commit and could be merged separately if that would be better).

### DIFF
--- a/.github/styles/Vocab/Docs/accept.txt
+++ b/.github/styles/Vocab/Docs/accept.txt
@@ -15,6 +15,7 @@ hardcoded
 [Hh]ostname
 jq
 [Nn]amespace
+npm
 parallelization
 pgAdmin
 PgBouncer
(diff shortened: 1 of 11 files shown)


hab-anka (changes requested):
Nesting looks good. The new topics and TOC entries use Headline Style though... Not sure if the overwrite is working.

lornajane:
Oops! We didn't review and merge the change that checks the headers so the functionality isn't available in other branches yet. It's #69, let's make this PR wait until that one is merged and I can make sure everything is as it should be.

lornajane:
Thanks for catching this, rebased to include the additional check, and updated some titles/vocab. Could you take another look?

hab-anka (approved):
Looks good!

Issue #77, opened 2021-07-20
M3 overview


lornajane:
Add initial product content following existing templates/conventions

PR #138, opened 2021-07-26
Reorganized structure and filled in examples to use Aiven for M3

hab-anka:

- Reorganized and reworded structure of the overview (as suggested here: https://docs.google.com/document/d/10rKMD3Y-eSU55FtxH53LzONbAgm5UdwQP3XXbX8SxcU/edit)
- Added concrete examples of implementations
- Standardized the naming (product name Aiven for M3).

### DIFF
--- a/docs/products/m3db/index.rst
+++ b/docs/products/m3db/index.rst
@@ -26,7 +26,7 @@ Handle and analyse the time-stamped data from multiple connected devices and ser
 
 With Aiven for M3, you can set up the following example solutions:
 
-- Monitor IoT deployments, and use time series analysis to identify seasonal patterns, average usage and potential inefficiences.
+- Monitor IoT deployments, and use time series analysis to identify seasonal patterns, average usage and potential inefficiencies.
 
 - Measure the performance of your applications, detect problems, respond to incidents and plan maintenance. 
 
@@ -50,7 +50,7 @@ M3 is highly compatible with other Aiven products for the following tasks:
 Check out all the features on our `M3 product page <https://aiven.io/m3#full-feature-list>`_. 
 
 
-Get strated with Aiven for M3
+Get started with Aiven for M3
 -----------------------------
 
 Take your first steps with Aiven for M3 by following our :doc:`getting-started` article, or browse through our full list of articles:
--- a/docs/products/m3db/index.rst
+++ b/docs/products/m3db/index.rst
@@ -41,15 +41,13 @@ Handle and analyse the time-stamped data from multiple connected devices and ser
 
 With Aiven for M3, you can set up the following example solutions:
 
-- Monitor IoT deployments, and use time series analysis to identify seasonal patterns, average usage and potential inefficiencies.
+- Monitor IoT deployments, applications performance, financial trends, and even self-driving cars.
 
-- Measure the performance of your applications, detect problems, respond to incidents and plan maintenance. 
+- Detect problems, respond promptly to incidents and plan maintenance.
 
-- Cross-reference contextual data to understand financial trends. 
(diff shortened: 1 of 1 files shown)


docs/products/m3db/index.rst
```diff
@@ -33,12 +35,43 @@ Check out our :doc:`getting-started` guide for Aiven for M3 and take your first
     💻 :doc:`howto`
 
 
-M3DB resources
---------------
+Ways to use Aiven for M3
```
lornajane, line 38: This section seems out of place in developer documentation, let's focus on technical capabilities and integrations, but the use cases work better on the marketing site (where we hopefully already have this information)

docs/products/m3db/index.rst
```diff
@@ -33,12 +35,43 @@ Check out our :doc:`getting-started` guide for Aiven for M3 and take your first
...
+
+- Cross-reference contextual data to understand financial trends. 
+
+- Process self-driving car data with faster data ingest and queries, and with stronger data compression. 
+
+- Predict future sales by using statistical models with historic data and cross-referencing it with consumer trends. 
+
+
+
+Works best with other Aiven products
```
lornajane, line 56: Grafana is the only Aiven product, we don't offer Prometheus - and we do offer InfluxDB but M3 replaces it, you wouldn't probably use them together. How about calling this section "Plays nicely with others" or "Integrates with your existing tools" or something like that?

lornajane (commented):
Really minor suggestions here, I think this is pretty close to being ready!

docs/products/m3db/index.rst
hab-anka: I'd like to leave it, and see how it works, this is actually not on the marketing pages, and I am not sure if it's easily understood what people can do with these products.

docs/products/m3db/index.rst
hab-anka: Good suggestion, I will reword.

docs/products/m3db/index.rst
lornajane: If the goal of this section is to explain what people can do, that does seem useful, but it currently misses that goal and instead strays into marketing-to-developers territory which we would want to avoid. Please refactor the language to give the "When to choose M3" information in terms of technical problems.

ftisiot: I would agree with @lornajane, as of now the section seems too much marketing oriented (e.g. "Process **self-driving car data** with faster data ingest and queries, and with stronger data compression."). 

what about "What can you do with M3" section rephrasing the points contained in the section like
* Monitor IoT deployments, applications performance, financial trends and even self-driving cars
* Detect problems, respond promptly to incidents and plan maintenance
* Provide fast data ingest and queries, with strong data compression
* Build statistical models on the historic data to predict future trends

It contains more or less the same info, but with the eye of the developer

hab-anka: Thanks both, and thanks @ftisiot for pointing me to the right wording (perspective).

lornajane: Could you swap this bit with the section below about the integrations, and remove the reference to self-driving cars, please. That is a fun example that we could use in a tutorial but not a mainstream or technical application, and the marketing speak undermines the trustworthiness of documentation.

I'm also not sure that we can build statistical models to predict the future with M3 ... @ftisiot would we integrate with other tools for predictions, is that what you had in mind? Let's avoid over-promising in developer documentation if that feature doesn't exist unless we already have a howto on achieving it with an integration.

lornajane:
Fixes #77

lornajane (approved):
Thanks! This is ready to go

PR #142, opened 2021-07-29
Add an importer script for help articles

lornajane:
* Install pandoc https://pandoc.org/
* Run this script from the command line, with the URL of the page to import

  python import-help-articles.py https://help.aiven.io/en/articles/5074009-forking-an-aiven-service

### DIFF
--- a/utils/import-help-article.py
+++ b/utils/import-help-article.py
@@ -1,3 +1,4 @@
+import argparse
 import os
 import requests
 import sys
@@ -6,10 +7,12 @@ from urllib.parse import urlparse
 import pypandoc
 from bs4 import BeautifulSoup
 
-if len(sys.argv) < 2:
-    sys.exit("Give the URL with the command")
+parser = argparse.ArgumentParser(description="Convert an existing help file")
+parser.add_argument('url', help="The URL of the source file")
+args = parser.parse_args()
 
-page_url = sys.argv[1]
+page_url = args.url
+print(page_url)
 
 # tangent: grab a filename
 url_pieces = urlparse(page_url)
--- a/utils/import-help-article.py
+++ b/utils/import-help-article.py
@@ -1,0 +1,59 @@
+import os
+import requests
+import sys
+from urllib.parse import urlparse
+
+import pypandoc
+from bs4 import BeautifulSoup
+
+if len(sys.argv) < 2:
+    sys.exit("Give the URL with the command")
+
+page_url = sys.argv[1]
+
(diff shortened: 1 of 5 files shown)


utils/import-help-article.py
```diff
@@ -0,0 +1,59 @@
...
+import sys
+from urllib.parse import urlparse
+
+import pypandoc
+from bs4 import BeautifulSoup
+
+if len(sys.argv) < 2:
+    sys.exit("Give the URL with the command")
+
+page_url = sys.argv[1]
```
SmuliS, line 12: Not requesting any changes here, so this is more of a FYI. I find that using [argparse](https://docs.python.org/3/library/argparse.html) makes building Python CLI tools super easy and more human friendly. Just wanted to share my thought here :)

diawarr (approved):
Works fine for me :)

One thing to note is that the pandoc conversion adds pipe characters at least to the start of some list items - this doesn't seem to affect the built output, but just something to keep in mind in case it has any impact at some stage ;)

lornajane:
Can you give an example of where you see the additional character, @diawarr ? I guess we will need a few more HTML transformations as we run into things.

diawarr:
I tested with https://help.aiven.io/en/articles/2843308-migrating-elasticsearch-data-to-aiven and https://help.aiven.io/en/articles/5241419-configuring-java-ssl-to-access-kafka - my guess is that it adds the pipe if there are any subelements within the main <li><p class="no-margin> element, even if it's just a link or a <br/> at the end of the content ;)

utils/import-help-article.py
lornajane: :+1: Thanks for the tip! Take a look at the updated version if you have a minute

lornajane:
@diawarr Do you think the additional pipe character is a showstopper? I believe it's actually the very "correct" behaviour when downconverting from a richer format (html) into something more simple (rst) or presenting any multiline bullet points.

diawarr:
@lornajane - no, I don't see it having any impact on the built output, so there's no need to fix it as such, it was more something that would be worth noting for anyone editing the files, as most of the converted files will need at least some manual editing anyway ;)

lornajane:
Totally agree that most of the content needs rewriting anyway so probably not much will survive unchanged when it's fitted into our preferred content format. Glad to hear that you don't think it's a big problem, I've pushed the docs now so I'll convert to a proper pull request and request your review now all the pieces are there.

diawarr (approved):

PR #150, opened 2021-08-03
Add workflows to index pages

aiven-hh:
Add workflows to index devportal pages on changes to main and help center pages daily on schedule. Requires `ES_URL` to be added to repository secrets.

### DIFF
--- a/.github/workflows/index-devportal.yaml
+++ b/.github/workflows/index-devportal.yaml
@@ -1,0 +1,26 @@
+name: Index devportal pages
+
+on:
+  push:
+    branches: [ main ]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python 3.6
+      uses: actions/setup-python@v2
+      with:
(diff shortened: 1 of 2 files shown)


gskhan:
@lornajane can you please add the secret? Most important, we are still not using prod ES but the sandbox

lornajane:
Quick update from security that we are OK to add the repo secrets for this, I'll proceed.

lornajane:
How can I see this workflow running? I've added the environment variable as a repo secret and for the devportal deploy - but at least the devportal version won't/shouldn't index unless we're pushing to main. Do I test by adding the branch name of this branch temporarily to see it build? And why aren't the cron workflows running?

aiven-hh:
Also I couldn't really test these myself either but I tried to keep them as simple as possible and hope for the best. They just install deps and run single make command each.

lornajane:
I created the index and pushed a temporary commit to cause the actions to run on this branch and check things work as we expected. These completed successfully so I think we can move forward with this.

lornajane (approved):
Triggering the actions does produce content in the index

PR #151, opened 2021-08-03
postgresql: mention how to change backup time

sigmaris:
I've seen a few support cases with customers asking how to change the time of day the daily backups happen, and I think it's worth mentioning how to do this in the docs. I wasn't totally sure if it was better to put this here or in the howto / DBA tasks section, though.


lornajane (approved):
This is a great tip! Thanks :)

Issue #74, opened 2021-07-19
Add a feedback mechanism


lornajane:
We need to collect feedback from users on whether they found the content helpful or not. This should be in the form of a :+1: and :-1:  button pair, and an optional comment. Users need to understand that we can't identify them on devportal so they need to contact support if they need a response (or open a GitHub issue).

Each data point should have:
 - the URL of the page that the form was on
 - whether the user chose :+1: or :-1:
 - their comment, if any
 - the date and time that the data was submitted

## Tech Info

Implementation details are open for discussion but my suggestion:
 - netlify form
 - netlify function (can also inspect this data via netlify console and/or integrate with Slack easily)
 - Aiven postgresql, in the same workspace as our postgresql

## Example implementations

- https://docs.newrelic.com/docs/apis/intro-apis/introduction-new-relic-apis/

## Next steps

A follow-on issue will be needed for reporting on the collected data - create it as the final step in this task once we know how this issue chose to store data.

PR #152, opened 2021-08-04
Add feedback form and store results in pg

aiven-hh:
Setup:

* Get service url from Aiven console without `sslmode=require`, example: `postgres://avnadmin:pass@service-name-project-name.aivencloud.com:12691/defaultdb`
* Add service url to environment with key `PG_URL` (for Netlify also)
* Download `ca.pem`CA certificate from Aiven console and save it as `netlify/functions/ca.pem`, this is currently the CA cert for dev instance
* Run `make create-feedback-table` to create the table in PG.

Implements #74


hab-anka:
Can we please change the default text? These sound a bit formal and "document" is not super accurate ....

Did you find this document useful? > Did you find this useful?

What did you think about the document? > Anything specific you want to share?

Thank you for your feedback. > Thanks for your opinion!

lornajane:
In the preview setup, I can see that the forms work, but the functions aren't in place to fire yet so the data does not arrive in the database. Could you please add:
 - pick up the URL of the page and pass it with the form? This would let us build other integrations straight off the forms data/trigger, not just from the function/database end of things. And I guess this data is available when clicking.
 - add/enable (or explain to me if it's there and I didn't see it) any anti-spam measures that we can use here.

lornajane:
Note that the database URL has been added to both the netlify environment variables and to the github action secrets (why does GitHub need it?), and that the cert has been updated in the repo to match.

lornajane (changes requested):
Need to pick up the page URL earlier in the process, and add some anti-spam (see more detailed comment on the PR). Thanks for already updating the wording we requested :)

aiven-hh:
Github does not need the secret, my comment was a bit unclear. I meant that wherever you run the `make create-feedback-table` should have the variable set (locally).

Netlify has Akismet filtering on by default (based on the input text) but do you want additional filtering like honeypot field or recaptcha (https://docs.netlify.com/forms/spam-filters/).

You can see the page url in netlify submissions already if you use the download as CSV functionality. The web ui does not show it by default.

lornajane:
A honeypot would seem like a good idea (I'm not an expert though), let's add one.

lornajane:
Thanks for the update on the postgres settings - I created the table and removed the secret from the repo since we don't need it over here. It's all starting to make sense :)

aiven-hh:
I have added the honeypot field and netlify seemed to notice it saying `Extra spam prevention enabled via honeypot field.` in form submissions.

Also returning the page url. I don't think url needs to be added to the form. Netlify takes it from referrer header and adds it to the data passed to the function and I would guess to any other integrations as it is available in netlify form submissions. If it is not available to other integrations for some reason then I would add it.

lornajane:
This is looking really good, thanks for advising and adding the honeypot. I don't know if it will help but at least we tried.

We will be reviewing the comments from time to time, but we can't identify the user or respond quickly, and I would like to make that clear from the form. After the text "Anything specific you want to share?" could we therefore add "(Please don't share any personal details here. If you need a response, support@aiven.io is the best way to reach us)" to try to help users understand what they're interacting with.

lornajane:
@aiven-hh any danger you can update the wording on this? I'm keen to have it in action!

aiven-hh:
Sorry I didn't notice the previous comment. I added the text with the same font right after the old one, tell me if it needs any changes.

lornajane (approved):
This looks good to me, thanks very much!

PR #155, opened 2021-08-05
`avn cloud` documentation 

ftisiot:
Added the file `docs/tools/cli/cloud.rst` documenting the `avn cloud`.
Pretty simple and based on what @lornajane did for `avn project`

### DIFF
--- a/docs/tools/cli.rst
+++ b/docs/tools/cli.rst
@@ -63,6 +63,8 @@ Manage the payment cards on your account.
 
 Use ``avn cloud list`` to see all the currently-available clouds on Aiven. This is useful for looking up the cloud name to use with ``service`` commands.
 
+:doc:`See detailed command information <cli/cloud>`.
+
 
 ``credits``
 '''''''''''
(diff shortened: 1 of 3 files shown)


docs/tools/cli.rst
```diff
@@ -63,6 +63,8 @@ Manage the payment cards on your account.
```
hab-anka, line 63: Could we also make the `cloud `heading linked?

hab-anka (approved):
A few small suggestions (link, short desc and punctuation) - some in a commit.

docs/tools/cli.rst
ftisiot: If we do so we have to change all the titles from being code like ``cloud`` to cloud. @lornajane what do you think?

ftisiot: The drawback if we don't link from the title is that we'll need to add aSee detailed command information .... for each command.. that could become quite repetitive

lornajane: Let's leave it as it is. Linking titles isn't a common UX pattern (or the links would be internal ones to let you link directly to the section, not to somewhere else) and I think people should read the high level information and then decide if they need more info.

lornajane (approved):
Small but mighty!

PR #157, opened 2021-08-06
Added initial content for OpenSearch access control

diawarr:
Content converted from Intercom, restructured and edited.

### DIFF
--- a/docs/products/opensearch/concepts/access_control.rst
+++ b/docs/products/opensearch/concepts/access_control.rst
@@ -1,7 +1,7 @@
 Access control
 ==============
 
-Aiven for OpenSearch supports index-level access control lists (ACL) to control permissions.
+Aiven for OpenSearch supports index-level access control lists (ACL) to control permissions. This approach allows you to limit the operations that are available to specific connections and to restrict access to certain data sets, which improves the security of your data.
 
 You can grant the following permissions:
 
@@ -14,6 +14,9 @@ You can grant the following permissions:
 Rules are defined separately for each user as ``pattern/permission`` combinations. The ``pattern`` defines the indexes that the permission applies to. Patterns are glob-style, where ``*`` matches any number of characters and ``?`` matches any character. 
 
 When multiple rules match, they are applied in the order listed above. If no rules match, access is denied.
+
+Example
+-------
 
 As an example, we can use the following set of rules:
 
--- a/docs/products/opensearch/concepts/access_control.rst
+++ b/docs/products/opensearch/concepts/access_control.rst
@@ -1,0 +1,81 @@
+Access control
+==============
+
+Aiven for OpenSearch supports index-level access control lists (ACL) to control permissions.
+
+You can grant the following permissions:
+
+* ``deny``: no access
+* ``admin``: full access to APIs and documents 
+* ``readwrite``: full access to documents
+* ``read``: allow only searching and retrieving documents
+* ``write``: allow updating, adding, and deleting documents
+
+Rules are defined separately for each user as ``pattern/permission`` combinations. The ``pattern`` defines the indexes that the permission applies to. Patterns are glob-style, where ``*`` matches any number of characters and ``?`` matches any character. 
+
+When multiple rules match, they are applied in the order listed above. If no rules match, access is denied.
+
+As an example, we can use the following set of rules:
+
+* ``logs_*/read``
+* ``events_*/write``
+* ``logs_2018*/deny``
+* ``logs_201901*/read``
+* ``logs_2019*/admin``
+
+This set would allow the service user to
+
+* add documents to ``events_2018`` (second rule),
+* retrieve and search documents from ``logs_20171230`` (first rule),
+* gain full access to ``logs_20190201`` (fifth rule), and
+* gain full access to ``logs_20190115`` (fifth rule, as the ``admin`` permission gets higher priority than the ``read`` permission in the fourth rule.
+
+The same set would deny the service user to
+
+* gain any access to ``messages_2019`` (no matching rules),
+* read or search documents from ``events_2018`` (the second rule only grants ``write`` permission), and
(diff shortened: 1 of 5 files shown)


docs/products/opensearch/concepts/access_control.rst
```diff
@@ -0,0 +1,81 @@
+Access control
+==============
+
+Aiven for OpenSearch supports index-level access control lists (ACL) to control permissions.
```
lornajane, line 4: Can we say a little bit about why this is useful? For example, to grant access to create or change data, restrict access to certain data sets for reading, something else? Probably the word "security" belongs here too.

docs/products/opensearch/concepts/access_control.rst
```diff
@@ -0,0 +1,81 @@
...
+* ``admin``: full access to APIs and documents 
+* ``readwrite``: full access to documents
+* ``read``: allow only searching and retrieving documents
+* ``write``: allow updating, adding, and deleting documents
+
+Rules are defined separately for each user as ``pattern/permission`` combinations. The ``pattern`` defines the indexes that the permission applies to. Patterns are glob-style, where ``*`` matches any number of characters and ``?`` matches any character. 
+
+When multiple rules match, they are applied in the order listed above. If no rules match, access is denied.
+
+As an example, we can use the following set of rules:
```
lornajane, line 18: Add a subheading to split the example from the information

docs/products/opensearch/howto/control_access_to_content.rst
```diff
@@ -0,0 +1,34 @@
...
+
+   This creates a set of rules that grant full access (``_*/admin``, ``*/admin``) for each existing service user.
+
+   .. note::
+      After you switch on ACL support, the service does not automatically add an ACL for new service users that you add. By default, access is denied.
+
+   
+4. If you want to enforce index rules in a limited fashion for requests that use the ``_mget``, ``_msearch``, and ``_bulk`` APIs for this service user, switch on **Enable extended ACLs**.
+   
+5. To define new ACLs:
```
lornajane, line 25: This is almost a different task. I think they should stay together since they're both quite short but let's have something like:
 - one section on enabling ACLs or extended ACLs as a task
 - and then, another section how to create each individual ACL as another task, this one will be repeated, the other one only needs to be done once.

lornajane (changes requested):
Thanks! I made some minor comments to change but it's in good shape

docs/products/opensearch/concepts/access_control.rst
diawarr: Done ;)

docs/products/opensearch/concepts/access_control.rst
diawarr: Done ;)

docs/products/opensearch/howto/control_access_to_content.rst
diawarr: I'm not sure about this - I understand the use case for adding and editing ACLs when the feature is already in use, but is there a use case for just switching on ACLs without defining new ones or editing any existing ones?

lornajane: I don't think there's a case for switching on and not creating at least one ACL. But I think there are many cases where it's enabled and new ACLs would be created, or existing ones edited. And in that case, the task could be separate.

docs/products/opensearch/concepts/access_control.rst
```diff
@@ -0,0 +1,84 @@
...
+* gain full access to ``logs_20190201`` (fifth rule), and
+* gain full access to ``logs_20190115`` (fifth rule, as the ``admin`` permission gets higher priority than the ``read`` permission in the fourth rule.
+
+The same set would deny the service user to
+
+* gain any access to ``messages_2019`` (no matching rules),
+* read or search documents from ``events_2018`` (the second rule only grants ``write`` permission), and
+* write to or use the API for ``logs_20171230`` (the first rule only grants ``read`` permission).
+
+.. note::
```
hab-anka, line 42: I wonder if this note does not sit together with the ordered list above that introduces the types of permissions (start line 6).

docs/products/opensearch/howto/control_access_to_content.rst
```diff
@@ -0,0 +1,34 @@
+Control access to content in your service
+=========================================
+
+This article shows you how to set up access control lists (ACL) for content in Aiven for OpenSearch services.
+
+See the :doc:`related documentation <../concepts/access_control>` for more information on the permissions and patterns that are supported in ACL rules.
+
+To define ACLs for your service:
+
+1. Log in to the Aiven web console and select your OpenSearch service.
```
hab-anka, line 10: Should we link the console from here?

docs/products/opensearch/howto/control_access_to_content.rst
hab-anka: I think we could have a separate task in the same topic (two sub-tasks with their own headings) - as these are very related and rather small. Also, as creating user ACLs has 7 steps on its own, maybe being separate, that looked a bit more approachable and easy.

hab-anka (approved):
Small suggestions, otherwise looks perfectly good.

docs/products/opensearch/howto/control_access_to_content.rst
diawarr: OK - that's now updated ;)

docs/products/opensearch/concepts/access_control.rst
diawarr: True, it probably is better there - that's now moved ;)

docs/products/opensearch/howto/control_access_to_content.rst
diawarr: Might as well ;)

lornajane (approved):
Looking good

PR #158, opened 2021-08-09
Added `avn events, credits, cards` commands

ftisiot:

Added docs about `avn` cli events, credits and cards

### DIFF
--- a/docs/tools/cli/card.rst
+++ b/docs/tools/cli/card.rst
@@ -1,0 +1,111 @@
+Command reference: ``avn card``
+==================================
+
+Here you’ll find the full list of commands for ``avn card``.
+
+
+Manage credit cards
+-------------------
+
+Commands for managing credit card details associated to Aiven projects.
+
+
+``avn card add``
+'''''''''''''''''''''''
+
+Adds a new credit card to the Aiven account.
+
+.. list-table::
+  :header-rows: 1
+  :align: left
+
+  * - Parameter
+    - Information
+  * - ``--cvc``
+    - Credit card security code
+  * - ``--exp-month``
+    - Card expiration month (1-12)
+  * - ``--exp-year``
+    - Card expiration year
+  * - ``--name``
+    - Name on the credit card
+  * - ``--number``
+    - Credit card number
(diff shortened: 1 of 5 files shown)


docs/tools/cli/card.rst
```diff
@@ -0,0 +1,111 @@
...
+  * - ``--number``
+    - Credit card number
+  * - ``--update-project``
+    - Assign card to a project
+
+**Example:** Add a credit card to a named project.
+
+::
+
+  avn card add --cvc 000         \
```
hab-anka, line 41: I wonder if we should not use example values in the examples, such as name, card number. We can maybe agree on the standard values we can use for that?
A generic question - same goes for other examples here, and in other CLI commands.

hab-anka (approved):
General comment about example values - should we make them more realistic (using realistic names, numbers, etc.)?

docs/tools/cli/events.rst
```diff
@@ -0,0 +1,38 @@
...
+List project events
+-------------------
+
+Commands for listing events related to a particular project.
+
+
+``avn events``
+'''''''''''''''''''''''
+
+Lists instance or integration creation, deletion or modification events.
```
hab-anka, line 16: Is this like a log of events? It is not exactly clear what the events are? Should we maybe show more realistic info in the example?

hab-anka (approved):
For the events, I wonder if we can describe it better - what kind of events are meant here? Is it like "log" information for the project?

docs/tools/cli/card.rst
lornajane: A good question, I updated some of the examples.

docs/tools/cli/events.rst
lornajane: I had to look into this to answer the question and it turns out these are not logs (as I had thought) but more like an audit trail of operations performed on the project. I've updated this page to give more information and an output example

lornajane:
I had to do so much research to answer my own questions that I ended up editing the pages rather than commenting on them all! @ftisiot , @hab-anka any comments?

PR #165, opened 2021-08-12
sample of using partial includes

lornajane:
This is not for merging - just a small test to see how we could single-source different snippets of content.


lornajane:
We won't merge this, closing.

PR #167, opened 2021-08-12
Add OpenSearch sample dataset

### LINKED ISSUE #166: Add sample dataset for OpenSearch
Let's add a sample dataset and instructions on how to use it for OpenSearch.

YES!

jonatasbaldin:
Hi there!

This PR adds a HowTo guide to explore OpenSearch using a sample dataset.

Fixes #166.

### DIFF
--- a/docs/products/opensearch/howto/sample-dataset.rst
+++ b/docs/products/opensearch/howto/sample-dataset.rst
@@ -1,16 +1,14 @@
-Sample dataset
---------------
-Here you will find different sample datasets for OpenSearch and how to interact with them.
+Sample dataset: recipes
+=======================
 
-Before exploring the available datasets, we need an OpenSearch server to load the data on. A quick and easy way to get a server is through `Aiven for OpenSearch <https://aiven.io>`_. If don't have an Aiven account yet, `sign up <https://console.aiven.io/signup?utm_source=github&amp;utm_medium=organic&amp;utm_campaign=devportal&amp;utm_content=repo>`_ and enjoy our free trial!
-
-After logging in the Aiven Console, chose the OpenSearch version, select the cloud and plan of your choice, give the service a name and hit the "Create Service" button. In a couple of minutes, you will have an OpenSearch server up and running.
+Databases are more fun with data, so to get you started on your OpenSearch journey we picked this open data set of recipes as a great example you can try out yourself.
 
 Epicurious recipes
-==================
+------------------
+
 A dataset from `Kaggle <https://www.kaggle.com/hugodarwood/epirecipes>`_ with recipes, rating and nutrition information from `Epicurious <https://www.epicurious.com>`_.
 
-Let's take a look at the a sample document:
+Let's take a look at a sample recipe document:
 
 .. code:: json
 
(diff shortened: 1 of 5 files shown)


docs/products/opensearch/howto/sample-dataset.rst
```diff
@@ -0,0 +1,169 @@
+Sample dataset
+--------------
+Here you will find different sample datasets for OpenSearch and how to interact with them.
```
lornajane, line 3: Is there more than one dataset? Or just recommending the one?

docs/products/opensearch/howto/sample-dataset.rst
```diff
@@ -0,0 +1,169 @@
...
+
+Before exploring the available datasets, we need an OpenSearch server to load the data on. A quick and easy way to get a server is through `Aiven for OpenSearch <https://aiven.io>`_. If don't have an Aiven account yet, `sign up <https://console.aiven.io/signup?utm_source=github&amp;utm_medium=organic&amp;utm_campaign=devportal&amp;utm_content=repo>`_ and enjoy our free trial!
```
lornajane, line 5: Now that these instructions have moved to devportal, this could instead become "We will need an OpenSearch service to use for this example" and then the rest of the instructions could follow I think. It would also be OK to include the signup link.

docs/products/opensearch/howto/sample-dataset.rst
```diff
@@ -0,0 +1,169 @@
...
+
+After logging in the Aiven Console, chose the OpenSearch version, select the cloud and plan of your choice, give the service a name and hit the "Create Service" button. In a couple of minutes, you will have an OpenSearch server up and running.
+
+Epicurious recipes
+==================
```
lornajane, line 10: Let's engage the user a little here, it's important to be factual but I think "a great way to try the service is to play with it, and who doesn't like to play with their food?" type of approach to introduce some fun-but-harmless sample data to get people going would be good before diving in.

docs/products/opensearch/howto/sample-dataset.rst
```diff
@@ -0,0 +1,169 @@
...
+        "fat": 1.0,
+        "protein": 1.0,
+        "rating": 5.0,
+        "sodium": 959.0,
+    }
+
+
+Load the data
+'''''''''''''
+Follow the steps below to load the sample data into your OpenSearch service:
```
lornajane, line 48: I think this needs a heads up that there are other ways to do this but the example uses python to read the data and push it up to OpenSearch.

docs/products/opensearch/howto/sample-dataset.rst
```diff
@@ -0,0 +1,169 @@
...
+''''''''''''''''''''''
+After playing around with the sample queries, can you use OpenSearch queries to answer some these questions?
+
+1. Find all vegan recipes and order them by ``calories``.
+2. Find all recipes with ``vegan`` on the title but without the words ``cheese``, ``meat`` or ``egs`` on any other field.
+3. Use one query to count how many ``vegan`` and ``vegetarian`` recipes there are.
+
+Clean up
+''''''''
+To clean up the environment and, run the following commands:
```
lornajane, line 165: I'm not sure what this clean up is doing, does it remove the data from the index?

I wonder if we should prompt the user to power off or delete their services through the web console if they aren't using them.

lornajane (commented):
A few quick comments on the wording, I will run through with tech review as well.

docs/products/opensearch/howto/sample-dataset.rst
lornajane: updated

docs/products/opensearch/howto/sample-dataset.rst
lornajane: I edited this

docs/products/opensearch/howto/sample-dataset.rst
lornajane: Removed the section so we need to put it back if it was important and I just didn't understand something!

PR #169, opened 2021-08-13
adding cloud security general topic to platform

hab-anka:

Adding the general cloud security topic from https://help.aiven.io/en/articles/977466-cloud-security-overview into Platform concepts.


hab-anka:
Hi team @diawarr and @lornajane - I added a general topic to the platform section (on cloud security). The idea is to develop this section with very high level concepts in Aiven, starting with this one. I am proposing to distinguish the KEY concepts from ALL concepts in the container topic. Your feedback is welcome!

lornajane:
I like the idea of a "curated collection" over just having the automatic collection index pages. But having both on the page means we're seeing the cloud security topic twice. I'm not sure if it is a huge problem, but I'm definitely noticing it.

hab-anka:
I think in general we need to work more on how topics are included in navigation (both in the nav - which I think right now includes topics in alphabetical order, and in container topics). Ideally we would bring all topics in manually, in the order of their importance for the user flow. Which would prevent from having duplicates - just like this.

lornajane (approved):
Thanks!

PR #171, opened 2021-08-13
Use incremental builds for preview

lornajane:

We have been using a full rebuild in the `make livehtml` command and it's getting slower now we have more content. Switch to incremental building, and add support for a more full build (with README docs for reference). Also stop including the CONTRIBUTING file as it's triggering a warning.


diawarr (approved):
Looks good :)

Issue #159, opened 2021-08-10
Add Google Analytics


gskhan:
https://github.com/sphinx-contrib/googleanalytics

GA ID Property UA-68276046-1

lornajane:
Fixed by #172, thanks!

PR #172, opened 2021-08-13
Add google analytics

aiven-hh:
Adds google analytics script to template. Implements #159. Decided not to use sphinx-contrib/googleanalytics -plugin because it does not work on newer sphinx version that we use and seems to be unmaintained overall. Plugin injects the script so this is basically the same thing but at least this works without any hacks (see plugin issues/prs for more info if interested).

### DIFF
--- a/_templates/base.html
+++ b/_templates/base.html
@@ -1,0 +1,16 @@
+{% extends "!base.html" %}
+
+{%- block extrahead %}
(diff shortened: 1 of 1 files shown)


lornajane (approved):
Thanks for the change and also the extra information about the plugin.

PR #173, opened 2021-08-16
Fix hyperlink syntax

derrix060:
# Fix hyperlink

There are a few cases in the README file that hyperlink was written using the markdown syntax


lornajane (approved):
Oops! Thanks for the fix :)

Issue #135, opened 2021-07-22
CLI docs for `avn user` command


lornajane:
Document the commands with this subcommand, including examples and an explanation of what it does

lornajane:
Fixed in PR #175

PR #175, opened 2021-08-16
Added `avn user` documentation

ftisiot:
Added `avn user` documentation with different file for `avn user access-token` which has several options.
Looking for feedback around structure since this could be used for `avn service` as well

### DIFF
--- a/docs/tools/cli/user.rst
+++ b/docs/tools/cli/user.rst
@@ -57,7 +57,7 @@ Retrieves the current user information such as:
 * Authentication method
 
 
-**Example:** Retrieves the information for the currently logged user.
+**Example:** Retrieve the information for the currently logged user.
 
 ::
 
@@ -76,7 +76,7 @@ An example of user information:
 ``avn user login``
 ''''''''''''''''''''
 
-Logs in the user.
+Logs the user in.
 
 
 .. list-table::
@@ -90,7 +90,7 @@ Logs in the user.
   * - ``--token``
     - Logs in the user with a pre-created token 
 
-**Example:** Log in the user ``john.doe@example.com`` with prompted password.      
+**Example:** Log the ``john.doe@example.com`` user in, and prompt for password.      
 ::
 
   avn user login john.doe@example.com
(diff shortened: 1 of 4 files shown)


hab-anka:
One general comment, I think all the commands should be on the same navigation level. Here we're nesting the `access token` under `user`, and I understand this may make sense, but from the nav perspective this is hidden and only visible via the link. 
I suggest we make them sibling in the nav, and then link from one to another - like you did in your `user `doc.

docs/tools/cli/user.rst
```diff
@@ -0,0 +1,129 @@
...
+
+Logs out the user.
+
+
+**Example:** Log out the user.      
+::
+
+  avn user logout
+
+``avn user tokens-expire``
```
hab-anka, line 119: I wonder if this command does not logically belong to the next file (`avn user access token`)?

hab-anka (commented):
One general comment about structure and a few edits in comments for grammar.

lornajane:
Fixes #135

docs/tools/cli/user.rst
lornajane: It does but I don't think we can fix that with the documentation. How about signposting the other page of commands (I think there are probably a few places we'd want to do a "see also" type of command linking.

ftisiot: fixed

docs/tools/cli/user.rst
```diff
@@ -0,0 +1,129 @@
...
+
+Here you'll find the full list of commands for ``avn user``.
+
+
+Create and manage users and sessions
+------------------------------------
+
+Commands for managing project users and ``avn`` client sessions.
+
+``avn user access-token``
```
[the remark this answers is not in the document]
ftisiot: Fixed

docs/tools/cli/user/user-access-token.rst
```diff
@@ -0,0 +1,126 @@
...
+    ====================  ============  ================================  ===============  ================  ====================  ===========  ===================
+    2021-09-15T15:29:14Z  XCJ3+bgWywIh  Test token                        2592000          true              2021-08-16T15:29:14Z  192.168.1.1  aiven-client/2.12.0
+    2021-08-16T16:26:10Z  6JsKDclT3OMQ  To be used with Python Notebooks  3600             true              null                  null         null
+
+
+
+``avn user access-token revoke``
+''''''''''''''''''''''''''''''''
+
+Revokes the specified user access token.
```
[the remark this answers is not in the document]
ftisiot: fixed thanks

lornajane (approved):
Thanks very much!

PR #181, opened 2021-08-18
Upgrade to Sphinx v4

lornajane:

We have a couple of bugs that have already been fixed and all our dependencies now have good support for Sphinx 4.0 so this PR bumps all the dependencies so we can test everything works as we expect.


diawarr (approved):
Seems legit, I don't see any issues with the changes ;)

Issue #126, opened 2021-07-22
CLI docs for `avn account` command


lornajane:
Document the commands with this subcommand, including examples and an explanation of what it does

lornajane:
Fixed in #185

PR #185, opened 2021-08-23
Initial commit for `avn account`

ftisiot:

Documentation for `avn account`: there are several options split in 3 files.

The `avn account authentication-method` probably needs better examples.

### DIFF
--- a/_toc.yml
+++ b/_toc.yml
@@ -23,6 +23,9 @@ entries:
     entries:
       - file: docs/tools/cli
         entries:
+        - file: docs/tools/cli/account
+          entries:
+            - glob: docs/tools/cli/account/*
         - file: docs/tools/cli/card
         - file: docs/tools/cli/cloud
         - file: docs/tools/cli/credits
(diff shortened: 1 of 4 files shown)


ftisiot:
Fixed  #126

ftisiot:
Added link

encima (approved):
Aside from the generic `LGTM`, this is a solid, concise overview of the functionality. You already added the link to the `concepts` page and I think it is super important that that is clear for all users; especially because we use `account` with at least 3 different meanings.

Big fan of the example use cases, in particular the explanation of their outcome.

Possibly off-topic for this PR but I cannot help but feel that this should be auto built from the code/openapi spec.

lornajane:
Off topic but I think we can't _quite_ build this ( with the words, and examples, and everything ) from the code but we should definitely contribute some wording improvements up into the CLI at some point and then we could get closer with generated code.

Related (but still off topic) I think there's something where we should be able to produce things like https://developer.aiven.io/docs/products/postgresql/reference/list-of-advanced-params.html from somewhere in the core codebase....

lornajane (approved):
If it's good enough for @encima then it's good enough for me. Thanks @ftisiot :)

lornajane:
Just going to rebase for that conflict before I merge, hold onto you hats

PR #191, opened 2021-08-25
Add OpenGraph meta tags

### LINKED ISSUE #139: Add OpenGraph metadata
OpenGraph makes for richer links and shared content, add some opengraph tags, see https://ogp.me/

aiven-hh:
Adds opengraph sphinx extension that adds title, type, url,
description and possible image ogp meta tags to page head.

Fixes #139


aiven-hh:
As a side note, index page should probably start with some intro text instead of the search to make description tag a bit nicer.

lornajane (approved):
This is a great addition, thanks for adding the image too :)

PR #198, opened 2021-09-03
Add PHP closing code block on the PG connect howto

jonatasbaldin:


lornajane:
In PHP, it's normal to exclude the closing tag when it is the last thing in the file. See also: https://www.php.net/manual/en/language.basic-syntax.phptags.php

PR #208, opened 2021-09-16
build: automate requirements installation and use venv to install requirements only locally

pkalliok-aiven:
This simply takes one step away from the build process.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,50 +4,63 @@
 # You can set these variables from the command line, and also
 # from the environment for the first four.
 SPHINXOPTS    ?=
-SPHINXBUILD   ?= sphinx-build
-ES_URL		  ?=
-PG_URL		  ?=
+SPHINXBUILD   ?= ./myenv/bin/sphinx-build
+AUTOBUILD     ?= ./myenv/bin/sphinx-autobuild
+ES_URL        ?=
+PG_URL        ?=
 SOURCEDIR     = .
 BUILDDIR      = _build
+PYTHON        = ./myenv/bin/python
 
 # Put it first so that "make" without argument is like "make help".
-help:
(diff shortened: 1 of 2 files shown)


pkalliok-aiven:
Ah.  It seems to break the netlify build (for some reason, setting up a virtualenv in netlify is broken).  I can investigate but wouldn't want to put too much time to this :/

Makefile
```diff
@@ -4,50 +4,63 @@
...
 # Put it first so that "make" without argument is like "make help".
-help:
+help: requirements-stamp
 	@$(SPHINXBUILD) -M help "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)
 
-.PHONY: help Makefile
+.PHONY: help Makefile livehtml livehtmlall spell create-index index-devportal index-helpcenter create-feedback-table
+
+myenv:
+	python -m venv myenv
```
vladan, line 22: maybe be explicit with `python3` here? (not sure if it's the default in macos)

Makefile
```diff
@@ -4,50 +4,63 @@
...
+
+requirements-stamp: requirements.txt myenv
+	@echo Installing sphinx dependencies...
+	./myenv/bin/pip install -r requirements.txt
```
vladan, line 26: not sourcing the env here might not set the python path correctly

lornajane:
I'd be super happy to have this moved to the `README` so that people could choose to use it, or if they wanted to do something different on their setup, or it's being used by netlify then the makefile doesn't make assumptions but the instructions are right there if needed.

lornajane:
I'm abandoning this but additional instructions for the `README` would be welcome (let's leave the build alone since Netlify does this its own way)

PR #211, opened 2021-09-17
Edited plugins article according to planning changes

diawarr:
Removed reference to optional toggling of additional plugins and replaced the description with a vague "coming soon" reference.


lornajane (approved):
Thanks!

PR #212, opened 2021-09-17
Fix typos in PG migration docs

ivanr-aiven:


lornajane (approved):
Thanks very much for the small but important fix!

PR #217, opened 2021-09-23
Add TLS/SSL certificates page

jonatasbaldin:
Hi there 👋 

On making the PostgreSQL code examples self executable, we faced the need to have instructions on how to download the CA certificates from Aiven.

This pages explains the Aiven's certificate infrastructure, which services require the certificate download and how to download it.

It was written based on the [TLS / SSL Certificates](https://help.aiven.io/en/articles/869463-ssl-tls-certificates) article from the Help Aiven docs (we probable need to setup the redirect from there to this page).


docs/platform/concepts/tls-ssl-certificates.rst
```diff
@@ -0,0 +1,53 @@
...
+=========================
+
+You can download the CA certificate through the Aiven `web console <https://console.aiven.io>`_ by accessing the service page and clicking on the ``Download`` button on the ``CA Certificate`` line.
+
+.. image:: /images/platform/ca-download.png
+    :alt: A screenshot of the Aiven web console service page, showing where is the CA certificate download button.
+
+Or, you can use the ``avn`` :doc:`command-line tool <docs/tools/cli>` with the following command::
+
+  avn service user-creds-download --username <username> <service-name>
```
ftisiot, line 53: Can you add here that the username is the `avnadmin` by default?

ftisiot (commented):
Just a minor comment from my side, looks good

lornajane (commented):
I started to comment to say that we needed to add OpenSearch but the more I think about it, this list is just one more thing to maintain for every product release. Since the majority of services use the letsencrypt certs now (and I think that's what we're always aiming for), could we just say "other services ..." and then add a general note about older services potentially being different?

Whichever way you go on adding all current products or making it general: this is in great shape, the image is nice, and thanks for mentioning the redirect!

docs/platform/concepts/tls-ssl-certificates.rst
lornajane: I didn't, because I am not sure how it helps. People might want to use other users, or something.

lornajane:
Moved the best bit (nice screenshot!) to a separate howto on downloading the CA so we can link to it more easily, and tidied up the list of products to just mention the exceptions to our rule. Thanks @jonatasbaldin

lornajane (approved):
I did refactor, but this was a good replacement for the original page, thanks!

PR #218, opened 2021-09-23
Fix PostgreSQL NodeJS code sample

jonatasbaldin:
Hi there 💃

This PR makes the PostgreSQL NodeJS code sample article sample self-executable.

As you can see on the code sample, there's a bug from the ``pg`` library when using the `connectionString` parameter. Basically, the `?sslmode=require` string messes around with the `ssl` configuration, so I stripped it off.

Another workaround is to _not_ use the `connectionString`, but use all the parameters separated, like:
```js
const client = new Client({
  user: 'dbuser',
  host: 'database.server.com',
  database: 'mydb',
  password: 'secretpassword',
  port: 3211,
})
```

I still prefer the first one, when the user has less trouble copying/pasting connection details but has a bit more of code boilerplate.

What do you think?


code/products/postgresql/connect.js
```diff
@@ -1,15 +1,23 @@
...
 const config = {
-    connectionString: POSTGRESQL_URI
+    // We need to strip the ?sslmode=required from the connection string
+    // due to a bug on the `pg` library
+    // https://github.com/brianc/node-postgres/issues/2558#issuecomment-855703980
+
+    connectionString: postgreSqlUri.replace('?sslmode=require', ''),
     ssl: {
-        rejectUnauthorized: false,
+        ca: fs.readFileSync(path.join(__dirname, '/ca.pem')),
```
aiven-hh, line 14: The slash in not necessary before ca.pem.

aiven-hh:
Is php page still missing the step to download the cert or is that in another pr?

lornajane (changes requested):
This change makes some assumptions about how the user is running the code, that I think makes it less portable. If it's an existing project, creating the directory isn't the right thing (and strictly speaking, is not a pre-requisite, so shouldn't be included).

We can't always assume filesystem access (for example if this is a lambda function usage, or the code is deployed to a cloud service like Heroku) - so to include the CA, perhaps we can create some separate content on how to handle the certificates in different environments, rather than hardcoding a filename.

lornajane:
Replaced by #265

PR #243, opened 2021-09-24
Add OpenSearch icon file in expected location

lornajane:

We were missing an icon from the homepage.


ftisiot (approved):
Icon fixed!

PR #244, opened 2021-09-24
initial set of `avn service` docs

ftisiot:
Initial set of `avn service` docs


ftisiot:
Solve #219 #220 #221 #222 #223 #224 #225 #226 #227 #233 #234 #257 #259

ftisiot:
Hi @anelook and @lornajane, the `avn service` is our big monster, I would like an initial feedback on the current content and structure. Is it too much for a page? should we divide it better?
Thanks!

lornajane:
(sorry I pressed the wrong button there!)

This is definitely a huge task, and of course the sooner we start the sooner it'll be done, so I'm impressed already!

Are you intending to get all of `avn service *` into one pull request? For `avn account` I grouped commands together rather than having purely an alphabetical list, in case people wanted to do a thing that they didn't already know the command name for.

What are you using as the basis for when to go into a subpage or not? I think we might need to do more subpages for this command that has so much at the top level that the page is really, really long.

ftisiot:
Are you intending to get all of `avn service *` into one pull request? Nope, I started documenting... and added a bit too much in one go :D

As of now I created a subpage only when there were multiple subcommands (`connector` is the example) but I agree I'll subdivide the content per area, for example, all the `avn service acl*` will go into a separate file.

ftisiot:
Rebased! Should be ready for the final check before merging into main

lornajane:
Replaced by #332

PR #264, opened 2021-09-29
Add three examples of how tutorials can be presented

anelook:
Suggesting three options how to present all our tutorials on a single page.
Goal -  users should be able to easily find topics they are interested in, even if the list of tutorials is very long.


anelook:
Closing for now. Will come back to this topic when we are ready to have a list of tutorials

PR #267, opened 2021-09-30
Added a brief description of watermarks

diawarr:
A very brief explanation for watermarks - there would be quite a lot of detail added, but in the end I decided to keep it to a similar level of brevity as the other entries here ;)

Also fixed a couple of cross-reference links that I missed earlier.


docs/products/flink/concepts/important_concepts.rst
```diff
@@ -27,7 +27,17 @@ Events may still arrive after the allowed lateness for the window, so your appli
 
 For more information, see the `Apache Flink documentation <https://ci.apache.org/projects/flink/flink-docs-release-1.13/docs/dev/datastream/operators/windows/>`_.
 
-Another concept closely related to windows in Apache Flink is *watermarks*. For more information on watermarks, see the `Apache Flink documentation <https://ci.apache.org/projects/flink/flink-docs-release-1.13/docs/dev/datastream/event-time/generating_watermarks/>`_.
+
+Watermarks
+----------
+
+Another concept closely related to windows and event time in Apache Flink is *watermarks*. Flink uses watermarks as a mechanism to measure progress in event time; they flow as part of the data stream, carrying a timestamp that declares the event time reached in the data stream.
```
ftisiot, line 34: I'm struggling with the following sentence, I believe that there might be a piece missing:

> they flow as part of the data stream, carrying a timestamp that declares the event time reached in the data stream

Maybe

> they flow as part of the data stream, carrying a timestamp that declares the **MINIMUM** event time reached in the data stream."

ftisiot (commented):
It's good! 
I struggled with a watermark sentence, I saw that the sentence is similar in the Flink documentation, but I felt we were missing a **MINIMUM** word possibly. 
I also believe a diagram would make it more clear

docs/products/flink/concepts/important_concepts.rst
diawarr: Good point, that's now added - I also added a basic diagram (pretty much copied from the Flink documentation), which might help a bit ;)

docs/products/flink/concepts/watermarks.rst
```diff
@@ -0,0 +1,31 @@
...
+
+Another concept closely related to windows and event time in Apache Flink is *watermarks*. Flink uses watermarks as a mechanism to measure progress in event time; they flow as part of the data stream, carrying a timestamp that declares the minimum event time reached in the data stream.
+
+This allows Flink to set points in the stream when all events up to a certain timestamp should have arrived, so that operators can set their internal event time to the value of the watermarks that reach them.
+
+For example, the following diagram shows how watermarks fit into an out-of-order stream, where the events are not ordered by their timestamps:
+
+.. mermaid::
+
+    graph TD;
```
ftisiot, line 12: The graph now looks VERY long. Not sure there are alternatives...

ftisiot (commented):
Overall structure is good, the new  watermark `TD` chart is very long now, not sure we can avoid it somehow

docs/products/flink/concepts/watermarks.rst
diawarr: It's a bit of a tricky case - the horizontal orientation means that it's quite small, and reducing the number of items kind of defeats the point of the graph... Would it make sense to leave it out for now?

PR #271, opened 2021-10-07
Reinstate the repetition rule in Vale for 2.11.1 release

lornajane:

There was a bug in the repetition rule in Vale in the 2.11.0 release, so we disabled the rule. The bug is fixed, so this reinstates the rule


lornajane:
#273 will fix the build error, merge that first

PR #272, opened 2021-10-07
Add a rule to check OpenSearch Dashboards capitalization

lornajane:

After the fix from @tvainika in #270, add a rule to make sure that the linter catches this if it happens again


lornajane:
#273 will fix the build error, merge that first

PR #274, opened 2021-10-08
Fixing CRD get example

davidespo:
```bash
$ kubectl get "pgs.aiven.io" davide-k8s-pg
error: the server doesn't have a resource type "pgs"
```
Not sure when this changed in the CRD template. Matching value from 

https://github.com/aiven/aiven-charts/blob/main/charts/aiven-operator-crds/templates/aiven.io_postgresqls.yaml#L9

### DIFF
--- a/docs/tools/kubernetes.rst
+++ b/docs/tools/kubernetes.rst
@@ -129,7 +129,7 @@ You can verify the status of your service with the following command.
 
 .. code:: bash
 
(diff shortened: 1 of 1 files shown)


davidespo:
Thanks @dbeech. Not sure why is it still blocked... @lornajane Is there a specific team that the reviewer needs to join?

lornajane:
And we are sorted, thanks, both of you :)

PR #275, opened 2021-10-11
added update and service types, update, terminate, plans

ftisiot:
Added `service types`, `service update`, `service terminate`, `service plans` commands


ftisiot:
Fixes #256 and #255

PR #280, opened 2021-10-12
[Don't merge] Switch to new icon set

lornajane:

We're refreshing our icon set, this adds them to everywhere they are used.

(Review and approve if it's good, but don't merge yet because it's a co-ordinated release)


lornajane:
No longer required.

PR #284, opened 2021-10-13
Add connect example for java

snuyanzin:
The PR provides an example of connection to Postgresql from Java


mdedetrich:
Does it make sense that instead of manually downloading a `jar` and passing it in with `cp` we use a barebones Maven `pom.xml` file with Postgres JDBC as a dependency?

Most Java developers don't download jar's manually these days so such an example using Maven might be more representative.

lornajane:
Thanks @snuyanzin , thanks @reta  :)

PR #285, opened 2021-10-14
Aiven cli service flink

ftisiot:
Added the Flink commands


lornajane:
Well, this branch adds the line, so if we merge this now, it will add it, which we don't want. So yes, I think the error (meaning the case-insensitive opensearch spelling) was introduced in both branches and should be fixed in both branches.

anelook (approved):
Run all the commands, found only one issue that cancelling a failed (or already cancelled) job will return a long confusing error. But Flink team is already aware of it and they have a Jira ticket. Therefore - approving this pull request. It is very helpful stuff for the CLI users

lornajane:
is someone rebasing for conflicts and then merging this @anelook or @ftisiot ? Do you need me to do something?

lornajane:
@ftisiot could you please comment on the status of this pull request? It's approved, but has conflicts.

ftisiot:
Weirdly enough I didn't see any notifications about this. I'll probably need to add additional work on it since we're going to add a couple more parameters to table and job creation this or the next week.

PR #305, opened 2021-10-25
Add a script to generate Service EOL table from Aiven API

aprepo:
- Installs tabulate as new dependency
- Script pulls data from Aiven API
- Output is rst table

Currently we have help articles that are maintained by hand: https://help.aiven.io/en/articles/4318172-eol-for-major-versions-of-aiven-services

The actual EOL information is managed in the Aiven DB and any changes to the services are are done automatically based on that information. Customers can also ask for extensions to the EOL dates.

Maintaining the information in multiple places is overhead and we should show the up2date information on thedeveloper portal.

This should be run as part of the build to generate a page with the EOL information.

See Aiven Jira: https://aiven.atlassian.net/browse/SI-1429

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -51,3 +51,6 @@ index-helpcenter:
 create-feedback-table:
 	python "$(SOURCEDIR)/scripts/create_feedback_table.py" \
 		--pg-url="$(PG_URL)"
+
+eol-table:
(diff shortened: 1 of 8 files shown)


lornajane:
@aprepo Can you say something about why Kafka 3 is showing here and whether we should be filtering the API call when creating this table?

lornajane:
Closing this, due to inactivity, not because I don't want the feature. Let's revisit when there's time to look at the data sources again and agree what makes sense here.

PR #306, opened 2021-10-26
Add Sitemap

lornajane:

We need a sitemap, this adds a plugin that generates one, and also advertises this in robots.txt

### DIFF
--- a/conf.py
+++ b/conf.py
@@ -26,6 +26,7 @@ author = 'Aiven Team'
 # extensions coming with Sphinx (named 'sphinx.ext.*') or your custom
 # ones.
 extensions = [
+    'sphinx_sitemap',
     'sphinx_panels',
     'sphinxcontrib.mermaid',
     'sphinx_external_toc',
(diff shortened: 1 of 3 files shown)


gskhan:
👋  @lornajane 
is it possible to add `changefreq` and `priority` to the sitemap items. And if we can add `lastmod` it will be awesome but we don't even have it in the marketing site right now.

https://www.sitemaps.org/protocol.html

lornajane:
I'm not sure what we'd do with priority, but (it and) the other two requests are open as issues on the project:
 - https://github.com/jdillard/sphinx-sitemap/issues/3
 - https://github.com/jdillard/sphinx-sitemap/issues/12

I suggest that we would need to build/contribute these ourselves. Should we deploy the sitemap as-is for now and then work on this, or do you want to wait until these features area also available?

gskhan:
@lornajane ok, we can push without those, to start
but the robots file is pointing to /docs/sitemap.xml where I can only see the sitemap in /sitemap.xml
just making sure that's correct

lornajane:
Good catch! Take another look now please?

gskhan (approved):
we're 👍

PR #307, opened 2021-10-26
Added a brief explanation of Kafka connector types

diawarr:
A somewhat simplified explanation of the differences between standard and upsert Kafka connectors, may require additional examples at some point.


ftisiot (approved):
Good Work

PR #313, opened 2021-11-02
Added information on public beta services

diawarr:
- Description of public beta services and related SLA (ticket HH-608 in Jira)


lornajane (approved):
Thanks! I'm so sorry for the slow review ...

PR #315, opened 2021-11-03
Fix M3 landing page link wording

lornajane:

This was a review comment in a previous PR but I somehow missed pushing the change, so it gets its own PR


ftisiot (approved):
Good one!

PR #317, opened 2021-11-04
Edited Flink introductory content based on feedback

diawarr:
- Some clarifications to the content on the benefits of Flink and data streams vs. batch processing


ftisiot (approved):
Looks good!

PR #324, opened 2021-11-09
quote environment variable to prevent shell wildcard expansion

carobme:


ftisiot (approved):
Thanks a lot! Good ones!

PR #326, opened 2021-11-12
Updated content on OpenSearch plugin availability

diawarr:
- Updated content and plugin list now that the inclusion of new OpenSearch plugins has been deployed to production

### DIFF
--- a/docs/products/opensearch/reference/plugins.rst
+++ b/docs/products/opensearch/reference/plugins.rst
@@ -28,6 +28,6 @@ These plugins are enabled on all Aiven for OpenSearch services by default:
 * `Notebooks <https://opensearch.org/docs/dashboards/notebooks/>`__
 * `Performance Analyzer <https://opensearch.org/docs/monitoring-plugins/pa/index/>`__
 * `OpenSearch Dashboards Reports <https://github.com/opensearch-project/dashboards-report>`__
-* `Scheduler for Dashboards Reports <https://github.com/opensearch-project/dashboards-report>`__
+* `Scheduler for Dashboards Reports <https://github.com/opensearch-project/dashboards-reports>`__
 * `OpenSearch Dashboards GANTT Charts <https://opensearch.org/docs/latest/dashboards/gantt/>`__
 * `OpenSearch Dashboards Trace Analytics <https://opensearch.org/docs/latest/monitoring-plugins/trace/index/>`__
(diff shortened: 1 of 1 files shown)


docs/products/opensearch/reference/plugins.rst
```diff
@@ -1,7 +1,7 @@
 Plugins available with Aiven for OpenSearch
 ===========================================
 
-Aiven for OpenSearch includes by default a standard a set of plugins. We are also working on adding a further set of new OpenSearch plugins. The exact plugins are listed in the sections further down this page.
+Aiven for OpenSearch includes a standard a set of plugins. In addition to the plugins that were previously available in Aiven for Elasticsearch, the standard set of plugins also includes some that are new for OpenSearch.
```
anelook, line 4: a standard set (extra 'a')

anelook, line 4: I'm not sure I understand correctly this phrase - "includes some that are new for OpenSearch."  The plugins  that were added recently are not new for OpenSearch. I'd rephrase it - something like this - 
"the list now includes additional plugins designed and developed specifically for OpenSearch".

docs/products/opensearch/reference/plugins.rst
```diff
@@ -21,17 +21,13 @@ These plugins are enabled on all Aiven for OpenSearch services by default:
 * Mapper Size
```
anelook, line 21: Some plugins have links, others don't, why so and what's the difference?

anelook (changes requested):
Thx for the article updates :) 
I've added couple of notes. There is a typo and I suggest some rephrasing.

(optional idea) I also wonder if the list of plugins can be divided into two or more subgroups to increase readability, I assume that those that have links are different from those that don't.

docs/products/opensearch/reference/plugins.rst
diawarr: True, well spotted - that's now fixed ;)

diawarr: Sure, that's now edited ;)

docs/products/opensearch/reference/plugins.rst
diawarr: The ones without links didn't have any previously either, but I've added them in now ;)

anelook (approved):
Thank you, looks great!

PR #328, opened 2021-11-13
Upgraded Sphinx to v4.3.0

runwuf:
`Sphinx 4.0.3` breaks when `make livehtml` with `Python 3.10.0`
<pre>
ImportError: cannot import name 'Union' from 'types' (/Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/types.py)
Command exited with exit code: 1
</pre>


lornajane (approved):
Thanks! Seems solid to me and good to know it will be OK on newer Python too.

PR #337, opened 2021-11-17
Add Beta label to Flink product.

anelook:
Added label to 
 - Flink card information on the landing page
 - Flink main menu item
 - corrected styles for the menu, so that visually they look on the same line

### DIFF
--- a/docs/products/flink/index.rst
+++ b/docs/products/flink/index.rst
@@ -1,5 +1,5 @@
-Flink :badge:`beta,cls=badge-secondary text-black badge-pill`
-=============================================================
+Apache Flink :badge:`beta,cls=badge-secondary text-black badge-pill`
+====================================================================
 
 What is Aiven for Apache Flink?
 -------------------------------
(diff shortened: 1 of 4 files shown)


docs/products/flink/index.rst
```diff
@@ -1,5 +1,5 @@
-Aiven for Apache Flink
-=======================
+Flink :badge:`beta,cls=badge-secondary text-black badge-pill`
```
lornajane, line 1: Would it still fit if it was "Apache Flink" ?

ftisiot: Yep, seen that we used Apache Kafka, we should do the same for Flink

lornajane:
Can we take another look at this one? There's a LOT of commits and some feedback waiting @anelook

anelook:
Updated the name, it fits into the space. However, it is not very consistent with other services, where we have short titles.

lornajane (approved):
Looks good, and I don't think it's too long

Issue #333, opened 2021-11-17
section `avn service connection-pool` doesn't contain the `update`, `list`, `delete` commands


ftisiot:
# What should we add to our content collections?

section `avn service connection-pool` doesn't contain the `update`, `list`, `delete` commands

# Is this something you would work on yourself?

[X] I plan to open a pull request for this

PR #339, opened 2021-11-18
Added content for `avn service connection-pool`

ftisiot:

Added content for `avn service connection-pool`

### DIFF
--- a/docs/tools/cli/service/connection-pool.rst
+++ b/docs/tools/cli/service/connection-pool.rst
@@ -31,7 +31,7 @@ Creates a new :doc:`PgBouncer connection pool </docs/products/postgresql/concept
   * - ``--pool-mode``
     - The :ref:`pool mode <pooling-modes>`. Possible values are ``transaction``, ``session`` and ``statement``
 
-**Example:** In the service ``pg-doc`` Create a new connection pool named ``cp-analytics-it`` for the database ``it-analytics`` with:
+**Example:** In the service ``demo-pg`` Create a new connection pool named ``cp-analytics-it`` for the database ``it-analytics`` with:
 
 * username ``avnadmin``
 * pool-size of ``10`` connections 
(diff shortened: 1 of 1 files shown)


ftisiot:
Fixed #333

docs/tools/cli/service/connection-pool.rst
```diff
@@ -31,17 +31,108 @@ Creates a new :doc:`PgBouncer connection pool </docs/products/postgresql/concept
...
+**Example:** In the service ``demo-pg`` Create a new connection pool named ``cp-analytics-it`` for the database ``it-analytics`` with:
 
 * username ``avnadmin``
 * pool-size of ``10`` connections 
 * ``transaction`` pool-mode
 
 ::
 
-  avn service connection-pool-create pg-doc \
+  avn service connection-pool-create demo-pg \
```
anelook, line 42: Interesting, when I'm trying to create  a connection pool with this command I get `ERROR	command failed: Error: {"errors":[{"message":"No valid password exists for user avnadmin.","status":400}],"message":"No valid password exists for user avnadmin."}`.

anelook (commented):
I would either move, or add this article to PG docs (CLI or not). It is not a global functionality, but specific for Postres, right?

I couldn't run the commands, getting an error related to a password. Wanted to check with you what can be wrong and try it again.

lornajane:
I'll chime in (but @anelook please keep driving this PR review) to say that I think all the CLI docs should stay together but I'd be very happy to have a per-product CLI cheatsheet, especially for specific commands like this one.

docs/tools/cli/service/connection-pool.rst
ftisiot: This is weird, I just tried again and the command works as expected, maybe you can try again recreating the bug and open an issue in JIRA?

ftisiot:
I'm with @lornajane, this is a command page and should stay together with the other commands. Yeah a cheatsheet would be good, should we open issues for this?

docs/tools/cli/service/connection-pool.rst
anelook: Mystery, can't reproduce the error any more! Will keep an eye if I see it again

anelook (approved):
This time round all went smoothly! Tested commands without issues.

anelook:
My opinion here is that the position of this command in commands tree is wrong, it belongs to PostreSQL specifically (can't be used with anything else, right?)

PR #342, opened 2021-11-18
Add steps to migrate an ES service to OS using TF

ak-aiven:
Many Aiven users are Terraform users and they need guidance on how to migrate from ES to OS using TF.


.gitignore
```diff
@@ -4,3 +4,4 @@ _build/*
 __pycache__
 .netlify
 node_modules
+.idea/*
```
anelook, line 7: I believe you should hide `.idea/*`  it in your global .gitignore.

docs/products/opensearch/howto/upgrade-to-opensearch.rst
```diff
@@ -43,3 +43,58 @@ Extra information for Terraform users
 -------------------------------------
 
 If you manage your infrastructure with our :doc:`Terraform provider </docs/tools/terraform>` then please note that you will need to upgrade the provider to the latest release in order to have support for OpenSearch.
+
```
anelook, line 46: When migrating through terraform, the steps described at the top of the file are not relevant, right? 

I feel that the terraform information stands on its own and for readers will be easier to understand it if it is put in a separate file. Can we move it to upgrade-to-opensearch-terraform.rst (or similar) and add a proper header "Upgrade to OpenSearch with Terraform"?

docs/products/opensearch/howto/upgrade-to-opensearch.rst
```diff
@@ -43,3 +43,58 @@ Extra information for Terraform users
...
+To perform the upgrade in place using Terraform, you will need to take the following steps:
+
+1. Change the ``elasticsearch_version = 7`` to ``opensearch_version = 1``. This is the equivalent to clicking the migrate button in the console.
+
+.. code-block::
```
anelook, line 51: Is it important to keep the block with `elasticsearch_version = 7`?  it is what user already have, right? Unless it gives information, I'd remove this block and only leave what user should have in the end.

docs/products/opensearch/howto/upgrade-to-opensearch.rst
```diff
@@ -43,3 +43,58 @@ Extra information for Terraform users
...
+
+    terraform state rm 'aiven_elasticsearch.es'
+
+3. Add the OpenSearch service to the Terraform state
+
+.. code-block::
+
+    resource "aiven_opensearch" "os" {
```
anelook, line 87: I'm also a bit confused why the change above is not equal to this one. Is it that we need to change elasticsearch_version to opensearch_version in one step and only then update name to import it to terraform state?

docs/products/opensearch/howto/upgrade-to-opensearch.rst
anelook: Or, do we want to clarify why there are two code snippets?

anelook (changes requested):
Thank you for adding this valuable information to dev portal!

I've added couple of comments with attempt to make it more clear to the readers. Could you see if  they make sense to you? If not - let's discuss ;)

docs/products/opensearch/howto/upgrade-to-opensearch.rst
ak-aiven: Very true. I can move this to a separate doc.

docs/products/opensearch/howto/upgrade-to-opensearch.rst
ak-aiven: Yes, the first block is what the user will already have if they manage their infrastructure with TF and I wanted to highlight the change that needs to be made to make the migration happen.

docs/products/opensearch/howto/upgrade-to-opensearch.rst
ak-aiven: Yes, these steps must be taken separately and linearly. If not TF will not be able to migrate properly.

anelook (changes requested):
Could you add an entry to the menu as well? It is in ``_toc.yml``. 
And could you remove ``.idea/*`` form project's .gitignore? those are specific to your computer, they should be part of your global git ignore file.

anelook (approved):
Looks good to me!

PR #343, opened 2021-11-18
Adding Aiven Terraform provider v1 to v2 migration guide

z0r1k:
Subj 👆


z0r1k:
It looks like preview link is the description is not correct one. Here is a proper link - https://deploy-preview-343--developer-aiven-io-preview.netlify.app/docs/tools/terraform/v1v2.html

z0r1k:
Linter complains about the following – `/home/runner/work/devportal/devportal/docs/tools/cli/service.rst:2: WARNING: Duplicate explicit target name: "columns".`. 
Does it cause the failure? Because this article is coming from master and already published.

lornajane:
Don't worry, I can take a look when I review

docs/tools/terraform/v1v2.rst
```diff
@@ -0,0 +1,216 @@
+Upgrading the Aiven Terraform Provider from v1 to v2
+====================================================
+
+Version 2 of the Aiven Terraform Provider was released in `October of
+2020 <https://aiven.io/blog/aiven-terraform-provider-v2-release>`__. You
```
lornajane, line 5: I think it would be better to say "We recommend you upgrade to v2 if you haven't already, and this guide will walk you through the process". Don't reference other possible approaches unless we have docs or something else to point to.

docs/tools/terraform/v1v2.rst
```diff
@@ -0,0 +1,216 @@
...
+   Card ID
+-  Work is being done to deprecate ``aiven_service`` in order to support
+   individual service configuration better, using ``aiven_kafka`` for
+   example
+-  New resources has been added to the Aiven Terraform provider, such as
+   ``aiven_flink``, ``aiven_opensearch``, etc.
+-  More details of all of the changes can be found in the
+   `Changelog <https://github.com/aiven/terraform-provider-aiven/blob/master/CHANGELOG.md>`__.
+
+Original setup
```
lornajane, line 23: I'm a little confused. Are we walking people through installing an old version here? I think if they are following an upgrade guide, we can assume they already have an old version so we can skip this step completely.

docs/tools/terraform/v1v2.rst
```diff
@@ -0,0 +1,216 @@
...
+``terraform init``
+
+``terraform plan``
+
+``terraform apply``
+
+Upgrade Terraform
+-----------------
+
+Kick things off with 0.13
```
lornajane, line 108: I'm not sure what "Kick things off" means in this context. Do we mean that they should first upgrade Terraform itself? Reading ahead, it looks like there's specifically an upgrade from <0.13 to specifically 0.13 and then another upgrade to get beyond that version, is that correct? I think we should start this section with an explanation of why there are multiple upgrade steps.

docs/tools/terraform/v1v2.rst
```diff
@@ -0,0 +1,216 @@
...
+
+You will likely have an existing state file, by running:
+``terraform state replace-provider registry.terraform.io/-/aiven registry.terraform.io/aiven/aiven``
+you will replace old Aiven terraform provider references to the new
+format.
+
+Also update ``required_version`` from ``>= 0.12`` to ``>= 0.13`` in your
+``versions.tf`` file if you have any.
+
+We can remove the old Terraform folder ``rm -rf ~/.terraform.d``
```
lornajane, line 124: Is there any risk in removing this file? If they could keep it in case they need to rollback or something, we can move this instruction to a "finishing up" section at the end or something, once they've tested that things still work.

docs/tools/terraform/v1v2.rst
```diff
@@ -0,0 +1,216 @@
...
+          version = "2.3.1"
+        }
+      }
+    }
+
+After that you can run: ``terraform 0.13upgrade`` to see fixes
+recommended by Hashicorp. More information
+`here <https://www.terraform.io/upgrade-guides/0-13.html>`__.
+
+Bump it to the latest
```
lornajane, line 144: Here we have "Bump it" but does it mean "Upgrade Terraform"?

docs/tools/terraform/v1v2.rst
```diff
@@ -0,0 +1,216 @@
...
+Migration strategy – since ``aiven_service`` and ``aiven_X``, i.e.
+``aiven_kafka`` are different kind of resources just by rewriting code
+we would cause destructive actions.
+
+Also running ``terraform state mv <a> <b>`` will not do it because it is
+a different resource type.
+
+To solve it we would have to:
+
+-  Change the code first
```
lornajane, line 176: I think this step is covered in this section - but what about the other two steps?

docs/tools/terraform/v1v2.rst
```diff
@@ -0,0 +1,216 @@
...
+
+Here is the example of the steps above:
+
+::
+
+    $ terraform state list | grep kf
+    $ terraform state rm aiven_service.kf
+
+    $ terraform import aiven_kafka.kf mischa-demo/kafka-test-gr0gu
+    $ terraform plan
```
lornajane, line 209: Is there a way before the user runs apply that they can check things have worked well? Does plan tell us something we can use?

docs/tools/terraform/v1v2.rst
```diff
@@ -0,0 +1,216 @@
...
+    $ terraform apply
+
+For more examples please check Aiven Examples
```
lornajane, line 212: I love that there are examples in the aiven-examples repo, but why aren't we directing users to the terraform repo itself? I thought the project had its own docs.

docs/tools/terraform/v1v2.rst
```diff
@@ -0,0 +1,216 @@
...
+`repository <https://github.com/aiven/aiven-examples/tree/master/terraform>`__
+on GitHub.
+
+If you have any issues, please contact our `support <mailto:support@aiven.io>`__.
```
lornajane, line 216: The last line isn't needed, let's stick specifically to the steps needed for the provider upgrade.

lornajane (changes requested):
Thank you so much for writing this up, it's super useful! I've got quite a few questions (I'm not terraform native either, which might be obvious) about what we need to include but I think this is going to be an excellent addition to the docs.

docs/tools/terraform/v1v2.rst
z0r1k: Idea here was just to demonstrate potential original setup using v1 provider (which they most probably have) so they could compare it to v2 to see the difference. 

But we can replace it with something like "We assume you already have set some services up using Aiven Terraform provider v1.3+ and Terraform 0.12+". Wdyt?

docs/tools/terraform/v1v2.rst
z0r1k: I was trying to explain it here – https://github.com/aiven/devportal/pull/343/files/1d46af02cbe13d06a8217b8d20de341afb1bfb1c#diff-09583ec3f3df84677dd731f7ab9ec32f58af3b7d2f2e0f85b59ae048c236f9faR111
Basically if someone is using TF 0.12 they would need to migrate to the new TF syntax which was introduced in 0.13 first by using a command which was later removed. Then they would need to jump from 0.13 to the latest.

docs/tools/terraform/v1v2.rst
z0r1k: That file is an old Aiven TF provider (v1) binary. They could get it back from here – https://github.com/aiven/terraform-provider-aiven/releases/tag/v1.3.5 in case of rollback.

docs/tools/terraform/v1v2.rst
z0r1k: Yep, it implies updating TF version to the latest available.

docs/tools/terraform/v1v2.rst
z0r1k: Yep, it implies updating TF version to the latest available.

docs/tools/terraform/v1v2.rst
z0r1k: Here they are – https://github.com/aiven/devportal/pull/343/files/1d46af02cbe13d06a8217b8d20de341afb1bfb1c#diff-09583ec3f3df84677dd731f7ab9ec32f58af3b7d2f2e0f85b59ae048c236f9faR201

docs/tools/terraform/v1v2.rst
z0r1k: Yes, plan would show them what TF is planning to do. I also will modify the command to take the backup of the current state file.

docs/tools/terraform/v1v2.rst
z0r1k: Our TF provider repo has examples on how to setup certain services but not much in regards to v1 to v2 migration.

docs/tools/terraform/v1v2.rst
lornajane: I think that would be good. Just "this guide is for users already using version[whatever] and ready to upgrade" would work. Otherwise I'm following all the steps thinking this is how to upgrade. We could even signpost the project installation instructions in case people end up here somehow without being on v1.

docs/tools/terraform/v1v2.rst
lornajane: It's making more sense now! So people using our old terraform provider might be using old terraform? I thought the article was about upgrading which of our providers you use. Can we assume they are on old terraform? We almost need a "choose your own adventure" format so people can pick out the sections that they need to follow.

docs/tools/terraform/v1v2.rst
lornajane: Ah OK. I still think we could maybe put the cleanup after the "look, it works!" step.

docs/tools/terraform/v1v2.rst
lornajane: Let's reword the title to make it clear for anyone skim reading the navigation headings.

docs/tools/terraform/v1v2.rst
lornajane: Is there any danger that we could also get the relevant examples into the project they belong to? Perhaps as a separate issue we can work on after getting this excellent migration guide published though!

lornajane (changes requested):
A few comments in response to your responses, but I still would like to remove the v1 installation steps and make the other elements clearer. I didn't understand the first time I read it that we were upgrading terraform itself as well as using a newer Aiven terraform provider, so that needs to be clearer - also if people have already upgraded terraform, we should signpost which steps don't apply to them.

docs/tools/terraform/v1v2.rst
z0r1k: Yes, we can check what examples are available in terraform provider repo itself and then see what could be added.

z0r1k:
I removed TF v1 section and added more explanation in case TF < 1 is used and if TF >= 1 is used. I hope it is clearer now. 
Thanks.

docs/tools/terraform/v1v2.rst
lornajane: OMG now I've read the examples there, that's really helpful and gives a lot of context that I wasn't sure about!! I'm adding some edits from there to this post, please watch this space.

z0r1k:
I'll merge this one since it's been thru some rounds of review already and I would like to communicate this guide to a customer.  
I've created an issue to track any further changes which needs to be done to this guide. 
Ty.

lornajane:
Could you please wait for the review stage? I've spent a couple of hours on this today. Also if you have a deadline, that would have been useful to know a bit sooner.

z0r1k:
> Could you please wait for the review stage? I've spent a couple of hours on this today. Also if you have a deadline, that would have been useful to know a bit sooner.

Sure. I will wait then. 
Ty.

encima (approved):
Love it. Love it. All the LGTMs.

lornajane (approved):
Could you please both take another look? Particularly the last section has changed ... but also I switched up the order of the steps so that upgrading Terraform was one section, and upgrading the provider was separate (since some people might need to do only one of those) - hopefully this won't break anything and our old provider would work on new Terraform also! Let me know if I broke anything but I'm also adding my own approval so either of you can go ahead and merge.

z0r1k:
> Could you please both take another look? Particularly the last section has changed ... but also I switched up the order of the steps so that upgrading Terraform was one section, and upgrading the provider was separate (since some people might need to do only one of those) - hopefully this won't break anything and our old provider would work on new Terraform also! Let me know if I broke anything but I'm also adding my own approval so either of you can go ahead and merge.

Thanks a lot for improving the article. I double checked the flow and had to switch up few things but it is all good and works 💪

PR #348, opened 2021-11-19
Virtualenv

StarfallProjects:
Added the following:

- `env` and `venv` to `.gitignore`. These are commonly-used names for Python virtual environments.
- Added a recommendation to the README, advising people to use virtualenv or venv.

Using a virtual environment is generally a best practice ([explanation](https://towardsdatascience.com/why-you-should-use-a-virtual-environment-for-every-python-project-c17dab3b0fd0)), and also a convenience. I'm not suggesting enforcing use if people don't want to, but helping folks _not_ commit their venv folder by accident might save the occasional headache.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,6 @@ __pycache__
 .netlify
 node_modules
 
-# ignore common virtual environment names
+
 venv
 env
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,7 @@ _build/*
 __pycache__
 .netlify
 node_modules
+
+# ignore common virtual environment names
+venv
+env
--- a/README.rst
+++ b/README.rst
@@ -8,6 +8,8 @@ It is Python-based, with content in `ReStructuredText (rst) <https://docutils.so
 Running locally
 ---------------
 
+We recommend using a virtual environment like `virtualenv <https://virtualenv.pypa.io/en/latest/>`_ or `venv <https://docs.python.org/3/library/venv.html>`_.
+
 Install the dependencies::
 
     pip install -r requirements.txt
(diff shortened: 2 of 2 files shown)


anelook:
My opinion on this is that local tools, personal dev editor folders should go into a personal global to a machine ~/.gitignore, and not to the .gitignore of the repository.  It is convenient for a developer, and avoids mess in the repository file. 

Any reason why you prefer not to use your computer's global ~/.gitignore file?

StarfallProjects:
I've just been caught out before by contributors not doing it.

anelook:
Ah, I see. There are definitely different takes on this :)

README.rst
```diff
@@ -8,6 +8,8 @@ It is Python-based, with content in `ReStructuredText (rst) <https://docutils.so
 Running locally
 ---------------
 
+We recommend using a virtual environment like `virtualenv <https://virtualenv.pypa.io/en/latest/>`_ or `venv <https://docs.python.org/3/library/venv.html>`_. Call your virtual environment `env` or `venv`, or add you preferred name to `<.gitignore>`_.
```
lornajane, line 11: This is a really good tip and I'd like us to have it in the `README`. Let's recommend venv and our own choice of folder name, and show the command on setting it up - with  reminder to add the folder name to gitignore.

.gitignore
```diff
@@ -4,3 +4,7 @@ _build/*
 __pycache__
 .netlify
 node_modules
+
+# ignore common virtual environment names
```
lornajane, line 8: Remove this because we don't know what the user likes to call their directories - but in the README, let's prompt them to add whatever they chose to `.gitignore`. If they get this step wrong, it won't impact on their ability to work with the tool and we would pick up at review if they'd added something they shouldn't have (I hope!)

StarfallProjects:
@lornajane have updated per feedback

lornajane (approved):

lornajane:
Thanks! README changes merged, but I removed the gitignore changes (I realised my feedback wasn't terribly clear, sorry about that) since as @anelook mentioned, it's a local setup detail, not a project feature. Thanks for contributing :tada:

PR #351, opened 2021-11-22
Reword the link text to be meaningful

lornajane:
Previously this generated a warning about two links both labelled
"columns". This removes the duplicate but also tries to reword the text
to be meaningful even if standalone, to help accessibility.


ftisiot (approved):
Good fix, thanks Lorna!

PR #354, opened 2021-11-23
Fix typo in the command to import an external article

anelook:
Fixing the file name as part of the command to run automated article migration


lornajane (approved):
good catch, thanks!

PR #360, opened 2021-11-23
Migrate upgrading-elasticsearch-clients-to-opensearch

anelook:
There was recently added a help article on migration of elasticsearch clients to opensearch.


docs/products/opensearch/howto/upgrade-clients-to-opensearch.rst
```diff
@@ -0,0 +1,57 @@
...
+nodes, the OpenSearch project forked the open source client
+libraries. This means that you can simply switch the libraries without modifying your data
+model, or application logic.
+
+Please refer to `OpenSearch compatibility documentation <https://opensearch.org/docs/latest/clients/index/>`_ as
+the source of truth.
+
+To help you with migration, we provided some `code migration examples <https://github.com/aiven/opensearch-migration-examples>`_ in our repository.
+
+Java and Spring Boot
```
lornajane, line 48: We might need to add Spring and Boot as words to the allow list

lornajane (changes requested):
I find this article a little bit jumbled - I know it's a migration but I think we can get it a more more like a "How To" to fit the devportal content format we usually use with numbered steps:
 - Newer Elasticsearch clients don't work with OpenSearch, so you will need to switch.
 - First, pin your ElasticSearch libraries at <version>
 - Switch from that to the OpenSearch clients (using all the excellent resources we referred to)
 - Migrate from ElasticSearch to OpenSearch with your Aiven service (you can do this before or after switching the clients, and we have an excellent article for you.

Do you think we can refactor?

anelook:
Refactored, let me know if this is closer to what you meant.

lornajane (approved):
Looking good! Thanks :)

PR #382, opened 2021-11-29
Add make commands to generate configuration docs

lornajane:
Based heavily on https://github.com/aprepo/aiven-doc-generator (thanks @aprepo)
Generates include files that can be referred to from the main content,
and regenerated as needed.

### DIFF
--- a/docs/products/kafka/reference/advanced-params.rst
+++ b/docs/products/kafka/reference/advanced-params.rst
@@ -1,0 +1,6 @@
+List of advanced parameters
+============================
+
+When creating or working with an Aiven for Apache Kafka service, several configuration options are available. They are summarized below:
+
+.. include:: /includes/config-kafka.rst
--- a/Makefile
+++ b/Makefile
@@ -51,3 +51,35 @@ index-helpcenter:
 create-feedback-table:
 	python "$(SOURCEDIR)/scripts/create_feedback_table.py" \
 		--pg-url="$(PG_URL)"
+
+# (Re)Generate config listing for a service type
+service-type-config-pg:
+	python "$(SOURCEDIR)/scripts/aiven/service_type_config.py" "pg" > includes/config-pg.rst
+
+service-type-config-kafka:
+	python "$(SOURCEDIR)/scripts/aiven/service_type_config.py" "kafka" > includes/config-kafka.rst
+
+service-type-config-cassandra:
+	python "$(SOURCEDIR)/scripts/aiven/service_type_config.py" "cassandra" > includes/config-cassandra.rst
+
+service-type-config-opensearch:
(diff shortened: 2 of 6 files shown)


docs/products/kafka/reference/advanced-params.rst
```diff
@@ -0,0 +1,6 @@
+List of advanced parameters
+============================
+
+When creating or working with an Aiven for Apache Kafka service, several configuration options are available. They are summarized below:
```
ftisiot, line 4: Should we add a mention on how to use these settings? By adding the `-c` configuration flag during Service creation (https://developer.aiven.io/docs/tools/cli/service.html#avn-service-create) or update (https://developer.aiven.io/docs/tools/cli/service.html#avn-service-update)

ftisiot (changes requested):
Small note about taking these parameters into use, by adding links to the avn cli relevant sections

docs/products/kafka/reference/advanced-params.rst
lornajane: My feeling was that the list was too long, it's also in ALL the tools, the API, terraform, etc. However if we did ever build something like a CLI cheat sheet for a product, we could signpost that from this page. This is reference, so likely that other pages would link to here, if that makes sense.

lornajane:
An excellent point from @dbeech that these "objects" are in fact very rich and structured data payloads in their own right. We should try to drill into the available API response to get more information on the fields inside the objects.

lornajane:
I think this still needs some formatting work (I dislike the bullet points) but I switched from table to sort of definition list style to avoid trying to fit the nested properties into a table cell. Any thoughts, anyone?

anelook:
I also feel that the bullet points are redundant. 

Also, for returned values, would it make sense to add ":" or → to indicate a returned value?

lornajane:
@anelook can you suggest an alternative to the bullet points? I struggled to get a newline that still groups the description with the field itself. I'm also not sure what you mean by a returned value so I think I need an example if possible!

anelook:
Added a new commit where I removed the bullet points completely without replacing with something else. Structure still looks same, just no wasted space.
I added also a symbol " =>" to indicate the returned type.

@lornajane Do you think this looks better? Otherwise we can revert last commit.

lornajane:
I think this works pretty well! The data types aren't returned values, they're a description of what sort of data the field is. Does anyone want to approve and merge? I've just rebased to sort out the conflict in the nav file (we added too much great stuff at once, best problem ever)

anelook (approved):

anelook:
Approved and merging. But if you feel "=>" is not giving any value - I can remove it with a new PR :)

anelook:
Oh, wait, can't merge, @ftisiot  has to approve it!

ftisiot (approved):
Looks Good!

PR #391, opened 2021-12-01
Using Conduktor with Apache Kafka

lornajane:

Add some instructions for using Conduktor (a desktop application for Kafka) with an Aiven service. Their docs are pretty good but after following up on a support ticket, I found there was an extra step with a CA cert so I thought we could include it for any of our customers wanting to use this tool.


docs/products/kafka/howto/kafka-conduktor.rst
```diff
@@ -0,0 +1,31 @@
+Connect to Apache Kafka with Conduktor
+======================================
+
+`Conduktor <https://www.conduktor.io/>`_ is a friendly user interface for Apache Kafka, and it works well with Aiven. In fact, there is built-in support for setting up the connection. You will need to add the CA certificate for each of your Aiven projects to Conduktor before you can connect, this is outlined in the steps below.
+
+1. Visit the **Service overview** page for your Aiven for Apache Kafka service (the :doc:`../getting-started` page is a good place for more information about creating a new service if you don't have one already).
+
+2. *Only required if this is the first time you are connecting Conduktor to an Apache Kafka service in this Aiven project*. You will first need to add the CA certificate so that Conduktor knows to trust connections using this certificate.
```
ftisiot, line 8: I tried to avoid step 2 and the connection works even without it, should we remove?

ftisiot (changes requested):
Not sure why point #2 is needed, I tried to connect without it and it still works... perhaps we can skip it?

docs/products/kafka/howto/kafka-conduktor.rst
lornajane: I wrote this whole article because I saw a support request come in, and I needed step 2 to get it working on my system. I tried it without (like the Conduktor docs) and it gave me a Java SSL error. Maybe Step 2 should be a troubleshooting tip instead? It definitely is sometimes needed.

ftisiot (approved):
Good work!

PR #392, opened 2021-12-03
Kafka Topic Partition Info: Adding `--project` to the example shell script

z0r1k:
Adding project parameter so we won't assume user has just one project and use it as default.

Usage: `./script.sh <PROJ> <SVC>`

### DIFF
--- a/docs/products/kafka/howto/get-topic-partition-details.rst
+++ b/docs/products/kafka/howto/get-topic-partition-details.rst
@@ -29,14 +29,15 @@ For example, this bash script, with a help of ``jq`` utility, lists topics and t
 
 .. code:: bash
 
-    #!/bin/bash
-    serv=YOUR-KAFKA-SERVICE-NAME
-    cloud=$(avn service get $serv --json | jq -r '.cloud_name')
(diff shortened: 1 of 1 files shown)


anelook (approved):
Thank you for improving the example. Merging the change, but is it mandatory to have the project name? Or only recommendable?

lornajane:
(for information only; no action needed) I'd say it's neither default nor recommended. I don't have any data to back this up but I think most people `avn switch project next-project` so that they don't need to add `--project myproject` to every command. This updated example is valid though, and doesn't assume that the user does commonly use the CLI and knows to select a project.

z0r1k:
I guess it is correct that most of the people would have just one project and would switch in to it before, but to be frank I don't know should we just assume people actually doing it? My default would be to use `--project` all the time and then have a note like use `avn project switch` if you are tired of `--project` 🤷

z0r1k:
> Thank you for improving the example. Merging the change, but is it mandatory to have the project name? Or only recommendable?

If arguments are not supplied to the scripts it will fallback to `MY-PROJECT-NAME` and `MY-SERVICE-NAME`.

PR #398, opened 2021-12-07
Migrate Kafka authentication types

### LINKED ISSUE #353: Migrate kafka-authentication-types
Migrate https://help.aiven.io/en/articles/4331547-kafka-authentication-types

anelook:
Fixes #353

### DIFF
--- a/docs/products/kafka/concepts/auth-types.rst
+++ b/docs/products/kafka/concepts/auth-types.rst
@@ -35,7 +35,7 @@ allows alternative login methods for your service. For example, if you
 use Active Directory for authentication, SASL supports a Kerberos login
 method to provide access.
 
-You can use SASL/PLAIN and SASL/SCRAM with **Aiven for Apache Kafka**.
+We support SASL/PLAIN and SASL/SCRAM with **Aiven for Apache Kafka**.
 
 SASL/PLAIN
 ~~~~~~~~~~
@@ -68,13 +68,7 @@ then used to create an "identity" that holds:
 
 This identity is then by default stored in Zookeeper.
 
-Enabling different authentication methods in Aiven
---------------------------------------------------
+Enable SASL authentication
+---------------------------
 
-**TLS** is enabled by default for your **Aiven for Apache Kafka** service, if you want to additionally add **SASL** follow these steps:
-
-#. Log in to the Aiven web console and select your **Aiven for Apache Kafka** service.
-
-#. On the *Overview* page, scroll down to *Advanced configuration* .
-
-#. Switch on **kafka.authentication_methods.sasl** to enable SASL authentication.
+Follow the steps from our article :doc:`to enable SASL authentication <../howto/kafka-sasl-auth>`
--- a/docs/products/kafka/concepts/auth-types.rst
+++ b/docs/products/kafka/concepts/auth-types.rst
@@ -14,8 +14,7 @@ Authority (for example, `letsencrypt.org <http://letsencrypt.org>`_ ) for your d
 With this certificate and the right technical setup, you can use
 your domain to encrypt the traffic to your service.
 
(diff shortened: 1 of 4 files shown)


docs/products/kafka/concepts/auth-types.rst
```diff
@@ -0,0 +1,73 @@
...
+method to provide access.
+
+We support SASL/PLAIN and SASL/SCRAM with **Aiven for Apache Kafka**.
+
+SASL/PLAIN
+~~~~~~~~~~
+
+**PLAIN** relies on a combination of username
+and password to log in over a TLS connection, meaning that your traffic
+is encrypted. You should not use SASL/PLAIN without TLS, because in this case
```
ftisiot, line 44: should this be a warning?

docs/products/kafka/concepts/auth-types.rst
```diff
@@ -0,0 +1,73 @@
+Authentication types
+======================
+
+It is strongly recommended to use modern encryption protocols to protect data in transit sent to and from Apache Kafka.
+You can use one of the available options in **Aiven for Apache Kafka** services:
```
ftisiot, line 5: should we remove the `:` at the end?

ftisiot (changes requested):
Small changes required. I'm not a security expert, we should ask an SA to review as well. possibly Dave Beech who wrote something about this

docs/products/kafka/concepts/auth-types.rst
```diff
@@ -0,0 +1,73 @@
...
+
+#. **TLS authentication** : your Apache Kafka client validates the certificate
+   for your Apache Kafka broker and your broker validates the certificate
+   for your client.
+
+
+Simple Authentication and Security Layer
+----------------------------------------
+
+**Simple Authentication and Security Layer (SASL)** acts as a layer that
```
dbeech, line 32: I don't think we should mention Kerberos here - it could be misleading since we do not support that as an authentication type (SASL/GSSAPI).

docs/products/kafka/concepts/auth-types.rst
dbeech: Having a warning is fine but there's no risk here because Aiven does not support non-TLS connections at all. (i.e. SASL_SSL protocol is supported, SASL_PLAINTEXT is not).

docs/products/kafka/concepts/auth-types.rst
anelook: Removed

docs/products/kafka/concepts/auth-types.rst
anelook: Rephrased a bit. Could you check @dbeech if this reflects the reality?
 
I actually got here confused as well, because SASL/PLAIN  without TLS doesn't make any sense to me.  But after checking with Christopher Gwilliams we do have support for  SASL/PLAIN. And the content is based on the [help article ](https://help.aiven.io/en/articles/4331547-kafka-authentication-types)

dbeech: It is very confusing. There are two totally separate but similarly named things:
- The protocol, which can be `SASL_SSL` or `SASL_PLAINTEXT` (this is the part that dictates whether the communication link is encrypted)
- The authentication type, which can be SASL/PLAIN or SASL/SCRAM (this is the part that dictates how the credentials are handled, i.e. are we sending a plaintext password or an encrypted one)

When using SASL on Aiven, the user has no choice over the protocol - it's always `SASL_SSL`. They do get to choose the auth type `PLAIN` vs `SCRAM`, but even in the case of `PLAIN`, the "plain" creds are sent over an encrypted link, so the risk is somewhat mitigated.

lornajane (approved):
This is good to go! Thanks @anelook for the migration @ftisiot for the review and @dbeech for the tech support :)

PR #402, opened 2021-12-08
Migrate topic log compaction

### LINKED ISSUE #400: Migrate Topic Log Compaction

https://help.aiven.io/en/articles/3995429-topic-log-compaction

anelook:
Fixes #400


docs/products/kafka/concepts/log-compaction.rst
```diff
@@ -1,7 +1,169 @@
...
+    - K1
+    - 8
+  * - 6
+    - K4
+    - 14
+
+
+When a log is compacted it consists of head and tail, where head is the traditional  Apache Kafka log and new records get appended to the end of it.  Apache Kafka ensures that the records in the tail consist only of unique keys because only the tail section is scanned during compaction process while head section may contain duplicate keys.
+
+.. note:: Log compaction occurs inside a partition and if two records with the same key land in different partitions, they will not be compacted together.
```
ftisiot, line 98: Should we state that "this usually doesn't happen since the record key is used to select the partition, but for custom message routing this might be an issue"

docs/products/kafka/concepts/log-compaction.rst
```diff
@@ -1,7 +1,169 @@
...
+
+Segments
+~~~~~~~~
+
+If we look "under the hood" of the partition we will find that  Apache Kafka divides the partitions into **segments** which are files (name ends with ``.log`` ) stored on a file system for each partition. A segment file is part of the partition. As the log cleaner cleans log partition segments, the segments get swapped into the log partition immediately replacing the older segments.
```
ftisiot, line 103: I'm a bit lost here, since I don't understand the sentence "As the log cleaner cleans log partition segments, the segments get swapped into the log partition immediately replacing the older segments." What is happening? what is the log cleaner doing?

docs/products/kafka/concepts/log-compaction.rst
```diff
@@ -1,7 +1,169 @@
...
+
+The first offset of the segment, **base offset,** corresponds to the file name of the segment. The last segment in the partition is called an **active segment** and it is the only segment to which new messages are appended to. **During the cleaning process, an active segment is excluded and you may see duplicate records.** The user-age partition below contains a segment 04.log that has not yet been compacted, hence you will see duplicate records.
```
ftisiot, line 105: the segment 04.log should be `segment 04.log`?

ftisiot: Should the phrase **During the cleaning process, an active segment is excluded and you may see duplicate records.** be at the end?

docs/products/kafka/concepts/log-compaction.rst
```diff
@@ -1,7 +1,169 @@
...
+  :stub-columns: 1
+  :align: left
+
+  * - Offset
+    - Key
+    - Value
+  * - 7
+    - K4
+    - 9
+
```
ftisiot, line 163: Should we show the situation before and after the clean process?

docs/products/kafka/howto/configure-log-cleaner.rst
```diff
@@ -0,0 +1,30 @@
...
+=====================================
+
+By default the log cleaner is enabled, but the log compaction is disabled. To enable log compaction, follow these steps:
+
+#. Log into the `Aiven web console <https://console.aiven.io/>`_ and select your Aiven for Apache Kafka service.
+#. Scroll down to *Advanced configuration* and click on **Add configuration option**.
+#. Find ``log.cleanup.policy`` in the list and select it.
+#. Set the value to ``compact``.
+
+This change will affect all topics in the cluster that do not have a configuration override in place.
```
ftisiot, line 11: I believe this needs to be a warning, since it can affect existing topics and create disasters

docs/products/kafka/howto/configure-log-cleaner.rst
```diff
@@ -0,0 +1,30 @@
...
+
+Control the frequency and delay
```
ftisiot, line 13: should the title state that is **log cleaning frequency and delay**?

docs/products/kafka/howto/configure-log-cleaner.rst
```diff
@@ -0,0 +1,30 @@
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Before the cleaning begins, the cleaner thread will inspect the logs to find those with highest **dirty ratio** which is calculated as the ratio of the number of bytes in the head vs the total number of bytes in the log (tail + head). The default ratio value is 50%, but you can change it by change value of the property ``kafka.log_cleaner_min_cleanable_ratio`` from the *Advanced configuration* list.
```
ftisiot, line 16: Why is this ratio used? Adding a bit more explanation telling that this ratio allows to understand how many duplicated keys we have in a topic, and those will need to be compacted?

ftisiot (changes requested):
Good starting point, but very complex topic. I believe we need to explain a bit more about the process since now it seems a bit obscure to me. Probably a couple diagrams could be used Happy to help further

docs/products/kafka/concepts/log-compaction.rst
```diff
@@ -1,7 +1,169 @@
...
+   1002 -> "221B Baker Street"
+   ...
+   1001 -> "21 Jump St"
+   ...
+   1001 -> "Paper St"
+
+
+We have three different options on how long to retain the messages:
+
+* infinite log retention: all changes to user's address would be maintained in the logs. This would lead to the log growing in size without a bound.
```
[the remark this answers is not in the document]
anelook: changed. Will link once published

docs/products/kafka/concepts/log-compaction.rst
```diff
@@ -1,7 +1,169 @@
...
+
+Before compaction
+*****************
+
+.. list-table::
+  :header-rows: 1
+  :stub-columns: 1
+  :align: left
+
+  * - Offset
```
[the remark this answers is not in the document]
anelook: Which one do you refer to?

docs/products/kafka/concepts/log-compaction.rst
```diff
@@ -1,7 +1,169 @@
...
+    - 7
+  * - 5
+    - K1
+    - 8
+  * - 6
+    - K4
+    - 14
+
+
+When a log is compacted it consists of head and tail, where head is the traditional  Apache Kafka log and new records get appended to the end of it.  Apache Kafka ensures that the records in the tail consist only of unique keys because only the tail section is scanned during compaction process while head section may contain duplicate keys.
```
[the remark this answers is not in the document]
anelook: I must admit, I'm a bit lost here myself. We have another diagram in the original article, but it confuses me even more. Do you know more about this topic or should I research a bit deeper?

docs/products/kafka/concepts/log-compaction.rst
anelook: Added

docs/products/kafka/concepts/log-compaction.rst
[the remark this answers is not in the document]
ftisiot: The example above talks about addresses, this one only about key and value, if we could reuse the same addresses example it will be clearer IMHO

docs/products/kafka/concepts/log-compaction.rst
```diff
@@ -1,7 +1,169 @@
 Log compaction
 ==============
 
-One way to reduce the disk space requirements in Kafka is to use **log compaction**. This operation retains only the newest record for each key on a topic, regardless of whether the retention period of the message has expired. Depending on the application, this can significantly reduce the amount of storage required for the topic.
+One way to reduce the disk space requirements in Apache Kafka is to use **log compaction**. This operation retains only the newest record for each key on a topic, regardless of whether the retention period of the message has expired or not. Depending on the application, this can significantly reduce the amount of storage required for the topic.
 
-To make use of log compaction, all messages sent to the topic must have an explicit key. You can enable log compaction under the **Topics** section of the web console by setting the **Cleanup policy** to "compact".
+To make use of log compaction, all messages sent to the topic must have an explicit key. To enable log compaction follow the steps described in :doc:`how to configure log cleaner <../howto/configure-log-cleaner>`.
```
laysauchoa, line 6:
```suggestion
To make use of log compaction, all messages sent to the topic must have an explicit key. To enable log compaction, follow the steps described in :doc:`how to configure log cleaner <../howto/configure-log-cleaner>`.
```

ftisiot:
@anelook could you please review?

anelook:
Thank you @ftisiot, your changes make the article easier to read and understand.

ftisiot:
Now that we got the articles working together, @lornajane could you please check the end result?

docs/products/kafka/concepts/log-compaction.rst
```diff
@@ -1,7 +1,194 @@
...
+One way to reduce the disk space requirements in Apache Kafka is to use **compacted topics**. This methodology retains only the newest record for each key on a topic, regardless of whether the retention period of the message has expired or not. Depending on the application, this can significantly reduce the amount of storage required for the topic.
 
-To make use of log compaction, all messages sent to the topic must have an explicit key. You can enable log compaction under the **Topics** section of the web console by setting the **Cleanup policy** to "compact".
+To make use of log compaction, all messages sent to the topic must have an explicit key. To enable log compaction, follow the steps described in :doc:`how to configure log cleaner <../howto/configure-log-cleaner>`.
 
+
+How compacted topics work
+-------------------------
+
+An Apache Kafka topic represents a continuous stream of messages that typically get discarded after the message reaches a certain period of time or size. However for certain use cases we just need the most recent value for a certain key.
```
laysauchoa, line 12:
```suggestion
An Apache Kafka topic represents a continuous stream of messages that typically get discarded after the message reaches a certain period of time or size. However, for certain use cases we just need the most recent value for a certain key.
```

docs/products/kafka/concepts/log-compaction.rst
```diff
@@ -1,7 +1,194 @@
...
+
+For example, if we have a topic containing user's home address, on every update a message sent using ``user_id`` as primary key and home address as the value:
```
laysauchoa, line 14:
```suggestion
For example, if there is a topic containing a user's home address, on every update, a message is sent using ``user_id`` as the primary key and home address as the value:
```

docs/products/kafka/concepts/log-compaction.rst
```diff
@@ -1,7 +1,194 @@
...
+
+   1001 -> "4 Privet Dr"
+   1002 -> "221B Baker Street"
+   1003 -> "Milkman Road"
+   1002 -> "21 Jump St"
+   1001 -> "Paper St"
+   1001 -> "Paper Road 21"
+
+
+We have three different options on how long to retain the messages:
```
laysauchoa, line 26: Maybe instead of "we" answer something like "where" or "what has it"?

docs/products/kafka/concepts/log-compaction.rst
```diff
@@ -1,7 +1,194 @@
...
+
+* **infinite message retention**: all changes to user's address would be maintained in the logs. This would lead to the log growing in size without a bound. This option involves a risk to outgrow the disk capacity.
```
laysauchoa, line 28:
```suggestion
* **infinite message retention**: all changes to user's address would be maintained in the logs. This can lead to the log growing in size without a bound. This option involves the risk of outgrowing the disk capacity.
```

docs/products/kafka/concepts/log-compaction.rst
```diff
@@ -1,7 +1,194 @@
...
+* **simple message retention**: older records would be deleted after they reach a certain age or size.
```
laysauchoa, line 29:
```suggestion
* **simple message retention**: older records are deleted after they reach a certain age or size.
```

docs/products/kafka/concepts/log-compaction.rst
```diff
@@ -1,7 +1,194 @@
...
+* **compacted topic**: only latest version of the key's value is kept. In the example above, only the current address for a specific user would be kept.
```
laysauchoa, line 30:
```suggestion
* **compacted topic**: only latest version of the key's value is kept. In the example above, only the current address for a specific user are kept.
```

docs/products/kafka/concepts/log-compaction.rst
```diff
@@ -1,7 +1,194 @@
...
+
+With compacted topics, Apache Kafka removes from the topic any records for which there is a newer version (based on the record key) is available in the partition. This retention policy can be set per-topic, so a single cluster can have some topics where retention is enforced by size or time and other topics where retention is enforced by compaction.
```
laysauchoa, line 32:
```suggestion
With compacted topics, Apache Kafka removes any records from the topics for which there is a newer version (based on the record key) available in the partition. This retention policy can be set per topic, so a single cluster can have some topics where retention is enforced by size or time and other topics where retention is enforced by compaction.
```

docs/products/kafka/concepts/log-compaction.rst
```diff
@@ -1,7 +1,194 @@
...
+
+Compacted topic example
+-----------------------
+
+To understand better how compaction works we will look at a partition of a compacted topic before and after compaction has been applied.
```
laysauchoa, line 37:
```suggestion
To understand better how compaction works, we will look at a partition of a compacted topic before and after compaction has been applied.
```

docs/products/kafka/concepts/log-compaction.rst
```diff
@@ -1,7 +1,194 @@
...
+    - 1002
+    - 21 Jump St
+  * - 5
+    - 1001
+    - Paper St
+  * - 6
+    - 1001
+    - Paper Road 21
+ 
+You can notice that there are some records with duplicate keys (``1001`` and ``1002``), with the records having offset ``4``, ``5`` and ``6`` being the addresses updates. When applying compaction, we only keep records with the latest offset (newest values) and the older ones get discarded. The end result is the following:
```
laysauchoa, line 68:
```suggestion
You can notice that there are some records with duplicate keys (``1001`` and ``1002``), with the records having offset ``4``, ``5``, and ``6`` being the addresses updates. When applying compaction, we only keep records with the latest offset (newest values) and the older ones get discarded. The end result is the following:
```

clear sentence 👍

docs/products/kafka/concepts/log-compaction.rst
```diff
@@ -1,7 +1,194 @@
...
+  * - 6
+    - 1001
+    - Paper Road 21
+
+Compacted topic details
+-----------------------
+
+A compacted topic consists of an head and a tail:
+
+* the **head** is a traditional Apache Kafka topic where new records are appended. The head can therefore contain duplicated keys.
```
laysauchoa, line 93:
```suggestion
* the **head** is a traditional Apache Kafka topic where new records are appended. Therefore, the head can contain duplicated keys.
```

docs/products/kafka/concepts/log-compaction.rst
```diff
@@ -1,7 +1,194 @@
...
+* the **tail** contains one record per key. Apache Kafka compaction ensures that keys are unique in the tail. 
+
+.. Warning:: 
+
+  The compaction occurs **per partition**: if two records having the same key land in different partitions, they will not be compacted.
```
laysauchoa, line 98:
```suggestion
  The compaction occurs **per partition**: if two records have the same key land in different partitions, they will not be compacted.
```

docs/products/kafka/concepts/log-compaction.rst
```diff
@@ -1,7 +1,194 @@
...
+    - 1001 :badge:`delete,cls=badge-danger badge-pill`
+    - 4 Privet Dr
+  * - 2
+    - 1002 :badge:`delete,cls=badge-danger badge-pill`
+    - 221B Baker Street
+  * - 3
+    - 1003
+    - Milkman Road
+
+Lastly the records in the offset map are added in the tail.
```
laysauchoa, line 176:
```suggestion
Lastly, the records in the offset map are added in the tail. 
```

docs/products/kafka/concepts/partition-segments.rst
```diff
@@ -0,0 +1,65 @@
+Partition segments
+==================
+
+Apache Kafka divides topics partition data into **segment** files (with ``.log`` suffix) stored on the file system. Each segment file is named using the offset of the first message (a.k.a. **base offset**) contained. E.g. the segment file ``04.log`` contains the message with offset ``4`` as first entry.
```
laysauchoa, line 4:
```suggestion
Apache Kafka divides topics partition data into **segment** files (with ``.log`` suffix) stored on the file system. Each segment file is named using the offset of the first message (a.k.a. **base offset**) contained. For example, the segment file ``04.log`` contains the message with offset ``4`` as first entry. 
```
both are right, I just think looks better a non abbreviated version here as it is the beginning of a sentence but this is IMHO

docs/products/kafka/concepts/partition-segments.rst
```diff
@@ -0,0 +1,65 @@
+Partition segments
+==================
+
+Apache Kafka divides topics partition data into **segment** files (with ``.log`` suffix) stored on the file system. Each segment file is named using the offset of the first message (a.k.a. **base offset**) contained. E.g. the segment file ``04.log`` contains the message with offset ``4`` as first entry. 
+
+The last segment in the partition is called the **active segment** and it is the only segment to which new messages are appended to.
+
+In the example above you a topic partition is divided into multiple segments, with ``06.log`` being the active one:
```
laysauchoa, line 8:
```suggestion
In the example above the topic partition is divided into multiple segments, with ``06.log`` being the active one:
```
do we need to specify which topic partition is? I think it was meant to be **your** in any case.

docs/products/kafka/concepts/partition-segments.rst
```diff
@@ -0,0 +1,65 @@
...
+  :align: left
+
+  * - Offset
+    - Key
+    - Value
+  * - 6
+    - 1001
+    - Paper Road 21
+
+When the segment file reaches a certain size of age, Apache Kafka will create a new segment file. This can be controlled by the following settings:
```
laysauchoa, line 61: off topic: but can we know which size of age is that? I am not super familiar with kafka but do we know if `size of age` is a particular term on kafka? size x age in general so this why I am curious :)

docs/products/kafka/concepts/partition-segments.rst
```diff
@@ -0,0 +1,65 @@
...
+
+-  ``segment.bytes`` : create a new segment when current segment becomes greater than this size. This setting can be set during topic creation and defaults to 1GB.
```
laysauchoa, line 63: As we describing of all those segments as a segment. Would make sense to ignore their inner name and just treat as a third person?
```suggestion
-  ``segment.bytes`` : creates a new segment when current segment becomes greater than this size. This setting can be set during topic creation and defaults to 1GB.
```

docs/products/kafka/howto/configure-log-cleaner.rst
```diff
@@ -0,0 +1,34 @@
+Configure log cleaner for topic compaction
+==========================================
+
+The log cleaner is the process ensuring only the most recent value for a certain message key is kept within a partition for :doc:`compacted topics <../concepts/log-compaction>`.  In Aiven for Apache Kafka the log cleaner is enabled, but the log compaction is disabled by default. To enable log compaction, follow these steps:
```
laysauchoa, line 4:
```suggestion
The log cleaner is the process ensuring only the most recent value for a certain message key is kept within a partition for :doc:`compacted topics <../concepts/log-compaction>`.  In Aiven for Apache Kafka, the **log cleaner** is enabled, but the **log compaction** is disabled by default. To enable log compaction, follow these steps:
```

docs/products/kafka/howto/configure-log-cleaner.rst
```diff
@@ -0,0 +1,34 @@
...
+
+#. Log into the `Aiven web console <https://console.aiven.io/>`_ and select your Aiven for Apache Kafka service.
```
laysauchoa, line 6: I like that those steps is consistent across the files. I mean how the format is. 💯

docs/products/kafka/howto/configure-log-cleaner.rst
```diff
@@ -0,0 +1,34 @@
...
+#. Scroll down to *Advanced configuration* and click on **Add configuration option**.
+#. Find ``log.cleanup.policy`` in the list and select it.
+#. Set the value to ``compact``.
+
+.. warning:: This change will affect all topics in the cluster that do not have a configuration override in place.
+
+Log cleaning frequency and delay
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Before the cleaning begins, the cleaner thread will inspect the logs to find those with highest **dirty ratio** calculated as the number of bytes in the head vs the total number of bytes in the log (tail + head); you can read more about head and tail definition in the :doc:`compacted topic documentation <../concepts/log-compaction>`. The ratio provides an estimate of how many duplicated keys are present in a topic, and therefore need to be compacted.
```
laysauchoa, line 16:
```suggestion
Before the cleaning begins, the cleaner thread will inspect the logs to find those with highest **dirty ratio** calculated as the number of bytes in the head vs the total number of bytes in the log (tail + head); you can read more about head and tail definition in the :doc:`compacted topic documentation <../concepts/log-compaction>`. The ratio provides an estimation of how many duplicated keys are present in a topic; therefore, needs to be compacted.
```

laysauchoa, line 16: thinking about one thing versus another, do you think would be better to have a shorter version and with bold to make it easy to find: for example, **kafka vs rabbitmq** but for this piece here: 

> the number of bytes in the head vs the total number of bytes in the log (tail + head)

I do not have a suggestion but maybe if you have it, feel free to shorter

docs/products/kafka/howto/configure-log-cleaner.rst
```diff
@@ -0,0 +1,34 @@
...
+
+.. Tip::
+
+    For the log cleaner to start compacting a topic, the dirty ratio needs to be bigger than a threshold set to 50% by default. You can change this value by modifying the property ``kafka.log_cleaner_min_cleanable_ratio`` in the *Advanced configuration* section of the service overview.
```
laysauchoa, line 20:
```suggestion
    For the log cleaner to start compacting a topic, the dirty ratio needs to be bigger than the 50% default threshold. You can change this value by modifying the property ``kafka.log_cleaner_min_cleanable_ratio`` in the *Advanced configuration* section of the service overview.
```

docs/products/kafka/howto/configure-log-cleaner.rst
```diff
@@ -0,0 +1,34 @@
...
+
+The log cleaner can be configured to leave some amount of not compacted "head" of the log by setting compaction time lag. You can achieve this by setting two additional properties from the  *Advanced configuration*:
```
laysauchoa, line 22:
```suggestion
The log cleaner can be configured to leave some amount of not compacted "head" of the log by setting compaction time lag. You can achieve this by setting two additional properties from the *Advanced configuration*:
```

docs/products/kafka/howto/configure-log-cleaner.rst
```diff
@@ -0,0 +1,34 @@
...
+
+-  ``log.cleaner.min.compaction.lag.ms`` : setting to a value greater than 0 will prevent log cleaner from compacting messages with an age newer than a minimum message age. Allows to delay compacting records.
```
laysauchoa, line 24: does it make sense to connect those two sentences with an **and**? What does allows to delay compacting records: the process itself or the setting the value greater than 0?

docs/products/kafka/howto/configure-log-cleaner.rst
```diff
@@ -0,0 +1,34 @@
...
+
+-  ``log.cleaner.max.compaction.lag.ms`` : the maximum amount of time message will remain not compacted. Please note, that exact time of compression can be later when comparing to the maximum compaction lag, depending on the availability log cleaner threads and actual compaction time.
```
laysauchoa, line 26:
```suggestion
-  ``log.cleaner.max.compaction.lag.ms`` : the maximum amount of time a message will remain not compacted. Please note, that exact time of compression can be later when comparing to the maximum compaction lag, depending on the availability log cleaner threads and actual compaction time.
```
the availability **of** log cleaner threads? or is this a term itself?

docs/products/kafka/howto/configure-log-cleaner.rst
```diff
@@ -0,0 +1,34 @@
...
+
+Tombstone records
+~~~~~~~~~~~~~~~~~
+
+During the cleanup process, log cleaner threads also removes records that have a null value, also known as **tombstone** records. These records can be delayed from being deleted by configuring ``delete.retention.ms`` for the compacted topic.
+
+Consumers are able to read all tombstone messages as long as they reach the head of the topic before the period defined in ``delete.retention.ms`` (default 24 hours) is passed.
```
laysauchoa, line 33:
```suggestion
Consumers are able to read all tombstone messages as long as they reach the head of the topic before the period defined in ``delete.retention.ms`` (default: 24 hours) is passed.
```

laysauchoa (commented):
I like this article, it is very clear and the topic is more complex. I think I understood and my suggestions are only small improvements on the meaning and grammar, basically.

laysauchoa:
I added my review @anelook and @ftisiot, feel free to use it, answer or if you do not think it suits, you can also close because some are related to different ways of expressing the same thing :)

docs/products/kafka/concepts/partition-segments.rst
ftisiot: it was meant to be `size OR age`,thanks

docs/products/kafka/concepts/log-compaction.rst
```diff
@@ -1,7 +1,194 @@
...
+    - 1003
+    - Milkman Road
+  * - 4
+    - 1002
+    - 21 Jump St
+  * - 6
+    - 1001
+    - Paper Road 21
+
+Compacted topic details
```
lornajane, line 88: I'm a bit confused about this part. What can the user do with this information, can they connect to the head if they want to, or do they need to specify which part of the topic they want to connect to? Do we do this differently at Aiven than would be the case for other Kafka instances?

docs/products/kafka/concepts/log-compaction.rst
```diff
@@ -1,7 +1,194 @@
...
+-----------------------
+
+A compacted topic consists of an head and a tail:
+
+* the **head** is a traditional Apache Kafka topic where new records are appended. Therefore, the head can contain duplicated keys.
+* the **tail** contains one record per key. Apache Kafka compaction ensures that keys are unique in the tail. 
+
+.. Warning::
```
lornajane, line 96: This seems really useful, can we put it a bit further up the post? Getting the keys right does seem fairly fundamental!

docs/products/kafka/howto/configure-log-cleaner.rst
```diff
@@ -0,0 +1,38 @@
...
+==========================================
+
+The log cleaner is the process ensuring only the most recent value for a certain message key is kept within a partition for :doc:`compacted topics <../concepts/log-compaction>`.  In Aiven for Apache Kafka the log cleaner is enabled, but the **log compaction** is disabled by default. To enable log compaction, follow these steps:
+
+#. Log into the `Aiven web console <https://console.aiven.io/>`_ and select your Aiven for Apache Kafka service.
+#. Scroll down to *Advanced configuration* and click on **Add configuration option**.
+#. Find ``log.cleanup.policy`` in the list and select it.
+#. Set the value to ``compact``.
+
+.. warning:: This change will affect all topics in the cluster that do not have a configuration override in place.
```
lornajane, line 11: Can I configure some or all of these options per topic? I am not sure if that's what you mean, that this will only affect the topics that don't have specific configuration. Can we make it clearer which things can be configured at topic level and where to find the options for that?

lornajane (commented):
I had too many questions to be able to approve this! There's a lot of detail that I'm not sure how it's practically useful, so why it's here? And some other things that we could clarify.

In general, thanks for tackling this very in-depth topic. This is definitely getting close!

docs/products/kafka/concepts/log-compaction.rst
anelook: What can the user do with this information, can they connect to the head if they want to, or do they need to specify which part of the topic they want to connect to? 

- they can connect to any of the points in the stream, but this is not really related to the compaction topic.

Do we do this differently at Aiven than would be the case for other Kafka instances?

- It is the same

@ftisiot , could you correct if I'm wrong?

docs/products/kafka/concepts/log-compaction.rst
anelook: I've moved the warning to the end of the article, this is what you meant?

docs/products/kafka/howto/configure-log-cleaner.rst
anelook: Rephrased and added more information on global and per-topic configuration

anelook:
This information might be not actively practical, but as a concept it explains what happens when you select compaction configuration options. So I believe it is useful for the customers.

ftisiot:
Not sure why I can't further comment on the above

> What can the user do with this information, can they connect to the head if they want to, or do they need to specify which part of the topic they want to connect to?
 
 they can connect to any of the points in the stream, but this is not really related to the compaction topic.
 
Do we do this differently at Aiven than would be the case for other Kafka instances?
 
It is the same

I believe the main reason to explain the difference between tail and head is due to the fact that in the head there could be duplicated entries (not yet compacted). The topic itself can be used as any other topic and there is no difference at all in the way we handle it. Since we had this in our help article, and I believe the information is valuable, we kept it as separated and reformatted article

ftisiot:
Not sure what's the next step here is, In my opinion keeping both docs is worth also seen the traffic (the help page is ~60th position for traffic), what are your thoughts @lornajane and @anelook ?

lornajane:
If both of you are happy, then I'm happy! Let's publish if so, and we can iterate later if we want to.

anelook:
I agree, let's publish this.

ftisiot (approved):
LGTM!

PR #410, opened 2021-12-14
Updated OpenSearch SQL and alerting plugin entries

diawarr:
Titles and links for SQL and alerting plugins updated as they are now included in the OpenSearch repo (previously released from Open Distro for Elasticsearch).


anelook (approved):
Thanks! Looks great

PR #416, opened 2021-12-17
Trivial fixes in PG failover doc

joukost:

A few DSN->DNS and small clarification about hobbyist/startup
restoration - "since latest available backup" sounds like one might lose
a day's worth of data which is not true, as pghoard would periodically
upload WALs to be replayed on top of that. Could be clarified further
but it's described elsewhere in more detail anyway.


ftisiot (approved):
Thanks! Good eyesight!

PR #420, opened 2021-12-20
Added flex-wrap attribute to search page CSS

diawarr:
- Adding "flex-wrap: wrap;" to the CSS for the search page should fix pagination issues with a high number of pages in search results

### DIFF
--- a/_templates/search.html
+++ b/_templates/search.html
@@ -31,6 +31,7 @@
     list-style: none;
     display: flex;
     justify-content: center;
+    flex-wrap: wrap;
   }
 
   #search-results .pagination a {


lornajane:
@diawarr just checking in on this: it's still in draft, is that intentional? And can you remind me what the "pagination issues" are in this case?

diawarr:
@lornajane - yep, it was set to draft because this was more of a proposal for review, and the issue is that the page navigation on the search results page is currently always displayed on one line, so if there are a high number of pages for a specific search, some of the links are hidden (a somewhat extreme example, but this highlights the issue quite well: https://developer.aiven.io/search.html?q=aiven) ;)

anelook:
I agree that the change improves user experience. Tested in Chrome, Firefox and Safari.

anelook (approved):

PR #421, opened 2021-12-21
Add the generated config docs for MySQL

lornajane:

We were missing the actual config options file from the MySQL PR - this adds the file.


diawarr (approved):
Yep, works fine now :)

PR #429, opened 2021-12-22
add/redis: ssl connectivity

laysauchoa:
### Migration of Redis content.

Original URL:
https://help.aiven.io/en/articles/489559-redis-ssl-connectivity

Dev Portal:
https://developer.aiven.io/docs/products/redis/concepts/howto/manage-ssl-connectivity.html

Add relevant information about how to run the stunnel process and relevant documentation links.
I have ran the stunnel process myself and seems to be fine. 

Alternatively, this can be saved on tree as: `how to setup ssl  connectivity`

Fix: https://github.com/aiven/devportal/issues/433

### DIFF
--- a/.github/styles/Vocab/Docs/accept.txt
+++ b/.github/styles/Vocab/Docs/accept.txt
@@ -88,7 +88,8 @@ Postman
 Protobuf
 psql
 Python
-[Rr]edis
+Redis
+Redli
 refcard
 reindexing
 Sakila
@@ -121,3 +122,4 @@ Workbench
 Zabbix
 Zipkin
 ZooKeeper
+Stunnel
--- a/.github/styles/Vocab/Docs/accept.txt
+++ b/.github/styles/Vocab/Docs/accept.txt
@@ -88,7 +88,7 @@ Postman
 Protobuf
 psql
 Python
-Redis
+[Rr]edis
 refcard
 reindexing
 Sakila
--- a/docs/products/redis/concepts/ssl-connectivity.rst
+++ b/docs/products/redis/concepts/ssl-connectivity.rst
@@ -11,25 +11,25 @@ Aiven for Redis uses SSL encrypted connections by default. This is shown by the 
 .. Tip::
     You can find the ``Service URI`` on `Aiven console <https://console.aiven.io/>`_.
 
-Since **Redis 6**, the redis-cli tool itself supports SSL connections; therefore, you can connect directly to your service using::
+Since **Redis 6**, the ``redis-cli`` tool itself supports SSL connections; therefore, you can connect directly to your service using::
 
     redis-cli -u rediss://username:password@host:port
 
-Alternatively, you can use the third-party `redli tool <https://github.com/IBM-Cloud/redli>`_::
+Alternatively, you can use the third-party `Redli tool <https://github.com/IBM-Cloud/redli>`_::
 
     redli -u rediss://username:password@host:port
 
 
 Not every Redis client supports SSL-encrypted connections.
-In such cases, you would have to turn off SSL to use these clients, which is allowed but not recommended. In those cases, you have some options:
+In such cases, you would have to turn off SSL to use these clients, which is allowed but **not recommended**. You can use one of the following option to disable SSL
 
 
-Set up stunnel process
-~~~~~~~~~~~~~~~~~~~~~~
+Set up ``stunnel`` process
+~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-One workaround for this, it is to set up a `stunnel process <https://www.stunnel.org/index.html>`_ on the client side to handle encryption for the clients that do not support SSL connections. 
+If you want to keep SSL settings on database side, but hide it from the client side, you can set up a `Stunnel process <https://www.stunnel.org/index.html>`_ on the client to handle encryption. 
 
-You can use the following stunnel configuration, for example ``stunnel.conf``, to set up a stunnel process.
+You can use the following ``stunnel`` configuration, for example ``stunnel.conf``, to set up a ``stunnel`` process.
 ::
 
     client = yes
@@ -46,7 +46,7 @@ You can use the following stunnel configuration, for example ``stunnel.conf``, t
     ; Let's Encrypt by default without any explicit CAfile config.
     ; CAfile = /path/to/optional/project/cacert/that/you/can/download/from/aiven/console
 
-To understand the global options of the stunnel configuration, please check `Stunnel Global Options <https://www.stunnel.org/static/stunnel.html#GLOBAL-OPTIONS>`_.
+To understand the global options of the ``stunnel`` configuration, please check `Stunnel Global Options <https://www.stunnel.org/static/stunnel.html#GLOBAL-OPTIONS>`_.
 
 For ``service-level option``, the following parameters are configured:  
 
@@ -70,16 +70,17 @@ Note that when SSL is in use we have a separate service terminating the SSL conn
 Allow plain-text connections
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-Another alternative is to actually allow plain-text connections. To allow plain-text connections, you can change this setting with **Overview** > **Advanced configuration**, or using the :doc:`Aiven Command Line interface<../../../tools/cli>`.
+An alternative is disable database SSL allowing allow plain-text connections. To allow plain-text connections, you can change this setting with **Overview** > **Advanced configuration**, or using the :doc:`Aiven Command Line interface<../../../tools/cli>`.
 
 .. Warning::
     Allowing plain-text connections can have some implications regarding the security of your Redis service. If SSL is turned off, anyone who can eavesdrop on the traffic will be able to potentially connect and access your Aiven for Redis service.
 
-Alternatively, once installed, you should run::
+To disable SSL on an existing Redis instance use the following Aiven CLI command (with link to avn update command)
 
-    avn login  # if you haven't logged in previously
+.. code-block:: console
+
     avn service update myredis -c "redis_ssl=false"
 
-After this, the ``Service URI`` will change and point at the new location, it will also start with the ``redis://`` prefix denoting that it's a direct Redis connection which does not use SSL.
+After executing the command, the ``Service URI`` will change and point at the new location, it will also start with the ``redis://`` (removing the extra s) prefix denoting that it's a direct Redis connection which does not use SSL.
 
 
--- a/docs/products/redis/concepts/ssl-connectivity.rst
+++ b/docs/products/redis/concepts/ssl-connectivity.rst
@@ -1,18 +1,28 @@
 SSL connectivity
 ================
 
+Client support for SSL-encrypted connections
+--------------------------------------------
+
 Aiven for Redis uses SSL encrypted connections by default. This is shown by the use of ``rediss://`` (with double s) prefix in the ``Service URI``. 
 
 .. Tip::
     You can find the ``Service URI`` on `Aiven console <https://console.aiven.io/>`_.
 
-Since **Redis 6**, the redis-cli tool itself supports SSL connections; therefore, you can access connect directly to your service using::
+Since **Redis 6**, the redis-cli tool itself supports SSL connections; therefore, you can connect directly to your service using::
 
     redis-cli -u rediss://username:password@host:port
 
 Alternatively, you can use the third-party `redli tool <https://github.com/IBM-Cloud/redli>`_::
 
     redli -u rediss://username:password@host:port
+
+
+Not every Redis client supports SSL-encrypted connections. In those cases, you have some options:
+
+
+Set up stunnel process
+~~~~~~~~~~~~~~~~~~~~~~
 
 Not every Redis client supports SSL-encrypted connections. In such cases, you would have to turn off SSL to use these clients, which is allowed but not recommended. One workaround for this, it is to set up a stunnel process on the client side to handle encryption for the clients that do not support SSL connections. 
 
@@ -32,9 +42,13 @@ You can use the following stunnel configuration, for example ``example-stunnel.c
     ; CA cert available from Aiven console. Most environments trust
     ; Let's Encrypt by default without any explicit CAfile config.
     ; CAfile = /path/to/optional/project/cacert/that/you/can/download/from/aiven/console
+
 Note that when SSL is in use we have a separate service terminating the SSL connections before they are forwarded to Redis. This process has a connection timeout of its own independent of Redis' connection timeout. If you allow very long Redis timeouts this frontend service may end up closing the connection before the Redis timeout has expired. By the time of writing this timeout is set to 12 hours.
 
-Another alternative is to allow plain-text connections. You can change this setting with **Overview** > **Advanced configuration**, or using the :doc:`Aiven Command Line interface<../../../tools/cli>`.
+Allow plain-text connections
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+To allow plain-text connections, you can change this setting with **Overview** > **Advanced configuration**, or using the :doc:`Aiven Command Line interface<../../../tools/cli>`.
 
 .. Warning::
     Allowing plain-text connections can have some implications regarding the security of your Redis service. If SSL is turned off, anyone who can eavesdrop on the traffic will be able to potentially connect and access your Aiven for Redis service.
--- a/docs/products/redis/concepts/ssl-connectivity.rst
+++ b/docs/products/redis/concepts/ssl-connectivity.rst
@@ -1,0 +1,49 @@
+SSL connectivity
+================
+
+Aiven Redis uses SSL encrypted connections by default. This is shown by the use of ``rediss://`` (with double s) prefix in the ``Service URI``. 
+
+.. Tip::
+    You can find the ``Service URI`` on `Aiven console <https://console.aiven.io/>`_.
+
+Since **Redis 6**, the redis-cli tool itself supports SSL connections; therefore, you can access connect directly to your service using::
+
+    redis-cli -u rediss://username:password@host:port
+
+Alternatively, you can use the third-party `redli tool <https://github.com/IBM-Cloud/redli>`_::
+
+    redli -u rediss://username:password@host:port
+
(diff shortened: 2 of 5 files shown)


docs/products/redis/concepts/ssl-connectivity.rst
```diff
@@ -0,0 +1,85 @@
...
+--------------------------------------------
+
+Default support
+~~~~~~~~~~~~~~~
+Aiven for Redis uses SSL encrypted connections by default. This is shown by the use of ``rediss://`` (with double s) prefix in the ``Service URI``. 
+
+.. Tip::
+    You can find the ``Service URI`` on `Aiven console <https://console.aiven.io/>`_.
+
+Since **Redis 6**, the redis-cli tool itself supports SSL connections; therefore, you can connect directly to your service using::
```
ftisiot, line 14:
```suggestion
Since **Redis 6**, the ``redis-cli`` tool itself supports SSL connections; therefore, you can connect directly to your service using::
```

docs/products/redis/concepts/ssl-connectivity.rst
```diff
@@ -0,0 +1,85 @@
...
+
+    redis-cli -u rediss://username:password@host:port
+
+Alternatively, you can use the third-party `redli tool <https://github.com/IBM-Cloud/redli>`_::
+
+    redli -u rediss://username:password@host:port
+
+
+Not every Redis client supports SSL-encrypted connections.
+In such cases, you would have to turn off SSL to use these clients, which is allowed but not recommended. In those cases, you have some options:
```
ftisiot, line 24:
```suggestion
In such cases, you would have to turn off SSL to use these clients, which is allowed but **not recommended**. You can use one of the following option to disable SSL
```

docs/products/redis/concepts/ssl-connectivity.rst
```diff
@@ -0,0 +1,85 @@
...
+
+
+Set up stunnel process
+~~~~~~~~~~~~~~~~~~~~~~
+
+One workaround for this, it is to set up a `stunnel process <https://www.stunnel.org/index.html>`_ on the client side to handle encryption for the clients that do not support SSL connections.
```
ftisiot, line 30:
```suggestion
If you want to keep SSL settings on database side, but hide it from the client side, you can set up a `stunnel process <https://www.stunnel.org/index.html>`_ on the client to handle encryption. 
```

docs/products/redis/concepts/ssl-connectivity.rst
```diff
@@ -0,0 +1,85 @@
...
+Not every Redis client supports SSL-encrypted connections.
+In such cases, you would have to turn off SSL to use these clients, which is allowed but not recommended. In those cases, you have some options:
+
+
+Set up stunnel process
+~~~~~~~~~~~~~~~~~~~~~~
+
+One workaround for this, it is to set up a `stunnel process <https://www.stunnel.org/index.html>`_ on the client side to handle encryption for the clients that do not support SSL connections. 
+
+You can use the following stunnel configuration, for example ``stunnel.conf``, to set up a stunnel process.
```
ftisiot, line 32:
```suggestion
You can use the following ``stunnel`` configuration file, named for example ``stunnel.conf``, to set up a ``stunnel`` process.
```

docs/products/redis/concepts/ssl-connectivity.rst
```diff
@@ -0,0 +1,85 @@
...
+
+``connect`` => *[host:]port*
+  **Connect to a remote address** 
+
+
+
+``TIMEOUTclose`` => *seconds*
+  **Time to wait for close_notify**
+
+.. note:: It is important to make changes accordingly to your service. On the *Overview* page you can find your **Overview** > **Host** and **Overview** > **Port** to configure the ``connect`` parameter.
```
ftisiot, line 66:
```suggestion
.. note:: It is important to make changes accordingly to your service. On the *Overview* page you can find the **Host** and **Port** to configure the ``connect`` parameter.
```

docs/products/redis/concepts/ssl-connectivity.rst
```diff
@@ -0,0 +1,85 @@
...
+
+Note that when SSL is in use we have a separate service terminating the SSL connections before they are forwarded to Redis. This process has a connection timeout of its own independent of Redis' connection timeout. If you allow very long Redis timeouts, this frontend service may end up closing the connection before the Redis timeout has expired. By the time of writing this, timeout is set to 12 hours.
```
ftisiot, line 68: This paragraph is not clear in my opinion, where are the two timeouts? Can we explain it better?

docs/products/redis/concepts/ssl-connectivity.rst
```diff
@@ -0,0 +1,85 @@
...
+
+Allow plain-text connections
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Another alternative is to actually allow plain-text connections. To allow plain-text connections, you can change this setting with **Overview** > **Advanced configuration**, or using the :doc:`Aiven Command Line interface<../../../tools/cli>`.
```
ftisiot, line 73:
```suggestion
The alternative is to disable database SSL allowing plain-text connections. To allow plain-text connections, you can change this setting with **Overview** > **Advanced configuration**, or using the :doc:`Aiven Command Line interface<../../../tools/cli>`.
```

docs/products/redis/concepts/ssl-connectivity.rst
```diff
@@ -0,0 +1,85 @@
...
+
+.. Warning::
+    Allowing plain-text connections can have some implications regarding the security of your Redis service. If SSL is turned off, anyone who can eavesdrop on the traffic will be able to potentially connect and access your Aiven for Redis service.
+
+Alternatively, once installed, you should run::
```
ftisiot, line 78:
```suggestion
To disable ssl on an existing Redis instance use the following aiven cli command (with link to avn update command)
```

docs/products/redis/concepts/ssl-connectivity.rst
```diff
@@ -0,0 +1,85 @@
...
+
+    avn login  # if you haven't logged in previously
```
ftisiot, line 80: I would remove the login phase, we can assume one knows to login

docs/products/redis/concepts/ssl-connectivity.rst
```diff
@@ -0,0 +1,85 @@
...
+    avn service update myredis -c "redis_ssl=false"
+
+After this, the ``Service URI`` will change and point at the new location, it will also start with the ``redis://`` prefix denoting that it's a direct Redis connection which does not use SSL.
```
ftisiot, line 83:
```suggestion
After executing the command, the ``Service URI`` will change and point at the new location, it will also start with the ``redis://`` (removing the extra s) prefix denoting that it's a direct Redis connection which does not use SSL.
```

ftisiot (changes requested):
Good content! I left few comments.
generally I would try avoiding referencing "this" or "that" and instead reference the action or object you are acting on.

docs/products/redis/concepts/ssl-connectivity.rst
```diff
@@ -0,0 +1,85 @@
...
+Alternatively, you can use the third-party `redli tool <https://github.com/IBM-Cloud/redli>`_::
+
+    redli -u rediss://username:password@host:port
+
+
+Not every Redis client supports SSL-encrypted connections.
+In such cases, you would have to turn off SSL to use these clients, which is allowed but not recommended. In those cases, you have some options:
+
+
+Set up stunnel process
```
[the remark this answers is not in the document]
laysauchoa: I guess stunnel is more like the name of the process then a command itself, so I think makes sense to leave as it is

docs/products/redis/concepts/ssl-connectivity.rst
laysauchoa: I'd like to keep because it goes from those steps even if the user knows what to do. But you can not run the other command without the first one.

docs/products/redis/concepts/ssl-connectivity.rst
laysauchoa: I guess this is one of the alternatives not the only one, so imho `The alternative` may be not the proper term.

.github/styles/Vocab/Docs/accept.txt
```diff
@@ -83,7 +83,7 @@ Postman
 [Pp]rivatelink
 psql
 Python
-Redis
+[Rr]edis
```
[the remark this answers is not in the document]
lornajane: Please remove this change, Redis is a product name and should always be written with its capital letter. It means that if you use it in lowercase, for example in a command, it should have code/literal markup so that the spell check doesn't trip over it.

.github/styles/Vocab/Docs/accept.txt
```diff
@@ -115,4 +115,6 @@ wget
 Workbench
 Zabbix
 Zipkin
-ZooKeeper
\ No newline at end of file
+ZooKeeper
+redli
```
lornajane, line 119: The product name is "Redli" (with a capital initial letter). The command is lowercase but then should always use literal markup: ``redli``. Please be very careful to avoid adding case-insensitive entries for proper nouns to the spelling list; we should not use these words in this way.

.github/styles/Vocab/Docs/accept.txt
```diff
@@ -115,4 +115,6 @@ wget
...
+[Ss]tunnel
```
lornajane, line 120: The same comment here. The product is "Stunnel", and you can also use the command `stunnel` if you need to - but with literal/code markup and not as a word, so the spellchecker should not recognise the all-lowercase version of the word.

docs/products/redis/concepts/ssl-connectivity.rst
lornajane: You could use "An alternative" here if you prefer. It's up to you whether you accept the suggested wording change (and you can make more changes after you do, if you want)

docs/products/redis/concepts/ssl-connectivity.rst
lornajane: I agree with Francesco on this one. Users are not beginners, so they will be able to sort out the basics on the command: they are just looking for the specific thing they need. This is a tricky balance to strike and I'm not sure we have it well expressed in our guidelines at this point.

lornajane:
Added a few comments since this one seemed to have got stuck in a discussion. @laysauchoa please also pick up on the changes suggested by the spellchecker, and revisit the commit history. We don't squash, so your commits should have meaningful and unique commit messages.

.github/styles/Vocab/Docs/accept.txt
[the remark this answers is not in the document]
laysauchoa: Oki

docs/products/redis/concepts/ssl-connectivity.rst
laysauchoa: oki, I will use `An alternative`

docs/products/redis/concepts/ssl-connectivity.rst
laysauchoa: This one I still do not agree but I will make the changes cause maybe you have different pov from mine and we should go what works for more people :)

docs/products/redis/concepts/ssl-connectivity.rst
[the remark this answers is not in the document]
laysauchoa: nvm, I guess is because the spell checker here, thanks for the suggestion!

docs/products/redis/concepts/ssl-connectivity.rst
laysauchoa: I'd like to keep the previous format if it is oki

docs/products/redis/concepts/ssl-connectivity.rst
laysauchoa: This was not clear for me as well, but I could clarify and now, it is. 

https://github.com/aiven/devportal/pull/429/commits/96c3eaae5088032a2f01aa93878d27e82e43c222

Thanks for pointing this so we could make it more clear. What do you think now?

docs/products/redis/concepts/ssl-connectivity.rst
```diff
@@ -0,0 +1,86 @@
...
+  **Time to wait for close_notify**
+
+.. note:: It is important to make changes accordingly to your service. On the *Overview* page you can find your **Overview** > **Host** and **Overview** > **Port** to configure the ``connect`` parameter.
+
+It is important to note that when SSL is in use, HAProxy will be responsible for terminating the SSL connections before they get forwarded to Redis. This process has a connection timeout set to 12 hours which is not configurable by the customer. If you allow very long Redis timeouts, this SSL-terminating HAProxy may end up closing the connection before the Redis timeout has expired. This timeout is independent of Redis timeout.
+
+Allow plain-text connections
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+An alternative is disable database SSL allowing allow plain-text connections. To allow plain-text connections, you can change this setting with **Overview** > **Advanced configuration**, or using the :doc:`Aiven Command Line interface<../../../tools/cli>`.
```
ftisiot, line 73: We usually don't use `**Overview** > **Advanced configuration**` but rather write `**Overview** in the **Advanced configuration** section`

docs/products/redis/concepts/ssl-connectivity.rst
```diff
@@ -0,0 +1,86 @@
...
+
+.. Warning::
+    Allowing plain-text connections can have some implications regarding the security of your Redis service. If SSL is turned off, anyone who can eavesdrop on the traffic will be able to potentially connect and access your Aiven for Redis service.
```
ftisiot, line 76:
```suggestion
    Allowing plain-text connections can have some implications regarding the security of your Redis service. If SSL is turned off, anyone who can eavesdrop on the traffic will be able to steal the credentials, thus potentially connect and access your Aiven for Redis service.
```

docs/products/redis/concepts/ssl-connectivity.rst
```diff
@@ -0,0 +1,86 @@
...
+
+To disable SSL on an existing Redis instance use the following Aiven CLI command (with link to avn update command)
```
ftisiot, line 78: missing link to avn update command

docs/products/redis/concepts/ssl-connectivity.rst
```diff
@@ -0,0 +1,86 @@
+SSL connectivity
+================
+
+Client support for SSL-encrypted connections
+--------------------------------------------
+
+Default support
+~~~~~~~~~~~~~~~
+Aiven for Redis uses SSL encrypted connections by default. This is shown by the use of ``rediss://`` (with double s) prefix in the ``Service URI``.
```
ftisiot, line 9:
```suggestion
Aiven for Redis uses SSL encrypted connections by default. This is shown, in the [Aiven Console](https://console.aiven.io/)  ``Service URI`` parameter, by the use of ``rediss://`` (with double s) prefix. 
```

docs/products/redis/concepts/ssl-connectivity.rst
```diff
@@ -0,0 +1,86 @@
...
+
+    redis-cli -u rediss://username:password@host:port
+
+Alternatively, you can use the third-party `Redli tool <https://github.com/IBM-Cloud/redli>`_::
+
+    redli -u rediss://username:password@host:port
+
+
+Not every Redis client supports SSL-encrypted connections.
+In such cases, you would have to turn off SSL to use these clients, which is allowed but **not recommended**. You can use one of the following option to disable SSL.
```
ftisiot, line 24: One of the two options doesn't really disable SSL but rather bypasses the SSL problem by removing the SSL bit locally.
```suggestion
In such cases, you would have to turn off or bypass the SSL to use these clients, which is allowed but **not recommended**. You can use one of the following option to achieve this.
```

ftisiot (changes requested):
Almost there, I added some little notes for changes needed before merging

ftisiot:
one last comment before approving, could you please add the command to start `stunnel` with the configuration file and share how to connect via `stunnel`?

Since we are showing what to write in the config file, it would be also good to show how to start the process and connect

docs/products/redis/concepts/ssl-connectivity.rst
laysauchoa: re-write cause it was not clear. thanks for pointing it out!

laysauchoa:
I am not sure we should add it here because stunnel organisation is linked on this article has a **howto** with it and it will also differs for different operation system. 

I am not sure it should be done here because there are many steps till you just start the process and I used virtual env to run it on linux cause MacOS was more complicated.

I'd think that the user knows **something about stunnel process** and can reuse the configuration steps. To show how to run a stunnel process, I would suggest to have in a different article or to follow the official instructions. The final goal imho would be to reuse the configuration offered here for our specific case.

wdyt?

ftisiot:
ok then I agree, let's leave it out

docs/products/redis/concepts/ssl-connectivity.rst
```diff
@@ -0,0 +1,86 @@
...
+
+Client support for SSL-encrypted connections
+--------------------------------------------
+
+Default support
+~~~~~~~~~~~~~~~
+Aiven for Redis uses SSL encrypted connections by default. This is shown by the use of ``rediss://`` (with double s) prefix in the ``Service URI`` on the `Aiven Console <https://console.aiven.io/>`_.
+
+.. Tip::
+    You can find the ``Service URI`` on `Aiven console <https://console.aiven.io/>`_.
```
ftisiot, line 12: We can now remove the tip since the info is included in the previous sentence

laysauchoa:
You can find the proposed changes together with alternatives: https://github.com/aiven/devportal/pull/429/commits/6ddf79eec24244d4a02eab55e526a052e2b63721

ftisiot:
One last comment, can we put this under the HowTo and rename it as "Manage SSL connectivity" since we are giving info about how to disable or bypass?

laysauchoa:
omg, you are so right @ftisiot fits much better there: https://github.com/aiven/devportal/pull/429/commits/5bf6536a5af829b59a9f2ed2f37eae711fef765b

thank youuu!

ftisiot (approved):
LGTM

PR #430, opened 2021-12-22
Update dependencies, including Furo theme

lornajane:

There have been some updates to our dependencies, bring them into the project.

This should fix a problem with a missing dark mode CSS file as well (which we only have an internal/Asana ticket for)


laysauchoa (approved):
💯

PR #434, opened 2021-12-23
migrate/redis: warning overcommit_memory

laysauchoa:
Original URL:
https://help.aiven.io/en/articles/5157337-redis-overcommit_memory-warning

Devportal:
https://developer.aiven.io/docs/products/redis/howto/warning-overcommit_memory.html

Fixes: https://github.com/aiven/devportal/issues/435

### DIFF
--- a/.github/styles/Vocab/Docs/accept.txt
+++ b/.github/styles/Vocab/Docs/accept.txt
@@ -116,4 +116,3 @@ Workbench
 Zabbix
 Zipkin
 ZooKeeper
-overcommit
--- a/.github/styles/Vocab/Docs/accept.txt
+++ b/.github/styles/Vocab/Docs/accept.txt
@@ -116,3 +116,4 @@ Workbench
 Zabbix
 Zipkin
 ZooKeeper
+overcommit
--- a/docs/products/redis/concepts/warning-overcommit_memory.rst
+++ b/docs/products/redis/concepts/warning-overcommit_memory.rst
@@ -1,0 +1,8 @@
+Warning ``overcommit_memory``
+=============================
+
+By starting a Redis service on `Aiven console <https://console.aiven.io/>`_, you may notice on **Logs** the following **warning** ``overcommit_memory``::
+
+    # WARNING overcommit_memory is set to 0! Background save may fail under low memory condition. To fix this issue add 'vm.overcommit_memory = 1' to /etc/sysctl.conf and then reboot or run the command 'sysctl vm.overcommit_memory=1' for this to take effect.
+
+This warning can be safely ignored. Aiven for Redis limits the memory usage differently from how the software is deployed, which means the available memory will never drop low enough to hit this particular failure case.
--- a/docs/products/redis/concepts/warning-overcommit_memory.rst
+++ b/docs/products/redis/concepts/warning-overcommit_memory.rst
@@ -1,8 +1,0 @@
-Warning ``overcommit_memory``
-=============================
-
-By starting a Redis service on `Aiven console <https://console.aiven.io/>`_, you may notice on **Logs** the following **warning** ``overcommit_memory``::
-
-    # WARNING overcommit_memory is set to 0! Background save may fail under low memory condition. To fix this issue add 'vm.overcommit_memory = 1' to /etc/sysctl.conf and then reboot or run the command 'sysctl vm.overcommit_memory=1' for this to take effect.
-
-This warning can be safely ignored. Aiven for Redis limits the memory usage differently from how the software is deployed, which means the available memory will never drop low enough to hit this particular failure case.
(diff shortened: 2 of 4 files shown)


.github/styles/Vocab/Docs/accept.txt
```diff
@@ -116,3 +116,4 @@ Workbench
 Zabbix
 Zipkin
 ZooKeeper
+overcommit
```
anelook, line 119: We don't really use overcommit as a word, I think we don't need the word in accept.txt and it should still run without errors. Could you check?

docs/products/redis/concepts/warning-overcommit_memory.rst
```diff
@@ -0,0 +1,8 @@
+Warning ``overcommit_memory``
+=============================
+
+By starting a Redis service on `Aiven console <https://console.aiven.io/>`_, you may notice on **Logs** the following **warning** ``overcommit_memory``::
```
anelook, line 4:
```suggestion
When starting a Redis service on `Aiven console <https://console.aiven.io/>`_, you may notice on **Logs** the following **warning** ``overcommit_memory``::
```

Is it when starting the service with the Console UI or with the terminal? I suspect it doesn't matter how you start the service, in this case I would remove " on `Aiven console <https://console.aiven.io/>`_"

docs/products/redis/concepts/warning-overcommit_memory.rst
```diff
@@ -0,0 +1,8 @@
...
+
+    # WARNING overcommit_memory is set to 0! Background save may fail under low memory condition. To fix this issue add 'vm.overcommit_memory = 1' to /etc/sysctl.conf and then reboot or run the command 'sysctl vm.overcommit_memory=1' for this to take effect.
+
+This warning can be safely ignored. Aiven for Redis limits the memory usage differently from how the software is deployed, which means the available memory will never drop low enough to hit this particular failure case.
```
anelook, line 8: "from how the software is deployed," - what do we mean by this? Is it different because we manage it? I'm not a Redis person, but I think is a bit wage explanation to understand.

docs/products/redis/concepts/warning-overcommit_memory.rst
```diff
@@ -0,0 +1,8 @@
+Warning ``overcommit_memory``
```
anelook, line 1: It is a better fit to HowTo section, instead of Concepts.
HowTo "Handle ``overcommit_memory`` warning". Maybe there is a better fitting word than "handle"!

anelook (changes requested):
Thx for migrating it :) I've left couple of small comments.
BTW, if you start a service now, could you still reproduce this error?

docs/products/redis/concepts/warning-overcommit_memory.rst
laysauchoa: It is meant to be like: from the default use of the software without aiven. I also do not think needs to be added, so I can drop it out.

docs/products/redis/concepts/warning-overcommit_memory.rst
laysauchoa: handle seems a good fit, imo

laysauchoa:
Hi @anelook , thanks for the review. I have addressed the comments :)

docs/products/redis/howto/warning-overcommit_memory.rst
```diff
@@ -0,0 +1,8 @@
+Handle warning ``overcommit_memory``
+====================================
+
+When starting a Redis service on `Aiven console <https://console.aiven.io/>`_, you may notice on **Logs** the following **warning** ``overcommit_memory``::
+
+    # WARNING overcommit_memory is set to 0! Background save may fail under low memory condition. To fix this issue add 'vm.overcommit_memory = 1' to /etc/sysctl.conf and then reboot or run the command 'sysctl vm.overcommit_memory=1' for this to take effect.
+
+This warning can be safely ignored as the available memory will never drop low enough to hit this particular failure case.
```
anelook, line 8: I feel that the part of the initial explanation why it is so still worth keeping :) in the original it says that "we limit Redis memory usage a little differently from how the software is usually deployed", we can keep it as "Aiven for Redis ensures that the available memory will never drop..."? 

What do you think?

anelook (changes requested):

docs/products/redis/howto/warning-overcommit_memory.rst
laysauchoa: I think it makes sense cause still saves the initial thought on it :)

laysauchoa: Like this: https://github.com/aiven/devportal/pull/434/commits/c78f1c2e29a963535abd57b6649100485f0770ce
@anelook ^

anelook: Yes! Exactly :) Thx for changing it!

anelook (approved):

Issue #437, opened 2022-01-03
Lua scripts with Aiven for Redis


laysauchoa:
# What should we add to our content collections?
**Migration of Redis content.** 

**Original URL:**
https://help.aiven.io/en/articles/1110366-eval-and-load-script-commands

**Dev Portal:**
https://developer.aiven.io/docs/products/redis/howto/enable-lua-scripts.html

# Is this something you would work on yourself?
Yes. PR: https://github.com/aiven/devportal/pull/438

[x] I plan to open a pull request for this

Related to https://github.com/aiven/devportal/issues/116

PR #438, opened 2022-01-03
add/redis: enable lua scripts on redis

laysauchoa:
Migration of Redis content

Original URL:
https://help.aiven.io/en/articles/1110366-eval-and-load-script-commands

Dev Portal:
https://developer.aiven.io/docs/products/redis/concepts/lua-scripts-redis.html

Fix: https://github.com/aiven/devportal/issues/437


lornajane (approved):
Looks good! It would benefit from an example if we can add one in the future, but it's an improvement as it is. Thanks @laysauchoa :)

PR #440, opened 2022-01-04
add/opensearch: generated config docs

laysauchoa:
Add generated config docs

Fix: https://github.com/aiven/devportal/issues/445

### DIFF
--- a/docs/products/opensearch/reference/advanced-params.rst
+++ b/docs/products/opensearch/reference/advanced-params.rst
@@ -1,6 +1,6 @@
 List of advanced parameters
 ============================
 
-When creating or working with an Aiven for OpenSearch service, several configuration options are available. They are summarized below:
+Below you can find a summary of every configuration option available for Aiven for OpenSearch service:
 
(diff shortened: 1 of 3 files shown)


docs/products/opensearch/reference/advanced-params.rst
```diff
@@ -0,0 +1,6 @@
+List of advanced parameters
+============================
+
+When creating or working with an Aiven for OpenSearch service, several configuration options are available. They are summarized below:
```
anelook, line 4: Several usually means "more than two but not many.", in our case it feels like way more than several :D Let's rephrase it a bit.

For "When creating or working" would it be more concise to use "When working with". It summarises all scenarios and is shorter

anelook (changes requested):
So happy to see the configuration options being added!
I have same minor comments for all of the commits of this type.

laysauchoa:
Fix comments proposed here: https://github.com/aiven/devportal/pull/440

laysauchoa:
reviewed

anelook (approved):

PR #442, opened 2022-01-04
add/m3db: generated config docs

laysauchoa:
Add generated config docs 
Fix: https://github.com/aiven/devportal/issues/447

### DIFF
--- a/_toc.yml
+++ b/_toc.yml
@@ -235,7 +235,10 @@ entries:
         title: HowTo
         entries:
           - glob: docs/products/m3db/howto/*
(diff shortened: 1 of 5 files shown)


laysauchoa:
Fix comments proposed here: #440

laysauchoa:
reviewed

anelook (approved):
Thanks for adding the references to each of the services :)

As we discussed - since we generate this from sources I believe we should aim at keeping docs for every property.

In case we later disagree using "every" to be on a safer side - I'll make the correction to the docs :D

PR #443, opened 2022-01-04
fix/redis and kafka: missing links for Reference

laysauchoa:
Add missing links for reference in index

Fix: https://github.com/aiven/devportal/issues/452


anelook (approved):
Thanks! I believe this sounds more appropriate for documentation.

Issue #450, opened 2022-01-04
CLI docs are missing links to detailed documentation sections from main commands overview page


lornajane:
# What's wrong?

On this page https://developer.aiven.io/docs/tools/cli there are a couple of commands (I spotted `service` and `ticket`) that don't link to further information, so a user doesn't know that they find out more until they find the navigation entry by themselves.

laysauchoa:
PR: https://github.com/aiven/devportal/pull/453

PR #453, opened 2022-01-05
fix/cli: missing links to the commands

laysauchoa:
On https://developer.aiven.io/docs/tools/cli there are a couple of commands
that are missing links to further information. Fix by adding the right
links.

Fix: https://github.com/aiven/devportal/issues/450


lornajane (approved):
Thanks!

Issue #465, opened 2022-01-11
Giving the community page some love


FloorD:
The community page at present is pretty barebones. I'd like to add little blocks (taking inspiration from the start page developer.aiven.io) to organize the references to Twitter, LinkedIn, the blog, and YouTube:

![Screenshot (9)](https://user-images.githubusercontent.com/1087213/148941297-42341781-5392-4a72-b5c8-8d021d20f677.png)

I'd also love to add a block to share (technical, community) events where people can meet us (we speak at / run a booth at). In a first integration this'd be hardcoded, but I'd love to eventually pull this data from a/our Google calendar, or using Meetup(s) API.

Month XX, [Event]()
_Who's speaking / what's happening?_

So, for example:
January 20, [PHP User Group Frankfurt](https://www.phpugffm.de/veranstaltungen-2/phpugffm-i-2022/)
_Join Lorna Jane for "Aiven for Laravel"_

laysauchoa:
Super ideas! :D

FloorD:
@lornajane WDYT? What could work best in terms of pulling in event data? A Google Calendar connection? Maybe https://developers.google.com/calendar/api/v3/reference/events/list#python 

In the past I've used the Meetup API, but Aiven doesn't have a Meetup page and creating one for this integration only adds a source we'd need to monitor. Unless you'd consider using Meetup to add discoverability for Aiven events, webinars, stuff?

lornajane:
I don't have a data source that we can automate on this, and I am not sure we'd want to pull in everything in any case. So getting the right data into one place is a first step, but I love the idea :heart:

PR #467, opened 2022-01-11
Update community page

FloorD:
Following #465, I've added some styling, and a first version of an event list (to be automated), to have something to look at. The Community page was very bare-bones before. 

A screenshot of what I did:

screencapture-localhost-8000-docs-community-html-2022-01-11-15_13_43

### DIFF
--- a/docs/community.rst
+++ b/docs/community.rst
@@ -3,13 +3,83 @@ Aiven community
 
 There are lots of other ways to learn more about Aiven, and interact with us.
 
-**Blog**: https://aiven.io/blog - to read tech news, tutorials, and updates on what we're up to.
 
-**Aiven Console**: https://console.aiven.io - your Aiven web interface. New accounts get a free trial.
+.. panels::
+    :card: shadow
 
-**Help articles**: https://help.aiven.io - more tech facing goodness to help you get things done.
+    🐦 **Twitter** We love to chat! Tell us what you're building with Aiven!
 
-**Twitter**: https://twitter.com/aiven_io - we love to chat! Tell us what you're building with Aiven :)
+    +++
 
-**YouTube**: https://www.youtube.com/c/Aiven_io - if video is your thing, cool, it's ours too.
+    .. link-button:: https://twitter.com/aiven_io
+        :type: ref
+        :text: To the Twitterverse
+        :classes: stretched-link
 
+    ---
+
+    📖 **Blog** To read tech news, tutorials, and updates on what we're up to.
+
+    +++
+
+    .. link-button:: https://aiven.io/blog
+        :type: ref
+        :text: To the blog
+        :classes: stretched-link
+
+    ---
+
+    📺 **YouTube** If video is your thing, cool, it's ours too.
+
+
+    +++
+
+    .. link-button:: https://www.youtube.com/c/Aiven_io
+        :type: ref
+        :text: To YouTube
+        :classes: stretched-link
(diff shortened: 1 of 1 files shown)


docs/community.rst
```diff
@@ -3,13 +3,83 @@ Aiven community
...
+
+January 20, `PHP User Group Frankfurt <https://www.phpugffm.de/veranstaltungen-2/phpugffm-i-2022/>`_
+
+Join Lorna Jane for "Aiven for Laravel" 
+
+------------
+
+February 5-6, `FOSDEM <https://fosdem.org/2022/>`_
+
+Join Francesco Tisiot and Olena Kutsenko for their sessions
```
laysauchoa, line 71: instead of names, I would suggest to add the github handle

TibsAtWork: Or both? Since the casual reader will not necessarily realise who the github id means.

ftisiot: Why not the Twitter handle?

laysauchoa: I think is like, most likely people want to contact you somehow and from github handle is more like looking what you are doing and github people can drop your message. Another option, it is e-mail. Up to the team. I'd be fine either way. :)

TibsAtWork: And yes, comparing that to what is there now, that is much nicer.

TibsAtWork:
And yes, comparing that to the previous page, it is much nicer.

anelook:
Cool info and give some energy to the page. 
How will we maintain this list?

FloorD:
> Cool info and give some energy to the page. How will we maintain this list?

The idea would be to hook an API in here. So maybe pulling from the Google calendar (and follow conventions like putting the speaker and talk title in the descr. of the calendar item), or from a (to be created) Meetup group. Until that's decided I'd be happy to maintain this. 

We can link names to Twitter accounts or GH (instead of just using usernames). Would need to figure out how we can get that to work in an automated workflow but should be possible.

laysauchoa:
Why do you think we need some external automation for the twitter names? Do you think something like this -> [@laysauchoa](https://twitter.com/laysauchoa) would work for this case? I have not found of twitter account can be modified, so I guess they cannot.

anelook:
Alternatively, a simple solution can be to keep up-to-date info in a google spreadsheet and inject it here as an iframe. (just got this idea from Felix @runwuf :D)

runwuf:
static google spreadsheet can be easily integrated by embed the code generated out of `File > Share > Publish to web` -> `Embed` -> `Publish`, but it is going to be static, to get a live/editable spreadsheet embed is a bit tricky, ping me if needed. I might document the steps after.

lornajane:
The styling improvements by themselves are worth it :) I'm not sure about maintaining a list of events by hand but if you want to do that then we can try, and just remove the section if it doesn't work out!

The build failed because it doesn't know our proper names (also can you use "Lorna Mitchell" for me please?), you can add them to the vocabulary file (info is in the readme) to permit those words.

laysauchoa:
rebase with the main page

laysauchoa:
The way that it is, you won't be redirected to the right links in none of the buttons. I had the same issue and one needs to remove the ` :type: ref`. Leaving here for your reference.

```
    .. link-button:: https://twitter.com/aiven_io
        :type: ref
        :text: To the Twitterverse
        :classes: stretched-link
```

PR #469, opened 2022-01-11
remove pg extension tsearch2

carobme:
PostgreSQL 9.6 isn't offered on the Aiven platform anymore, so the
tsearch2 extension is essentially gone now.


ftisiot (approved):
Thanks for checking this!

PR #482, opened 2022-01-17
fixed OpenSearch shards recommendation and incorporated sizing calculator

runwuf:
This pull request is to update OpenSearch shards recommendation numbers for best practice sizing.


docs/products/opensearch/concepts/indices.rst
```diff
@@ -46,14 +46,17 @@ These are only rough suggestions; the optimal values depend heavily on how you u
 Performance impact
 -------------------
 
-Having a large number of indices or shards affects the performance of your OpenSearch service. As an example, here are some figures from a three-node Aiven for OpenSearch business-8 cluster:
+Having a large number of indices or shards affects the performance of your OpenSearch service.  
+20 shards or few per GB of heap memory should be set as per `recommendation <https://www.elastic.co/guide/en/elasticsearch/reference/current/size-your-shards.html#shard-count-recommendation>`_.  As an example, an OpenSearch business-8 cluster comes with 8GB heap,
+20 x 8 = 160 or less would be the recommended configuration for number of shards.
```
reta, line 52: Would it make sense to include the shard size recommendations [1] as well? It think it would be valuable. 

[1] https://www.elastic.co/guide/en/elasticsearch/reference/current/size-your-shards.html#shard-size-recommendation

runwuf: Yes, thanks for your suggestion.  Shard size recommendation added!

reta (approved):

docs/products/opensearch/concepts/indices.rst
```diff
@@ -46,14 +46,19 @@ These are only rough suggestions; the optimal values depend heavily on how you u
...
 
-* 1,000 shards: No visible effect in performance.
-* 10,000 shards: Creating new shards starts to take more time. Increased variation in performance.
-* 15,000 shards: Creating new shards takes significantly longer, often tens of seconds.
-* 20,000 shards: Inserting new data randomly takes significantly more time (20 times more than the mean average). Significantly increased variation in performance.
+Size of shard impacts the recovery time after a failure, shard sizes between 10GB to 50GB should be set as per `shard size recommendation <https://www.elastic.co/guide/en/elasticsearch/reference/current/size-your-shards.html#shard-size-recommendation>`_.
 
-Aiven for OpenSearch takes a snapshot once every hour. With 10,000 shards, the cluster is continuously taking new backups and deleting old backups from storage. This naturally affects the service performance, as part of the capacity is continuously allocated to managing backups.
+The following interactive calculator can be used to tweak the OpenSearch configuration values.
```
lornajane, line 55: I think we could say a bit more about what the spreadsheet is, and how to use it. It should be read only (otherwise all the values will be reset often - I've changed this already), and I think we can tell users to take a copy of the sheet and use it to enter their own ballpark values to get an idea of how to configure opensearch for their needs. Probably with disclaimers.

docs/products/opensearch/concepts/indices.rst
```diff
@@ -46,14 +46,19 @@ These are only rough suggestions; the optimal values depend heavily on how you u
...
+
+.. raw:: html
```
lornajane, line 57: Please link to the sheet rather than embedding HTML inside the ReStructuredText format. I don't think people should use the sheet inside the page, but feel free to add a screenshot to hint where the link takes you.

lornajane (changes requested):
Please add a description to your pull request. This is required for all pull requests, and helps reviewers have the context of your change.

I love the spreadsheet planner tool here but adding one world-writeable sheet is probably going to lead to bad outcomes (people can't use it at the same time, it can be updated to give wrong calculations, etc). Could we therefore:
 - link to the sheet rather than embedding it, iframes aren't a standard practice in our docs
 - make the sheet read-only and explain what it does, link the user to it, and prompt them to make a copy. I'm also not sure how the user is expected to use the fields, or which fields should be used. I think a bit more explanation would make this even more useful.
 - Add some explanation also in the sheet about what this is, and link back to this page
 - Add the red cells to the key in the sheet

docs/products/opensearch/concepts/indices.rst
runwuf: Hi @lornajane Thank you for your feedback.  The sheet had all the cells protected to non-aiven/anonymous users except the yellow cells where it allows inputs but I can see there would be an issue when two or more users trying to use the calculator at the same time, so I have removed the embedded sheet.

I have updated all of the above as per your feedback, except the last one as I am not clear what needs to be done?
"Add the red cells to the key in the sheet"

lornajane: The sheet has some cells that have yellow background, some are blue, some are grey and some are red. The red background isn't in the key.

runwuf: Thanks! I updated the red cell description to the top of the sheet.

jhepp00 (approved):
All changes from @lornajane appear to have been merged at this point in time.

jhepp00 (approved):

jhepp00 (approved):

jhepp00 (approved):

lornajane (approved):

PR #496, opened 2022-01-21
Cassandra: Add documentation for the Cassandra upgrade

rsalevsky:

### DIFF
--- a/docs/products/cassandra/howto/upgrade.rst
+++ b/docs/products/cassandra/howto/upgrade.rst
@@ -1,7 +1,7 @@
 Perform a Cassandra major version upgrade
 ==========================================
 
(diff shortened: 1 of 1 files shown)


laysauchoa:
Sorry, I mistake that this was for review so I just applied some small fix so the build could run. Feel free to reset it if does not make sense.

lornajane:
Just a quick note that we probably need to create the landing page as well for Cassandra in order to be able to merge this, please let us know if you have a timescale in mind? We'll put something together!

lornajane:
@rsalevsky is this PR no longer needed? It would be good to add a comment when we close a PR without merge if possible.

rsalevsky:
@lornajane It is still needed but we (@diawarr and I) decided to have it as a help article and move the Cassandra content altogether. Just having a single article made no sense.

PR #499, opened 2022-01-25
updated postgres replica DNS naming from replica. to replica-

runwuf:
Updated postgres replica DNS naming from `replica.` to `replica-`
`replica.` DNS records are no longer used for single nodes
https://aiven.atlassian.net/browse/BF-873


dewan-ahmed (approved):
LGTM.

PR #500, opened 2022-01-25
redis: fix broken links

laysauchoa:
File name did not match with, so
it could not be accessed like that.

Signed-off-by: laysa.uchoa <laysa.uchoa@aiven.io>


TibsAtWork (approved):
Confirmed the new link works

PR #503, opened 2022-01-26
Migrated prevent full disk in PostgreSQL

### LINKED ISSUE #293: PostgreSQL disk full issues
Migrate existing article https://help.aiven.io/en/articles/2990664-preventing-postgresql-full-disk-issues

ftisiot:

Migrated https://help.aiven.io/en/articles/2990664-preventing-postgresql-full-disk-issues 

Fix #293

### DIFF
--- a/docs/products/postgresql/howto/prevent-full-disk.rst
+++ b/docs/products/postgresql/howto/prevent-full-disk.rst
@@ -1,0 +1,58 @@
+Prevent PostgreSQL full disk issues
+===================================
+
+If your Aiven for PostgreSQL service runs out of disk space, the service will start malfunctioning and precluding the proper backups creation. 
+
+To prevent this situation, Aiven automatically detects when your service is running out of free space and stops further write operations by setting the ``default_transaction_read_only``  parameter to ``ON``.
+
+With this setting in place, clients trying to execute write operations will start facing errors like::
+
+    cannot execute CREATE TABLE in a read-only transaction.
+
+To re-enable database writes you need to increase the available space, by either deleting data or upgrading to a larger plan.
+
+Increase free space by upgrading to a larger plan
+-------------------------------------------------
+
+When upgrading to a larger plan, new nodes with bigger disk space capacity, are created and replace the original nodes. Once the upgrade is finished and the free space increased, write operations are enabled again.
+You can upgrade the Aiven for PostgreSQL service plan via the `Aiven console <https://console.aiven.io/>`_ or :doc:`Aiven CLI </docs/tools/cli>`. 
+
+To perform a plan upgrade via the `Aiven console <https://console.aiven.io/>`_:
+
+* Go to the Aiven for PostgreSQL service page
+* In the *Overview* tab scroll down to the **Service plan** section and click on **Upgrade Plan**
+* Select the new plan with higher capacity and click **Upgrade**.
+
+Once the new nodes with increased disk capacity are up and running, the disk usage returns to below the critical level and the system automatically sets the ``default_transaction_read_only`` parameter to ``OFF`` allowing write operations again.
+
+.. Note::
+
+    If you temporarily upgrade your service to delete data, you may have to wait for the next backup to complete before you can downgrade to a smaller plan.
+
+Increase free space by deleting data
+------------------------------------
+
+To release space from a database, you can also delete data stored in it, but the database read-only mode also prevents them. 
+You can enable deletions by either enabling writes for a specific session or for a limited amount of time over the full database.
+
+Enable database writes for a specific session
+'''''''''''''''''''''''''''''''''''''''''''''
+
+If you want to enable writes for a session, login to the required database and execute the following command:
+
+::
+
+    SET default_transaction_read_only = OFF;
+
+You can then delete data within your session.
+
+Enable database writes for a limited amount of time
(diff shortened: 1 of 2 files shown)


docs/products/postgresql/howto/prevent-full-disk.rst
```diff
@@ -0,0 +1,58 @@
...
+Once the new nodes with increased disk capacity are up and running, the disk usage returns to below the critical level and the system automatically sets the ``default_transaction_read_only`` parameter to ``OFF`` allowing write operations again.
+
+.. Note::
+
+    If you temporarily upgrade your service to delete data, you may have to wait for the next backup to complete before you can downgrade to a smaller plan.
+
+Increase free space by deleting data
+------------------------------------
+
+To release space from a database, you can also delete data stored in it, but the database read-only mode also prevents them.
```
anelook, line 35:
```suggestion
To release space from a database, you can also delete data stored in it, but the database read-only mode also prevents this. 
```

anelook: I'd explain that this is because the database is in read-only transaction

anelook (changes requested):
Really nice read! I left some comments, nothing major, overall seems to be great as it is.Just some extra ideas on what can be smoothed.

docs/products/postgresql/howto/prevent-full-disk.rst
```diff
@@ -0,0 +1,58 @@
+Prevent PostgreSQL full disk issues
+===================================
+
+If your Aiven for PostgreSQL service runs out of disk space, the service will start malfunctioning and precluding the proper backups creation.
```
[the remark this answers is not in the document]
TibsAtWork: Hmm. Precluding doesn't feel quite right here - I think it would need to be "which will preclude".

I'd probably just say "will start malfunctioning and backup will not work." (or, if you prefer, "will not work properly", but I assume "will not work" is accurate enough?)

(I quite like the use of malfunctioning, though)

docs/products/postgresql/howto/prevent-full-disk.rst
```diff
@@ -0,0 +1,58 @@
...
+
+#. Go to the Aiven for PostgreSQL service page
+#. In the *Overview* tab scroll down to the **Service plan** section and click on **Upgrade Plan**
+#. Select the new plan with higher capacity and click **Upgrade**.
+
+Once the new nodes with increased disk capacity are up and running, the disk usage returns to below the critical level and the system automatically sets the ``default_transaction_read_only`` parameter to ``OFF`` allowing write operations again.
+
+.. Note::
+
+    If you temporarily upgrade your service to delete data, you may have to wait for the next backup to complete before you can downgrade to a smaller plan.
```
TibsAtWork, line 30: The next section talks about enabling the deletion of data without changing plan.

It's not obvious to me why, instead, you'd upgrade size, delete, and then downgrade size, when you could instead just delete.

I think this note either needs to explain that a bit more, or else perhaps it can go away?

TibsAtWork:
Apart from the two comments above, this reads well and I think is good.

docs/products/postgresql/howto/prevent-full-disk.rst
ftisiot: I believe the baseline is, if you want to have more time to delete the data you upgrade, delete and then resize

ftisiot: Will explain it better

anelook (approved):
Looks great to me. Not merging, in case @TibsAtWork has extra comments. :) P.S. I now like the word "malfunctioning" as well!

docs/products/postgresql/howto/prevent-full-disk.rst
```diff
@@ -0,0 +1,58 @@
+Prevent PostgreSQL full disk issues
+===================================
+
+If your Aiven for PostgreSQL service runs out of disk space, the service will start malfunctioning which will preclude the proper backups creation.
```
TibsAtWork, line 4: If you are keeping "precluding", then I think it needs to be
```
will start malfunctioning, which will preclude proper backup creation.
```
(so removing `the` and `s`), which is perfectly OK, or

ftisiot: Thanks fixed! Could you please do one more overall check and then proceed with the usual process?

docs/products/postgresql/howto/prevent-full-disk.rst
```diff
@@ -0,0 +1,58 @@
...
+
+#. Go to the Aiven for PostgreSQL service page
+#. In the *Overview* tab scroll down to the **Service plan** section and click on **Upgrade Plan**
+#. Select the new plan with higher capacity and click **Upgrade**.
+
+Once the new nodes with increased disk capacity are up and running, the disk usage returns to below the critical level and the system automatically sets the ``default_transaction_read_only`` parameter to ``OFF`` allowing write operations again.
+
+.. Note::
+
+    You might want to temporarily upgrade your service to a bigger plan to enable all users or applications to delete data without strict time limits. In such cases, you may have to wait for the next service backup to complete before you can downgrade to a smaller plan.
```
TibsAtWork, line 30: Nice, I like this.

Issue #302, opened 2021-10-25
Postgres Transaction wraparound


lornajane:
Migrate existing article https://help.aiven.io/en/articles/2183573-postgresql-transaction-id-wraparound

PR #507, opened 2022-01-26
Migrate "Postgres transaction wraparound" article

TibsAtWork:
Original is at
https://help.aiven.io/en/articles/2183573-postgresql-transaction-id-wraparound

Fix: https://github.com/aiven/devportal/issues/302

Change from the original are mainly:

1. Make the numbers friendlier (to me, at least)
2. Add link to PgHero
3. Add all the `®`
4. Add some references to actual documentation at the end - this wasn't in the original but I think is useful.
5. In the end I thought that `autovacuum` probably is a word for us, so added it.


docs/products/postgresql/concepts/transaction-id-wraparound.rst
```diff
@@ -0,0 +1,19 @@
+Transaction ID wraparound
+=========================
+
+The PostgreSQL® transaction control mechanism assigns a transaction ID to every row that is modified in the database; these IDs control the visibility of that row to other concurrent transactions.
+
+The transaction ID is a 32-bit number, where 2 billion (2 thousand million) IDs are always in the "visible past" and the remainder (about 2.2 billion) are reserved for future transactions and not visible to the running transaction. To avoid a transaction wraparound and having old, existing rows invisible when more transactions are created, PostgreSQL® requires an occasional cleanup and "freezing" of old rows. You can do this manually by executing ``VACUUM FREEZE``, but the autovacuum also does this automatically once a configured number of transactions have been created since the last freeze.
```
ftisiot, line 6:
```suggestion
The transaction ID is a 32-bit number, where 2 billion (2 thousand million) IDs are always in the "visible past" and the remainder (about 2.2 billion) are reserved for future transactions and not visible to the running transaction. To avoid a transaction wraparound and having old, existing rows invisible when more transactions are created, PostgreSQL® requires an occasional cleanup and "freezing" of old rows. 

You can do this manually by executing ``VACUUM FREEZE``, but the autovacuum also does this automatically once a configured number of transactions have been created since the last freeze.
```

ftisiot: Cutting the huge paragraph in two since the second sentence contains an action that a DBA could take?

ftisiot (changes requested):
Good one Tibs!

I'm not sure about the format, to me the top part is a concept, but the bottom, where we tell people to use `VACUM FREEZE`, tell which query to run to show the freeze age and the hints about the applications not reading the default could be an how to, something along the lines of "Check and avoid transaction ID wraparound" 

WDYT?

docs/products/postgresql/concepts/transaction-id-wraparound.rst
TibsAtWork: Thanks - splitting the paragraph is/was a good idea.

TibsAtWork:
I think you're right about the article being too much "here is what to do" for a "concepts" piece.

I can see two possibilities:

1. Keep the first two paragraphs as a "concepts" document, and move the rest to a "how to" document, quite likely named as you suggest. The two would link to each other.
2. Move the whole document to a "how to", again with suitable name, in which case the first two paragraphs are just "setting the scene".

I've just discovered I can argue quite strongly for both approaches 😢 - do you have an opinion?

ftisiot:
I would support the second option if there is nothing special Aiven does to prevent the wraparound. As you say the first couple paragraphs could serve as needed intro

docs/products/postgresql/howto/check-avoid-transaction-id-wraparound.rst
```diff
@@ -0,0 +1,21 @@
+Check and avoid transaction ID wraparound
+=========================================
+
+The PostgreSQL® transaction control mechanism assigns a transaction ID to every row that is modified in the database; these IDs control the visibility of that row to other concurrent transactions.
+
+The transaction ID is a 32-bit number, where 2 billion (2 thousand million) IDs are always in the "visible past" and the remainder (about 2.2 billion) are reserved for future transactions and not visible to the running transaction. To avoid a transaction wraparound and having old, existing rows invisible when more transactions are created, PostgreSQL® requires an occasional cleanup and "freezing" of old rows. 
+
```
ftisiot, line 7: What about adding subtitles for the actions? Like here "Execute `VACUUM` manually"

This way we are cutting the wall of text in pieces and people can easily spot the part they're interested into

docs/products/postgresql/howto/check-avoid-transaction-id-wraparound.rst
```diff
@@ -0,0 +1,21 @@
...
+You can do this manually by executing ``VACUUM FREEZE``, but the autovacuum also does this automatically once a configured number of transactions have been created since the last freeze.
+
```
ftisiot, line 9: Same as above, I would put a subtitle like "Check the autovacuum limits"

docs/products/postgresql/howto/check-avoid-transaction-id-wraparound.rst
```diff
@@ -0,0 +1,21 @@
...
+Aiven for PostgreSQL® sets that number to scale according to the database size, up to 1.5 billion transactions (which leaves 500 million transaction IDs available before a forced freeze), to avoid unnecessary churn for stable data in existing tables. To check your transaction freeze limits, run the following command in your PostgreSQL® instance::
+
+    show autovacuum_freeze_max_age
+
+This shows you the number of transactions that trigger autovacuum to start freezing old rows.
+
```
ftisiot, line 15: and here a subtitle like "Configure Autovacuum and client applications"

ftisiot (changes requested):
Almost perfect, I would add subtitles to break a bit the text in more digestible and easily findable pieces

docs/products/postgresql/howto/check-avoid-transaction-id-wraparound.rst
TibsAtWork: Yes, that seems like a good idea - adding more navigation seems useful.
I tried a ``rubric`` directive, but plain titles seems better.
Went with "Automatic or manual vacuuming" here, as a first attempt.

docs/products/postgresql/howto/check-avoid-transaction-id-wraparound.rst
```diff
@@ -0,0 +1,32 @@
...
+=========================================
+
+The PostgreSQL® transaction control mechanism assigns a transaction ID to every row that is modified in the database; these IDs control the visibility of that row to other concurrent transactions.
+
+The transaction ID is a 32-bit number, where 2 billion (2 thousand million) IDs are always in the "visible past" and the remainder (about 2.2 billion) are reserved for future transactions and not visible to the running transaction. To avoid a transaction wraparound and having old, existing rows invisible when more transactions are created, PostgreSQL® requires an occasional cleanup and "freezing" of old rows.
+
+Automatic or manual vacuuming
+-----------------------------
+
+You can do this manually by executing ``VACUUM FREEZE``, but the autovacuum also does this automatically once a configured number of transactions have been created since the last freeze.
```
ftisiot, line 11: Since we now have this sentence alone under a new subtitle maybe we need to give a bit more background about what "this" means.

ftisiot (commented):
Just a quick comment, giving a bit more context on the "Automatic or manual vacuuming" section since it references "this" without giving a lot of context

docs/products/postgresql/howto/check-avoid-transaction-id-wraparound.rst
TibsAtWork: Ooh, yes, that's well spotted. You're definitely right.

TibsAtWork: I've amplified a bit - hope the new version makes sense.

docs/products/postgresql/howto/check-avoid-transaction-id-wraparound.rst
ftisiot: One last change, we tend to use the present for titles, so this would be "Check the ``autovacuum`` frequency

docs/products/postgresql/howto/check-avoid-transaction-id-wraparound.rst
```diff
@@ -0,0 +1,32 @@
...
+Checking the ``autovacuum`` frequency
+-------------------------------------
+
+Aiven for PostgreSQL® sets that number to scale according to the database size, up to 1.5 billion transactions (which leaves 500 million transaction IDs available before a forced freeze), to avoid unnecessary churn for stable data in existing tables. To check your transaction freeze limits, run the following command in your PostgreSQL® instance::
+
+    show autovacuum_freeze_max_age
+
+This shows you the number of transactions that trigger autovacuum to start freezing old rows.
+
+Configuring client applications
```
ftisiot, line 22: Same here "Configure client applications"

docs/products/postgresql/howto/check-avoid-transaction-id-wraparound.rst
TibsAtWork: Well spotted - I'm learning!

docs/products/postgresql/howto/check-avoid-transaction-id-wraparound.rst
TibsAtWork: Thanks.

ftisiot (approved):
Good Work!

PR #510, opened 2022-01-26
Work on sorting out ® handling in devportal

TibsAtWork:
**Note** there's a lot to look at here. I shall continue rebasing as necessary to keep it up-to-date, so that it can be studied when convenient.

**IMPORTANT** Your local vale binary will need updating to *at least* 2.15.3, and I recommend going all the way to the latest release, 2.16.0. See https://vale.sh/docs/vale-cli/installation/ (and `homebrew` will upgrade you to 2.16.0)

Reorganise and (hopefully) improve our vale usage in several ways:

1. Move the vale configuration to `.github/vale`, and in particular the styles are now in `.github/vale/styles`
2. Create some initial rules to remind us to use ® marks - these were tested against https://github.com/aiven/devportal/pull/605
3. Move away from `accept.txt` and instead use a proper `aiven` dictionary
4. Add more "common replacements"
5. Name the exceptions for the "Sentence case headings" style in the style file (see [.github/vale/styles/capitalization_headings.yml](https://github.com/aiven/devportal/blob/tibs-vale-check-registered-marks/.github/vale/styles/Aiven/capitalization_headings.yml)).
6. Update the github workflow to use (at least) vale v2.15.3. The latest version is v2.15.5. The latest release of vale-actions (v1.5.0) is still using v2.15.2. However, the `master` branch of vale-actions is using v2.15.5, so I have set our workflow ([.github/vale/workflows/lint.yaml](https://github.com/aiven/devportal/blob/tibs-vale-check-registered-marks/.github/workflows/lint.yaml)) to use vale-actions `master`. This should probably be changed back to a specific release when possible, but need not be done in this PR
7. Add at least some documentation of how these things work (see [.github/vale/README.rst](https://github.com/aiven/devportal/blob/tibs-vale-check-registered-marks/.github/vale/README.rst)
8. Add some tests (see [Test files](https://github.com/aiven/devportal/blob/tibs-vale-check-registered-marks/.github/vale/README.rst#test-files) in the README)

An earlier version of this branch was used in https://github.com/aiven/devportal/pull/605 with a locally patched version of vale to make many of the ® and other corrections, and this branch has been rebased on top of that work.

Fixes: https://github.com/aiven/devportal/issues/672

-----

As well as passing the github checks below, you can also still do `make spell` locally.


README.rst
```diff
@@ -46,7 +46,9 @@ To run the spell check locally, you will need to have `Vale <https://github.com/
 * Check links: ``make linkcheck``
 * Check spelling: ``make spell``
 
-If the spellchecker is rejecting words that are valid (such as technology terms), double check the spelling and capitalisation, then add the word to ``.github/styles/Vocab/Docs/accept.txt``.
+For documentation on how we use Vale, see `our Vale README <.github/vale/README.rst>`_.
+
+If the spellchecker is rejecting words that are valid (such as technology terms), double check the spelling and capitalisation. If it's upset about the use of a capitalised word in a title, then add the word for ``.github/vale/styles/Aiven/capitalization_headings.yml``, and otherwise add it to the Aiven dictionary at  ``.github/vale/dicts/aiven.dic``. Please note that we try not to add the names of commands to the dictionary (for instance, ``jq`` or ``kcat``). See the section on dictionaries in `our Vale README <.github/vale/README.rst>`_
```
lornajane, line 51: Could we summarise this as "we don't add command names to the dictionary, these should be in code markup, that the spell check ignores" or something?

docs/products/grafana/reference/plugins.rst
```diff
@@ -95,7 +95,7 @@ Status history
 Status Panel - `Grafana <https://grafana.com/grafana/plugins/vonage-status-panel/>`__ | `GitHub <https://github.com/Vonage/Grafana_Status_panel>`__
     Use as a centralized view for the status of component in a glance.
 
-Statusmap - `Grafana <https://grafana.com/grafana/plugins/flant-statusmap-panel/>`__ | `GitHub <https://github.com/flant/grafana-statusmap>`__
+``Statusmap`` - `Grafana <https://grafana.com/grafana/plugins/flant-statusmap-panel/>`__ | `GitHub <https://github.com/flant/grafana-statusmap>`__
```
lornajane, line 98: (I'll comment once although we should apply whatever we agree to a few instances of this). I think Statusmap is a proper noun in this context, not a command or keyword that needs the code/literal markup. But I guess you did this for a reason, so can we discuss? It's like we'll create additional documentation about how to do things with the Statusmap plugin in future, for example, so this might not always be the only usage.

.vale.ini
```diff
@@ -1,6 +1,30 @@
-StylesPath = ".github/styles"
-Vocab = "Docs"
+# For more information, see ``.github/vale/README.rst``
+#
+# vale-action (https://github.com/errata-ai/vale-action) recommends
+# keeping the vale styles in the `.github` directory.
+# Since we have a README, styles, a dictionary, and some tests, we are
+# keeping related directories files in `.github/vale`
+#
```
lornajane, line 7: Remove this second paragraph below? I think "it's all working" now, right?

.github/workflows/lint.yaml
```diff
@@ -16,7 +16,9 @@ jobs:
       uses: actions/checkout@master
 
     - name: Vale
-      uses: errata-ai/vale-action@v1.3.0
+      # We want support for at least vale v2.15.3
+      #uses: errata-ai/vale-action@v1.3.0
+      uses: errata-ai/vale-action@master
```
lornajane, line 21: It was pinned because we had a breaking change at some point. Might be worth pinning again, and updating intentionally but I don't feel strongly about that.

(please remove commented out line 20, source control can give us that back if we need it)

.github/vale/styles/Aiven/aiven_spelling.yml
```diff
@@ -0,0 +1,8 @@
+extends: spelling
+message: "'%s' seems to be a typo"
```
lornajane, line 2: how about "seems not to be a recognised word" ? It's not necessarily mis-typing

.github/vale/README.rst
```diff
@@ -0,0 +1,456 @@
+========================
```
lornajane, line 1: Please revisit this file and condense it. As it is, people can't read all of it. What information is needed:
 - if you don't know what Vale is
 - if you're tackling a build failure (probably how most people meet it)
 - for a maintainer dealing with some more advanced feature

Could some of this information be linked to on Vale's own docs, for example? My concern here is that we aren't publishing something useful to the audience, and it's at a level of depth that we won't be able to actively maintain without a lot of deep work. I think 800 words max would be a sensible goal.

lornajane (changes requested):
The new Vale setup looks solid, and the text changes look right, I'm requesting changes on this PR but that sort of masks that I am VERY PLEASED with how this is coming along!

I can kind of squint and overlook the issues file, but READMEs are intended to be read by end users, and the vale README contains everything you know about Vale at this point in time. Please revisit to make it just what other people need to know, plus if you like a more in-depth section of notes for maintainers? It must be more concise for it to be useful to someone who didn't write it. I don't think we should merge it as-is, although I'm sorry to block progress on it

.github/vale/README.rst
TibsAtWork: I definitely agree.
There are three things here:

1. The useful kernel of things people want to know, hidden (as you say) amongst:
2. Notes I made for myself as I was learning, and
3. Things that should be in the vale docs somewhere, but aren't yet.

The second and third need removing as they're not part of the job of this repository (well, a little info from 3 needs to be kept, for your second point, but not at this detail, and only at an "if this happens, here is how to fix it" level).

The "extra" text that gets removed can safely live in the git history, possibly with a tag to make it easy for me to find.

lornajane:
Fixes #524

docs/products/grafana/reference/plugins.rst
TibsAtWork: Added the three words like this to the dictionary, and made them "plain" in this file.

.vale.ini
TibsAtWork: Oh, yes, thanks.

.github/workflows/lint.yaml
TibsAtWork: I've changed the comment to explain better why we're using `master`.
And removed the commented out line - thanks for spotting that, it was a pure mistake.

.github/vale/styles/Aiven/aiven_spelling.yml
TibsAtWork: Thanks, that's much better. I'd been staring at this so long I couldn't think of good phrases.
The pedant in me make me choose "does not seem to be a recognised word" - complain again if that's really too long...

README.rst
TibsAtWork: Much shortened, basically to just refer to the vale README. Thanks.

.github/vale/README.rst
TibsAtWork: There's now a much shortened vale README.
I *think* it's a good first try, and will be interested in whether I should cut more stuff.

TibsAtWork:
I think this is ready to be looked at again, @lornajane

lornajane:
This looks good but rebasing it and running `make spell` gives me a LOT of errors. Is that expected?

TibsAtWork:
I'd not expect lots of errors after rebasing - I shall have a look later on.

TibsAtWork:
Rebased and two minor fixes.
Running `vale-2.15.5 docs` locally works, as does `make spell` (which uses vale 2.16.0 on my machine).
Running `vale-2.14.0 docs` will give lots of errors relating to ®.

PR #514, opened 2022-01-27
pg upgrade: ANALYZE is executed automatically these days

aihessu:

It was pointed out that ANALYZE is run automatically (checked from source code, yes it is). No need to instruct user to do it manually.


ftisiot (approved):
LGTM

PR #399, opened 2021-12-07
Migrate best practices

anelook:
fixes #395

PR #519, opened 2022-01-31
Reorganised Apache Kafka Howto section

### LINKED ISSUE #518: Create HowTo Subcategories for Apache Kafka content

(more information is useful here! Were you trying to do something particular?)

Apache Kafka HowTo is becoming pretty long, subcategories are needed to keep the content in order

[X] I plan to open a pull request for this

ftisiot:

The Apache Kafka HowTo section was a huge list of articles. I tried dividing the existing content in the following categories:
* Tools
* Security
* Administration tasks
* Integrations
* Development tasks

Looking for feedback on the categories and the article/category association

Fixes #518


TibsAtWork:
I agree it's a good idea to split things up a bit more, and I think the reorganisation you've chosen looks sensible.

Is it worth doing some renaming at the same time?

Some articles are using titles with ``-ing``, for instance "Creating Apache Kafka topics automatically" - is it worth sorting those out in this PR?

I had to look inside "Get the best from Apache Kafka" to work out what it was. Might "Recommended best practices for Apache Kafka" be a better title? (not sure about this one)

laysauchoa:
for long list of sections os sub sections, do we have a thought whether they should be alphabetic? or importance? just thinking here

ftisiot:
I fixed the `-ing` main titles as @TibsAtWork suggested. 
Regarding the Best practices title, we tried to avoid the "best practices" naming,  you can check the original PR https://github.com/aiven/devportal/pull/399

lornajane:
I think the developer one is a bit of a kitchen sink at the moment, configuring data retention is different to what an application developer does. Could we revisit the order of links in each section as well please? As an example, adding a user seems like a more likely/popular use case than using SASL auth (but I might be wrong about that). Also does this take into account the shape of the articles that we already know we need to add, such as things that aren't migrated from the knowledgebase yet, or the code samples that we haven't created yet?

ftisiot:
Let me book some time on monday

PR #530, opened 2022-02-02
Update postgresql jdbc driver version

snuyanzin:

Since there is an RCE in postgresql jdbc driver, it's better to have link to fixed version
https://nvd.nist.gov/vuln/detail/CVE-2022-21724


laysauchoa (approved):
thank you for the update!

PR #539, opened 2022-02-05
re-add trademark symbols in mermaid graphs

ngilles-aiven:
The ® symbol was breaking the `mermaid` diagrams. It's not entirely evident that it is necessary to have them in diagrams, but should that be the case, this PR (re-)includes them (and maybe more usefully shows that it's possible to use arbitrary characters or emoji in the diagrams).

### DIFF
--- a/docs/products/flink/howto/real-time-alerting-solution-cli.rst
+++ b/docs/products/flink/howto/real-time-alerting-solution-cli.rst
@@ -12,10 +12,10 @@ This example involves creating an Apache Kafka® source topic that provides a st
 
     graph LR;
 
-        id1(Kafka)-- metrics stream -->id3(Flink);
-        id2(PostgreSQL)-- threshold data -->id3;
-        id3-. filtered data .->id4(Kafka);
-        id3-. filtered data .->id5(PostgreSQL);
+        id1("Kafka®)-- metrics stream -->id3("Flink®");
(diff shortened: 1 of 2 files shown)


lornajane:
Could you please add a description to your pull request? This is required for all contributions.

Tagging @TibsAtWork who is already working on #510 to tackle the trademarks topic, but I'm not sure if that includes diagrams.

lornajane:
I forgot the most important thing: Thanks! Super happy to see you contributing :blush:

ngilles-aiven:
> but I'm not sure if that includes diagrams.

Yeah, that was part of what we were discussing with @TibsAtWork and @ftisiot.
So, if they're needed, here they are .... with the extra goods new that @ftisiot can start putting his 🍍 directly in the diagrams!

TibsAtWork:
Thanks very much for finding out how to do this - it's going to be useful later on whether we use this PR or not.

I *think* we're OK with the diagram as-is, because we name correctly in the text that references it. Same for not putting "Apache" in front of the names. I shall double check later.

TibsAtWork:
@ngilles-aiven I'm closing this because I'm pretty sure we don't need the `®` in this diagram, but I shall keep a bookmark to the issue for later on! Thanks again for working out what our problem was.

PR #541, opened 2022-02-07
Add java example for Redis

snuyanzin:

The PR adds a java example to connect to Redis


laysauchoa:
lgtm! thanks for adding the example, @snuyanzin 👍

laysauchoa (approved):

laysauchoa:
just side topic, but there is also a example folder where this example can be added to the collection: https://github.com/aiven/aiven-examples

do you think is a good idea to add it there as well @snuyanzin? I guess it would be very helpful for others!

snuyanzin:
As I understand here (devportal) there are examples which could be shown as one file (not sure about that however it is my impression based on what I've seen in this repo).

At the same time in aiven-examples there could be a bit more complex examples e.g. full ready to use/run small maven/gradle project with corresponding file structure.

If my understanding is correct then yes, a bit later I can add a gradle project there

laysauchoa:
Actually, I was wrong about and I think you are right. Just doubled checked! In the case there, the example should be a bit more complete so this would not fit so well there.

Issue #545, opened 2022-02-07
Use ® and ™ on the devportal front page


TibsAtWork:
https://github.com/aiven/devportal/issues/484 is already underway to sort out use of ® and ™ after registered/trademarked names within devportal.

https://github.com/aiven/devportal/issues/485 exists to add a footer with appropriate disclaimer text.

Since the index page of devportal is prominent, it seems worth sorting out its use of ® and ™ now. It probably also makes sense to explicitly add disclaimer text at the end, which can be removed when https://github.com/aiven/devportal/issues/485 is done.

PR #546, opened 2022-02-07
Add ® and ™ marks to index page

TibsAtWork:
Fix: https://github.com/aiven/devportal/issues/545

Since the index page is the first thing people see for devportal, it seems
worth fixing it now.

1. Add necessary ® and ™ marks, including the asterisk needed for Redis
   (note that Redis' own usage doesn't require this to be superscripted)
2. Split the service names in the opening paragraph over multiple lines
   to make it easier to maintain the list (it still displays as one),
3. Refer to M3 rather than M3DB
4. Tidy some of the sentences in the final list


lornajane (approved):
Great start! Thank you :)

PR #560, opened 2022-02-09
Update Community page without events list #465

FloorD:
Changed formatting of the [Community page](https://developer.aiven.io/docs/community.html) to: 
screencapture-localhost-8000-docs-community-html-2022-02-09-21_23_50

Addresses #465


laysauchoa (approved):
just added a small fix before merging! lgtm :)

PR #563, opened 2022-02-10
Add more ® to "front" matter

TibsAtWork:
These changes make the names visible in the ToC sidebar have appropriate marks.

1. Add ® to `title` elements in the _toc.yml. Not all of these get
   displayed, but it seems sensible to correct them anyway.
2. Add missing ® to ClickHouse in the main index.
3. Add ® to those product index pages whose titles are used for the
   ToC.
4. Fix a `tip` in the opensearch index page that was giving
   reStructuredText syntax errors.

### DIFF
--- a/_toc.yml
+++ b/_toc.yml
@@ -57,7 +57,7 @@ entries:
 
   # -------- PRODUCT-SPECIFIC --------
   - file: docs/products/kafka/index
-    title: Apache Kafka
+    title: Apache Kafka®
     entries:
       - file: docs/products/kafka/getting-started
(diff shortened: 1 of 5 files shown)


lornajane:
I've got some concerns about the readability of having the additional symbols in the navigation bar. Could we please check if it's clear enough to use these in the main page occurrence and not necessarily in sidebars (where we often shorten things for readability/recognisability) ?

TibsAtWork:
Closing, as we almost certainly don't *need* to do this, and it makes the sidebar less useful.

PR #572, opened 2022-02-11
Replace elasticsearch example with opensearch

laysauchoa:
The data related to OpenSearch is pointing to Elasticsearch code,
instead of OpenSearch on the page.

This should point to opensearch-py and using this library, so this
commit fix this. Plus, some small grammar errors.


anelook (approved):
Looks good to me :)

PR #573, opened 2022-02-14
devportal: add iops explain for MySQL service

Aiqin-Aiven:
some startup plans are using different vm types as PostgreSQL,
write a separate but similar one for MySQL.

### DIFF
--- a/docs/products/mysql/reference/plan-iops.rst
+++ b/docs/products/mysql/reference/plan-iops.rst
@@ -1,7 +1,7 @@
 IOPS capability of plans
 ========================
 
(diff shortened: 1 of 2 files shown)


laysauchoa (approved):
lgtm! merging only the MySQL changes :)

laysauchoa:
lgtm! Maybe I was not clear in my last comment but I did not mean to duplicate the tables between those two references. In any case, for this commit we can merge the IOPS related to MySQL only. 

I am bringing the extra changes to a new PR and you can take a look if the values are correct before merging it, what do you think?

PR #574, opened 2022-02-14
Fixed Kafka duplicated target and Tip warnings

ftisiot:

In the Kafka docs we had several warnings regarding duplicated link targets and a Tip with missing indentation.

Fixes 516


TibsAtWork (approved):
Good fixes.

PR #581, opened 2022-02-17
Basic script to count the pages on help and developer sites

lornajane:

Script that does two things, one after the other:
 - gets the sitemap of developer.aiven.io and counts the number of pages on it (this step produces a lot of output, which we don't need)
 - spiders help.aiven.io, reporting how many links were found on the site, and how many were redirects to developer.aiven.io (this is an adaptation of the existing search index builder script)

Ideally we would make this run on build or something, in a way we can get the results in a human-friendly way. This is the starting point.

### DIFF
--- a/page_stats.py
+++ b/page_stats.py
@@ -1,0 +1,58 @@
+from usp.tree import sitemap_tree_for_homepage
+from bs4 import BeautifulSoup
+from urllib.parse import urljoin
+import os
+import requests
+
+
+def index_pages(es, base_url):
+    successes = []
+    skips = []
+    fails = []
+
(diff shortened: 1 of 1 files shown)


ftisiot:
The code works in my machine, what's the next step, do you want to merge this and create a new task to include it in the build?

ftisiot (approved):
LGTM

laysauchoa:
I am sorry that I came too late to check this PR. I have only noticed that one dependency was missing to be able to run locally for me. So I am not sure if you want it to be included in the `requirements.txt` or not. it would make sense imo. I open a small PR for you to check.

PR #586, opened 2022-02-21
OpenSearch: Get data mapping with Python

laysauchoa:
This adds sample code in `Howto Get data mapping with Python` for OpenSearch code. 
I will make reference to the OpenSearch section `How to write search queries with Python`. So I'm adding tags to refer later.

I have changed the order to first: HTTP subjects, Python subjects and NodeJS. 

We should think in another PR how to organize this file structure better.

Fixes: https://github.com/aiven/devportal/issues/587

### DIFF
--- a/docs/products/opensearch/howto/sample-dataset.rst
+++ b/docs/products/opensearch/howto/sample-dataset.rst
@@ -144,10 +144,10 @@ OpenSearch Python client offers a helper called bulk() which allows us to send m
 
 Get data mapping with Python
 ----------------------------
-When the data has been sent to OpenSearch cluster on :ref:`load the data with Python <load-data-with-python>`, no data structure was specified.
-This is possible because OpenSearch has `dynamic mapping feature <https://opensearch.org/docs/latest/opensearch/rest-api/index-apis/create-index/#mappings>`_ that automatically add the fields to their type mapping.
+When we sent the data to OpenSearch cluster on :ref:`load the data with Python <load-data-with-python>`, no data structure was specified.
+This is possible because OpenSearch uses **dynamic mapping** to automatically add the fields to their type mapping. 
 
-To check the mapping definition, we will use ``get_mapping`` method with the index name as argument.
+Use ``get_mapping`` method with the index name as the argument to check the mapping definition of your data.
 
 .. code:: python
 
@@ -161,7 +161,40 @@ To check the mapping definition, we will use ``get_mapping`` method with the ind
 
     schema = mapping_data[INDEX_NAME]["mappings"][doc_type]
     print(f"Fields: {list(schema.keys())} \n")
-    pprint(schema, width=80, indent=0)
+    pprint(schema, width=50, indent=1)
+
+You should be able to see this output:
+
+.. code-block:: bash
+
+    Fields: ['calories', 'categories', 'date', 'desc', 'directions', 'fat', 'ingredients', 'protein', 'rating', 'sodium', 'title'] 
+
+And the mapping with the fields and their respective types.
+
+.. code-block:: bash
+
+    {'calories': {'type': 'float'},
+    'categories': {'fields': {'keyword': {'ignore_above': 256,
+                                        'type': 'keyword'}},
+                    'type': 'text'},
+    'date': {'type': 'date'},
+    'desc': {'fields': {'keyword': {'ignore_above': 256,
+                                    'type': 'keyword'}},
+            'type': 'text'},
+    'directions': {'fields': {'keyword': {'ignore_above': 256,
+                                        'type': 'keyword'}},
+                    'type': 'text'},
+    'fat': {'type': 'float'},
+    'ingredients': {'fields': {'keyword': {'ignore_above': 256,
+                                            'type': 'keyword'}},
+                    'type': 'text'},
+    'protein': {'type': 'float'},
+    'rating': {'type': 'float'},
+    'sodium': {'type': 'float'},
+    'title': {'fields': {'keyword': {'ignore_above': 256,
+                                    'type': 'keyword'}},
+            'type': 'text'}}
+
 
 
 .. seealso::
--- a/docs/products/opensearch/howto/sample-dataset.rst
+++ b/docs/products/opensearch/howto/sample-dataset.rst
@@ -40,6 +40,56 @@ Let's take a look at a sample recipe document:
         "sodium": 959.0,
     }
 
+Sample queries with HTTP client
+-------------------------------
+
+With the data in place, we can start trying some queries against your OpenSearch service. Since it has a simple HTTP interface, you can use your favorite HTTP client. In these examples, we will use `httpie <https://github.com/httpie/httpie>`_ because it's one of our favorites.
+
+First, export the ``SERVICE_URI`` variable with your OpenSearch service URI address and index name from the previous script:
+
+.. code:: bash
+
+    export SERVICE_URI="YOUR_SERVICE_URI_HERE/epicurious-recipes"
+
+1. Execute a basic search for the word ``vegan`` across all documents and fields:
+
+.. code:: bash
+
+    http "$SERVICE_URI/_search?q=vegan"
+
+2. Search for ``vegan`` in the ``desc`` or ``title`` fields only: 
+
+.. code:: bash
+
+    http POST "$SERVICE_URI/_search" <<< '
+    {
+        "query": {
+            "multi_match": {
+                "query": "vegan",
+                "fields": ["desc", "title"]
+            }
+        }
+    }
+    '
+
+3. Search for recipes published only in 2013:
+
+.. code:: bash
+
+    http POST "$SERVICE_URI/_search" <<< '
+    {
+        "query": {
+            "range" : {
+                "date": {
+                "gte": "2013-01-01",
+                "lte": "2013-12-31"
+                }
+            }
+        }
+    }
+    '
+
+.. _load-data-with-python:
 
 Load the data with Python
 -------------------------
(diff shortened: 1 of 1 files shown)


docs/products/opensearch/howto/sample-dataset.rst
```diff
@@ -40,6 +40,56 @@ Let's take a look at a sample recipe document:
         "sodium": 959.0,
     }
 
+Sample queries with HTTP client
```
anelook, line 43: Not sure why we moved it to the start, I think we need to load data first before being able to make requests

docs/products/opensearch/howto/sample-dataset.rst
```diff
@@ -89,54 +139,68 @@ OpenSearch Python client offers a helper called bulk() which allows us to send m
...
     python epicurious_recipes_import.py
 
-Sample queries with HTTP client
--------------------------------
 
-With the data in place, we can start trying some queries against your OpenSearch service. Since it has a simple HTTP interface, you can use your favorite HTTP client. In these examples, we will use `httpie <https://github.com/httpie/httpie>`_ because it's one of our favorites.
+.. _get-mapping-with-python:
 
-First, export the ``SERVICE_URI`` variable with your OpenSearch service URI address and index name from the previous script:
+Get data mapping with Python
```
anelook, line 145: With data mapping examples for both JS and Python the article grew. Let's move "Get data mapping" (with examples of JS and Python) to a separate article? and add a reference in this article.

anelook (changes requested):
Good examples! I added a couple of structural comments. I think the chunk of article about mapping is worth separating. And not sure about HTTP part going before we load the data, I think the flow there should be opposite.

docs/products/opensearch/howto/sample-dataset.rst
laysauchoa: I agree but imo a new structure maybe fit more in a different PR because the issue is happening in other places, so it would be nice if we can discuss in this issue first:
https://github.com/aiven/devportal/issues/591

Just some rough ideas to what we think is the best for structuring this part.

anelook:
Left the comment in the corresponding issue :)

laysauchoa:
cool, I think this can be merged if the content is good and we can rearrange the structure in a different PR with more clear ideas that we will discuss in the issues

docs/products/opensearch/howto/sample-dataset.rst
```diff
@@ -89,54 +90,68 @@ OpenSearch Python client offers a helper called bulk() which allows us to send m
...
+    'sodium': {'type': 'float'},
+    'title': {'fields': {'keyword': {'ignore_above': 256,
+                                    'type': 'keyword'}},
+            'type': 'text'}}
+
+
+
+.. seealso::
+
+    More about `OpenSearch mapping <https://opensearch.org/docs/latest/opensearch/rest-api/index-apis/create-index/#mappings>`_
```
anelook, line 153: Let's make it a whole sentence - "Read more about in the official OpenSearch Documentation", so that the readers are aware where the links brings them

docs/products/opensearch/howto/sample-dataset.rst
```diff
@@ -89,54 +90,68 @@ OpenSearch Python client offers a helper called bulk() which allows us to send m
...
 
-3. Search for recipes published only in 2013:
+    Fields: ['calories', 'categories', 'date', 'desc', 'directions', 'fat', 'ingredients', 'protein', 'rating', 'sodium', 'title'] 
 
-.. code:: bash
+And the mapping with the fields and their respective types.
+
+.. code-block:: bash
+
+    {'calories': {'type': 'float'},
```
anelook, line 127: Can we format the json chunk in more readable way? seems to be a bit chaotic now :)

docs/products/opensearch/howto/sample-dataset.rst
```diff
@@ -89,54 +90,68 @@ OpenSearch Python client offers a helper called bulk() which allows us to send m
...
--------------------------------
 
-With the data in place, we can start trying some queries against your OpenSearch service. Since it has a simple HTTP interface, you can use your favorite HTTP client. In these examples, we will use `httpie <https://github.com/httpie/httpie>`_ because it's one of our favorites.
+.. _get-mapping-with-python:
 
-First, export the ``SERVICE_URI`` variable with your OpenSearch service URI address and index name from the previous script:
+Get data mapping with Python
+----------------------------
+When we sent the data to OpenSearch cluster on :ref:`load the data with Python <load-data-with-python>`, no data structure was specified.
+This is possible because OpenSearch uses **dynamic mapping** to automatically add the fields to their type mapping.
```
anelook, line 99: We opted here to use dynamic mapping. I feel that it makes sense to be a bit more specific that we let OpenSearch to use dymanic mapping

anelook (changes requested):
Added couple more comments :) they are not deal breaker, but they will improve the reading experience.

docs/products/opensearch/howto/sample-dataset.rst
```diff
@@ -95,14 +95,14 @@ OpenSearch Python client offers a helper called bulk() which allows us to send m
 
 Get data mapping with Python
 ----------------------------
-When we sent the data to OpenSearch cluster on :ref:`load the data with Python <load-data-with-python>`, no data structure was specified.
-This is possible because OpenSearch uses **dynamic mapping** to automatically add the fields to their type mapping. 
+When we sent the data to OpenSearch cluster on :ref:`load the data with Python <load-data-with-python>`, no data structure was specified. 
+As default, OpenSearch uses **dynamic mapping** to automatically add the fields to their type mapping.
```
anelook, line 99: I feel that "as default" is not exactly right. It is when the schema is not provided it uses dynamic mapping.

docs/products/opensearch/howto/sample-dataset.rst
```diff
@@ -142,15 +139,12 @@ And the mapping with the fields and their respective types.
...
-                                    'type': 'keyword'}},
+    'title': {'fields': {'keyword': {'ignore_above': 256, 'type': 'keyword'}},
             'type': 'text'}}
 
-
-
 .. seealso::
 
-    More about `OpenSearch mapping <https://opensearch.org/docs/latest/opensearch/rest-api/index-apis/create-index/#mappings>`_ 
+    Read more about in the `Official OpenSearch Documentation <https://opensearch.org/docs/latest/opensearch/rest-api/index-apis/create-index/#mappings>`_.
```
anelook, line 147: This reads a bit wrong - did you mean "Read more about it" Also, I don't think we need to use capital case here. so it should be something like this -

Read more about dynamic mapping in the `official OpenSearch documentation <https://opensearch.org/docs/latest/opensearch/rest-api/index-apis/create-index/#mappings>`_.

docs/products/opensearch/howto/opensearch-and-nodejs.rst
```diff
@@ -222,46 +222,6 @@ See if you can find recipes with misspelled pineapple 🍍
 
 Even though there is a typo in the word "pineapple", you still got relevant results. Try other search terms and different values for ``fuzziness`` to understand better how fuzzy queries work. What is your favourite food ingredient typo?
 
-Find best match with multiple search words
```
anelook, line 225: Does this belong to this PR?

docs/products/opensearch/howto/sample-dataset.rst
```diff
@@ -89,54 +90,62 @@ OpenSearch Python client offers a helper called bulk() which allows us to send m
...
-3. Search for recipes published only in 2013:
+    Fields: ['calories', 'categories', 'date', 'desc', 'directions', 'fat', 'ingredients', 'protein', 'rating', 'sodium', 'title'] 
 
-.. code:: bash
+And the mapping with the fields and their respective types.
+
+.. code-block:: bash
+
+    {'calories': {'type': 'float'},
+    'categories': {'fields': {'keyword': {'ignore_above': 256, 'type': 'keyword'}},
```
anelook, line 128: The formatting still looks chaotic. Maybe arrange it a bit manually? so at least the main properties are aligned and it is easier to read it
Screenshot 2022-02-23 at 17 11 19

anelook (changes requested):

docs/products/opensearch/howto/sample-dataset.rst
laysauchoa: thanks, fixed!

docs/products/opensearch/howto/sample-dataset.rst
laysauchoa: can you see if it makes sense now?

docs/products/opensearch/howto/sample-dataset.rst
laysauchoa: I adjusted manually
Screenshot 2022-02-23 at 17 20 02

do you think is better?

anelook: this is better! The normal format is still way more readable - like here - https://en.wikipedia.org/wiki/JSON. I suspect we want to save some space, since this JSON object is long when formatted properly. There are also online formatters that can help.

docs/products/opensearch/howto/sample-dataset.rst
```diff
@@ -95,8 +95,7 @@ OpenSearch Python client offers a helper called bulk() which allows us to send m
 
 Get data mapping with Python
 ----------------------------
-When we sent the data to OpenSearch cluster on :ref:`load the data with Python <load-data-with-python>`, no data structure was specified. 
-As default, OpenSearch uses **dynamic mapping** to automatically add the fields to their type mapping. 
+When we sent the data to OpenSearch cluster on :ref:`load the data with Python <load-data-with-python>`, no data structure was specified. OpenSearch uses **dynamic mapping** to automatically add the fields to their type mapping when no data structure is specified which is our case.
```
anelook, line 98: We repeat information and it reads a bit heavy to me, tbh.

Why not to be shorter - "When no data structure is specified, like in our case, OpenSearch uses **dynamic mapping** to automatically detect the fields."

docs/products/opensearch/howto/opensearch-and-nodejs.rst
laysauchoa: sorry, bad commit and add wrong file

docs/products/opensearch/howto/sample-dataset.rst
laysauchoa: I see it. I think it was something from the command line looked fine but copying and pasting, it somehow broke the formatting. Please, would be possible to check how it is now? I want to avoid the long json schema because it would take too much space of the page.

anelook (approved):
Thanks a lot for adding the changes!

PR #588, opened 2022-02-21
Fix typo in PG replication slot name

kevinbowman-aiven:

There's a typo in the name of the replication slot - it's been copied from the subscription name, not the slot name.


ftisiot (approved):
LGTM

PR #558, opened 2022-02-09
Order plugin list Grafana

FloorD:
Add more context to https://developer.aiven.io/docs/products/grafana/reference/plugins.html with external links and plugin descriptions. 

screencapture-localhost-8000-docs-products-grafana-reference-plugins-html-2022-02-09-20_35_21

I need to figure a few things out before I can submit the PR. 👇 

Plugins I could not find in the plugin marketplace on grafana.com: 
- Bar chart
- Bar gauge
- Gauge
- Geomap
- Getting Started
- Google Cloud Monitoring
- Jaeger
- Logs
- Loki
- Microsoft SQL Server
- News
- Node Graph
- Stat
- State timeline
- Status history
- Tempo
- TestData DB
- Time series
- Welcome
- Zipkin

_Google BigQuery_, which version?
https://grafana.com/grafana/plugins/doitintl-bigquery-datasource/
https://grafana.com/grafana/plugins/grafana-bigquery-datasource/

_Plotly_, which version?
https://grafana.com/grafana/plugins/natel-plotly-panel/
https://grafana.com/grafana/plugins/ae3e-plotly-panel/

_Getting Started_... seems to be a mistake? Maybe _Welcome_ too?

SimpleJson, is no longer maintained by Grafana. Alternatives are [JSON](https://grafana.com/grafana/plugins/simpod-json-datasource/) or [JSON API](https://grafana.com/grafana/plugins/marcusolsson-json-datasource/).

PR #595, opened 2022-02-23
Add descriptions and grouping to the postgresql extensions list

lornajane:

Request from a customer and support team, adding one-line descriptions for each extension. Layout is inspired by #558 which gives more space for words than the table layout did. I also tried to group similar extensions together but I think this part could use some work.

### DIFF
--- a/.github/styles/Vocab/Docs/accept.txt
+++ b/.github/styles/Vocab/Docs/accept.txt
@@ -34,8 +34,9 @@ etcd
 failover
 fileset[s]?
 Flink
+[Gg]eocoder
 Geomap
-geospatial
+[Gg]eospatial
 GitHub
 [Gg]o
 Gradle
--- a/docs/products/postgresql/reference/list-of-extensions.rst
+++ b/docs/products/postgresql/reference/list-of-extensions.rst
@@ -88,8 +88,6 @@ Geographical features
 ``postgis_legacy`` 
     The extension is not packaged or supported as an extension by the PostGIS project. The extension package is provided by Aiven for Aiven users.
 
-``postgres_fdw`` - https://www.postgresql.org/docs/current/postgres-fdw.html
-    Foreign data wrapper to enable access to data stored in external PostgreSQL servers
 ``address_standardizer`` - https://postgis.net/docs/standardize_address.html
     Returns a standard address format from certain rules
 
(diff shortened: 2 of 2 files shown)


ftisiot:
A question for clarification: in this version we're showing the complete link vs the Grafana one where we had link labels as "GitHub" or "Grafana Docs".
I would be in favour of adding a link label, also because we could possibly add more links (e.g. for DBLink one to PG docs and one to our example in devportal) WDYT?

lornajane:
I think labelling links, especially when the label is going to be the same in every entry, is a bit like throwing information away, and from an accessibility point of view, you have no idea where you are going (I can't hover; a screenreader user would only get "Docs" as the prompt to choose). So unless it's a big problem, having the information here is my preference, although not to the extent that I would want to argue about it.

TibsAtWork:
I hadn't thought of the accessibility argument for the links, but having them explicit does make it a lot more obvious where they're going, at the penalty of a "messier" appearance. So I'm happy to go with the explicit URLs.

Do we want to make this a "style guide" decision for such collection-of-plugins/things pages? (and specifically, does that mean #558 would need to do the same - that's a bit more complicated because there are often two urls...)

I do like the gathering into categories - I feel more comfortable with some sort of "contents" for such a long page. And I think the categories make sense, although with minor comments:

* Should `postgres_fdw` and `dblink` go together as "cross database" tools? Two items is on the edge...
* Would it be worth moving some of them to a `deprecated` or something category? I'm looking at `intagg` and`pg_prometheus`. (Again, it's only 2 items at the moment, but moving them out of the main categories may make a bit more sense for things that one probably shouldn't/can't use)
* Also, the entry for `pg_primetheus` says it is replaced by `promscale`, but we don't have an entry for that.

Anyway, once you're happy with the list I would gladly approve the merge.

lornajane:
Promscale is technically open source but it doesn't work well at all without an additional proprietary-licensed extension, so we don't offer it. I also took the liberty of removing from the list a bunch of extensions that are only available up until PG 10 (nobody should be deploying that version or enabling un-upgradeable extensions today, and it will save us having to remember to update when this is end of life later in the year). So that was chkpass, pg_partman, plv8, plcoffee and plls.

I don't think we need rules about you must or must not use links in certain ways; messier is absolutely the right way to describe this and there will be times when we come down on the other side of the compromise fence. Consistency within a page is probably the key (and we can edit it again later if we have got it wrong!)

TibsAtWork (approved):
I think this works well.

PR #597, opened 2022-02-24
Updated PostgeSQL logical replication is supported by GCP CloudSQL

runwuf:
Updated PostgeSQL logical replication is supported by GCP CloudSQL since Aug 30, 2021
https://cloud.google.com/sql/docs/release-notes#August_30_2021

### DIFF
--- a/docs/products/postgresql/concepts/aiven-db-migrate.rst
+++ b/docs/products/postgresql/concepts/aiven-db-migrate.rst
@@ -8,7 +8,7 @@ Logical replication is the default method and once successfully set up, this kee
 Regardless of the migration method used, the migration tool first performs a schema dump and migration to ensure schema compatibility.
 
 .. Note::
-    Logical replication also works when migrating from AWS RDS PostgreSQL 10+. Google Cloud Platform's PostgreSQL for CloudSQL does not support logical replication.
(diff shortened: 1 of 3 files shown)


paulgrainger-aiven (approved):

ftisiot (approved):
Lgtm

kevinbowman-aiven:
I think `aiven-db-migrate` needs more investigation - I had a problem with it where GCP needs the `pglogical` extension to enable logical replication, but `aiven-db-migrate` does a pre-flight-check that it can enable all extensions in the destination (and Aiven doesn't have the `pglogical` extension) so out-of-the-box it won't work.

ftisiot:
Thanks! Should we then create a separate article for GCP?

kevinbowman-aiven:
I have no opinion on that :)

In terms of things I'm pretty sure of, though:
* Logical replication from GCP into Aiven is possible by following the first half of https://cloud.google.com/sql/docs/postgres/replication/configure-external-replica up until (but not including) "Configure the external replica", and then switch to the instructions at https://developer.aiven.io/docs/products/postgresql/howto/setup-logical-replication.html
* aiven-db-migrate at best needs a bit more testing so we know which switches to use, and at worst doesn't work in this use-case without some changes

ftisiot:
I believe if the GCP is very different (e.g. we need to state more than 2 times "if GCP then don't/do this") then I believe a separated document (with links between the existing articles) would be my suggestion

Issue #598, opened 2022-02-24
Add dependencies to run `page_stats.py`


laysauchoa:
**How to reproduce?**
1) create a fresh environment
2) install the requirements.txt
3) Run `python3 page_stats.py`

Does it make to add this to `requirements.txt` file?

**Solved by:**
pip install ultimate-sitemap-parser

**Traceback:**
```
❯ python3 page_stats.py          
Traceback (most recent call last):
  File "/Users/laysa.uchoa/Documents/devportal/page_stats.py", line 1, in <module>
    from usp.tree import sitemap_tree_for_homepage
ModuleNotFoundError: No module named 'usp'
```

PR #599, opened 2022-02-24
add lib to run `page_status.py`

laysauchoa:
Add required lib to be able
to run the `page_stats.py` locally

Fixes: https://github.com/aiven/devportal/issues/598

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -122,6 +122,19 @@ To bring in a page from the previous platform:
 * Run ``python import-help-article.py [paste a URL]``
 * Take the resulting ``*.rst`` file and any images, and place them as appropriate in the file structure of the project
 
+Migration status
+-----------------
+
+You can check the migration status from ``https://help.aiven.io/en`` articles to the ``https://developer.aiven.io/`` articles by using the ``page_stats.py`` script available in this repository. 
+
+Install the dependencies::
+
+    pip install -r requirements-dev.txt
+
+To run the ``page_status.py`` script::
+
+    python page_stats.py    
+
 License
 -------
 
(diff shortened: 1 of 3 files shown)


laysauchoa:
related to: https://github.com/aiven/devportal/pull/581

lornajane:
Does pip have a concept of development requirements? I intentionally omitted this from the requirements because if it can't be installed, the site still works. It really isn't a requirement, and we would want the build to fail be of a problem with it

laysauchoa:
normally there is a dev requirements where those dependencies can be placed like test dependencies etc that the user does not to know if they want to use the library. I see that there are other libraries there related to this script, for example: ``BeautifulSoup`` but it is used in indexing scripts as well. I think I need help to divide which dependencies would go to this file and which dependencies would stay in the ``requirements.txt``.

Should we make sure that running dev requirements would install both dependencies? or should they be isolated?

Would be a good idea to include in the README that there is specific dependencies for development? or more info about the scripts?

lornajane:
I think the separate file and a `README` section would be helpful if we want to formalise the inclusion of the script in the project. I didn't do a whole lot with it because it is temporary, probably only you and I will run this to provide numbers on the migration, then we can remove the script when the knowledgebase is decommissioned.

I suggest we add a section to explain that we have this short-term need for migration metrics, and that the new requirements file and the script itself are needed. There's also a feature that is worth mentioning, set the `INDEX_DEBUG` variable to get the detailed output of the pages (similar to what the indexing script outputs). Sometimes, the extra info is useful.

lornajane (commented):
I think this needs a README update to go with it, you are really right that it probably should not just be a script by itself.

laysauchoa:
would that be a good solution for this script (even if temporary)?

lornajane (approved):
Thanks for taking the time on this!

PR #601, opened 2022-02-24
fixed Console instead of CLI references in ES and OS connectors

ftisiot:

We moved from CLI to Console examples for the sink ES and OS connectors but some of the old naming was kept in titles and descriptions


laysauchoa (approved):
lgtm!

PR #607, opened 2022-03-01
Don't build includes by themselves, and fix some minor build warnings

lornajane:

I noticed that my local build was showing some warnings when I went to fix a problem where we were indexing the include files without their surrounding content. So now I've stopped it from building includes as standalone HTML files, and fixed a few build warnings (not errors, but we should keep things tidy).


laysauchoa (approved):
build is clean, thanks

Issue #606, opened 2022-03-01
The SEARCH.md page talks about using Elasticsearch.


TibsAtWork:
The SEARCH.md page describes how search is handled for devportal. It talks about this in terms of Elasticsearch.

Presumably we are now using OpenSearch for this purpose, and so this page should be updated.

lornajane:
That's correct! We should update the docs (and possibly our dependencies)

PR #611, opened 2022-03-02
fix/scripts:use OpenSearch, not Elasticsearch

laysauchoa:
The SEARCH.md page describes how search is
handled for devportal. It uses Elasticsearch to do
the process.

Elasticsearch EOL as an aiven service is about to
reach in 23.03.2022; therefore, it makes sense to
replace the process using OpenSearch.

We are replacing here the Elasticsearch clients and
using the OpenSearch client.

Also, requirements are updated.

Fixes: https://github.com/aiven/devportal/issues/606

### DIFF
--- a/SEARCH.md
+++ b/SEARCH.md
@@ -43,15 +43,15 @@ The index is described with the following JSON:
 | description   | Short document description, used in help center results under the title text                                                                             |
 | content       | Main document content                                                                                                                                    |
 | source        | Document source, currently either `devportal` or `helpcenter`                                                                                            |
-| sort_priority | Document sort priority in search results, smaller priority documents are shown first and same priority documents are sorted by Elasticsearch query score |
+| sort_priority | Document sort priority in search results, smaller priority documents are shown first and same priority documents are sorted by OpenSearch query score |
 
 In addition to these `url` field should be provided with every document but it is not indexed or used in search queries. `url` is used in search result `href`.
 
 ## Search query
 
-The Elasticsearch index is used through the Netlify function in [netlify/functions/search/search.js](netlify/functions/search/search.js). The exact query can be found in the file but the idea is:
+The OpenSearch index is used through the Netlify function in [netlify/functions/search/search.js](netlify/functions/search/search.js). The exact query can be found in the file but the idea is:
 
-- use OR-style query matching `title`, `description` and `content` (with ES `match_phrase_prefix`)
+- use OR-style query matching `title`, `description` and `content` (with OS `match_phrase_prefix`)
 - give higher value to `title`
 
 # Creating the index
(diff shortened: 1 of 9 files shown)


SEARCH.md
```diff
@@ -43,23 +43,23 @@ The index is described with the following JSON:
...
-- use OR-style query matching `title`, `description` and `content` (with ES `match_phrase_prefix`)
+- use OR-style query matching `title`, `description` and `content` (with OS `match_phrase_prefix`)
 - give higher value to `title`
 
 # Creating the index
 
 The index is created with [scripts/create_index.py](scripts/create_index.py). You can run it with
 
 ```
-make create-index ES_URL=https://es.url/here
```
anelook, line 62: There are several more mentions of https://es.url/here, better to replace all of them and be consistent

anelook (commented):
The changes look good to me! 
How do we properly test this? Based on Netlify preview - search works. 

BTW, for some reason Netlify published twice info with preview, any reason for this?

lornajane:
I would test this by doing:
 - spin up an opensearch service on Aiven
 - set `ES_URL` environment variable in your terminal
 - run the indexing scripts
 - check that we have the data you'd expect

I think the preview deploys use the live netlify functions so the search functionality won't pick up changes there. There's a way to run the netlify functions locally, but you should check the netlify docs for that part, I've forgotten how it goes together.

laysauchoa:
> The changes look good to me! How do we properly test this? Based on Netlify preview - search works.
> 
> BTW, for some reason Netlify published twice info with preview, any reason for this?

they refer to different commits, it could be because I force pushed in the beginning.

I could see the same behavior here: https://github.com/aiven/devportal/pull/618

I tested as Lorna mentioned above but I did not figured out how to test with netlify locally. I gave some trials. The scripts are indexing the data correct to OpenSource. 

Screenshot 2022-03-07 at 11 12 49

SEARCH.md
laysauchoa: I could only find two in the docstrings. https://github.com/aiven/devportal/pull/611/commits/40b7f2da30c48d853a97691fbca20fd1e165ab67

lornajane:
I will try the test steps unless @anelook beats me to it, but let's deploy it together as I don't think we have documented rollback for this platform.

anelook:
@lornajane, I haven't look into running netlify locally yet, so probably will take me longer to figure it out (though I haven't read their docs yet - maybe it is indeed simple).

lornajane:
(removed the help wanted label as we have the SAs watching that tag and we need time rather than rescue)

this isn't simple to review and it's important to get it right so let's just keep trying to find the time to test it properly. This is on my radar but it needs a little bit more time than I had in the last few days. But it's an important change and we will find a way!

laysauchoa:
ok, clear! thanks @lornajane

lornajane:
I got to test this locally today so I'll write up how that works and plan to approve and deploy while I'm at my desk tomorrow ... Watch this space

laysauchoa:
thanks @lornajane would be possible to add links or write up after how you tested with netlify locally? I did not had time to look into it, but my future self may need this information.

lornajane (approved):
Thanks Laysa! Not an easy one to test but I think it's in good shape so I'm merging :)

PR #612, opened 2022-03-02
Provide links to onward resources from the upgrading to opensearch page

lornajane:
As well as upgrading the server, we have a lot of resources about the OpenSearch services, so add some links to the article (this is also where we link to when sending emails about the EOL for the old Elasticsearch service)

### DIFF
--- a/docs/products/opensearch/howto/upgrade-to-opensearch.rst
+++ b/docs/products/opensearch/howto/upgrade-to-opensearch.rst
@@ -47,4 +47,4 @@ Here are some related resources that may help you with your upgrade process
 * :doc:`Understand how OpenSearch relates to Elasticsearch <../concepts/opensearch-vs-elasticsearch>`
 * :doc:`Upgrade your application client libraries for OpenSearch <upgrade-clients-to-opensearch>`
 * :doc:`Plugins available on for Aiven for OpenSearch <../reference/plugins>`
-* :doc:`OpenSearch Dashboards <../dashboards/index>` replaces Kibana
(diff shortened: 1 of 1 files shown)


docs/products/opensearch/howto/upgrade-to-opensearch.rst
```diff
@@ -39,3 +39,12 @@ To perform the upgrade, visit the service overview page and choose the button la
...
 
+More resources
+--------------
+
+Here are some related resources that may help you with your upgrade process
+
+* :doc:`Understand how OpenSearch relates to Elasticsearch <../concepts/opensearch-vs-elasticsearch>`
+* :doc:`Upgrade your application client libraries for OpenSearch <upgrade-clients-to-opensearch>`
+* :doc:`Plugins available on for Aiven for OpenSearch <../reference/plugins>`
+* :doc:`OpenSearch Dashboards <../dashboards/index>` replaces Kibana
```
anelook, line 50: Only a minor comment - would it be better for the last one  - 
```suggestion
* :doc:`OpenSearch Dashboards <../dashboards/index>` as a replacement to Kibana
```

anelook: just to keep all phrases in similar style.

anelook (commented):
Just a small comment. If you feel that as it is, it reads better, I'm fine with it as well

lornajane:
Hey, the suggestion interface is "clickable" to me now! And your suggestion was a good one, thanks

anelook (approved):

PR #613, opened 2022-03-02
Add article about ClickHouse indexing and data processing

anelook:
This is information that I cut from the blog post, as it dives a bit deeper into how ClickHouse works.
I shaped it to focus on the primary index and data processing.

### DIFF
--- a/docs/products/clickhouse/concepts/indexing.rst
+++ b/docs/products/clickhouse/concepts/indexing.rst
@@ -1,21 +1,44 @@
 Indexing and data processing
 =============================
 
-In this article we look at how ClickHouse works with data and how this differs from other database management systems.
+In this article we look at how ClickHouse works with data and how this differs from other database management systems (DBMS).
 In particular we'll touch on such topics as **data indexing** and **a vector computation engine**.
 
-Understanding ClickHouse approach to data processing will help you learn why ClickHouse is perfect solution for some use cases and an awful choice for others.
+.. tip:: Understanding ClickHouse approach to data processing will help you learn why ClickHouse is perfect solution for some use cases and an awful choice for others.
 
-ClickHouse is :doc:`a columnar database <columnar-databases>`, this means that ClickHouse can read columns selectively retrieving only necessary information and omitting columns that are not needed for the request. The smaller the number of the columns you read, the faster and more efficient is the performance of the request. If you have to read many or all columns, then using columnar database becomes highly ineffective and we recommend reconsidering this solution.
+ClickHouse is a columnar database. This means that ClickHouse can read columns selectively retrieving only necessary information and omitting columns that are not needed for the request. The smaller the number of the columns you read, the faster and more efficient is the performance of the request. If you have to read many or all columns, then using columnar database becomes highly ineffective and we recommend reconsidering this solution.
+
+.. note:: Read more about columnar databases :doc:`in our separate article <columnar-databases>`.
 
 ClickHouse is designed to process massive chunks of information for a single request. To make this possible it shifts the focus from reading individual lines into scanning massive blocks of data for a request. This happens by using what is called **a vector computation engine**, which reads data in blocks that normally consist of thousands of rows, selected for a small set of columns.
 
+ClickHouse primary index
+------------------------
+
 To find a specific block, ClickHouse uses **a primary index**. However, unlike in other systems the primary index in ClickHouse has a couple of distinctive characteristics.
 
-First of all, not every row is indexed. By contrast, indexing happens at every ten thousandth's row (to be precise, **8192nd** is the default suggested value). Such type of index is descriptively called **a sparse index**.
+First of all, not every row is indexed. By contrast, ClickHouse indexes every 10K's row (or, to be precise, the **8192nd**, when using default settings). Such type of index is descriptively called **a sparse index** and a batch of rows is sometimes called **a granule**.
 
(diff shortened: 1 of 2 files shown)


docs/products/clickhouse/concepts/indexing.rst
```diff
@@ -0,0 +1,21 @@
+Indexing and data processing
+=============================
+
+In this article we look at how ClickHouse works with data and how this differs from other database management systems.
+In particular we'll touch on such topics as **data indexing** and **a vector computation engine**.
+
+Understanding ClickHouse approach to data processing will help you learn why ClickHouse is perfect solution for some use cases and an awful choice for others.
```
ftisiot, line 7: should this be a tip?

docs/products/clickhouse/concepts/indexing.rst
```diff
@@ -0,0 +1,21 @@
...
+
```
ftisiot, line 8: I would put a subtitle here like `Columnar Database` and expand a bit.

You might compare it with the traditional databases storing a row at the time.

docs/products/clickhouse/concepts/indexing.rst
```diff
@@ -0,0 +1,21 @@
...
+ClickHouse is :doc:`a columnar database <columnar-databases>`, this means that ClickHouse can read columns selectively retrieving only necessary information and omitting columns that are not needed for the request. The smaller the number of the columns you read, the faster and more efficient is the performance of the request. If you have to read many or all columns, then using columnar database becomes highly ineffective and we recommend reconsidering this solution.
+
+ClickHouse is designed to process massive chunks of information for a single request. To make this possible it shifts the focus from reading individual lines into scanning massive blocks of data for a request. This happens by using what is called **a vector computation engine**, which reads data in blocks that normally consist of thousands of rows, selected for a small set of columns.
+
+To find a specific block, ClickHouse uses **a primary index**. However, unlike in other systems the primary index in ClickHouse has a couple of distinctive characteristics.
```
ftisiot, line 13: I would add a subtitle like `ClickHouse primary index` and put the ordering sentence before, since the primary index mechanism described before works only if the column is ordered beforehand

docs/products/clickhouse/concepts/indexing.rst
```diff
@@ -0,0 +1,21 @@
...
+
+First of all, not every row is indexed. By contrast, indexing happens at every ten thousandth's row (to be precise, **8192nd** is the default suggested value). Such type of index is descriptively called **a sparse index**.
```
ftisiot, line 15:
```suggestion
First of all, not every row is indexed. By contrast, indexing happens for one row every 10.000 (to be precise, **1/8192** is the default suggested value). Such type of index is descriptively called **a sparse index**.
```

docs/products/clickhouse/concepts/indexing.rst
```diff
@@ -0,0 +1,21 @@
...
+
+Using sparse indexing has a significant consequence on advantages and limitations of ClickHouse, mainly because such primary key does not ensure uniqueness for a single searched item. Seeing that we index every ten thousandth's item, in order to find a specific individual row we need to iterating over thousands of items. That's why this approach is inadequate when working with individual rows and should be only used to deal when processing millions, or trillions of items.
```
ftisiot, line 17: Can you state a bit more a good example of this? With the current example I probably get where I shouldn't use it only.

Maybe we could use: `Select * from TBL where column > 42`, the index is a good choice. 
`Select * from TBL where column = 42` probably the index is not a good choice since it still needs to scan the whole block

ftisiot (changes requested):
Hi Olena, good piece and I def learnt something about ClickHouse. I requested few changes mainly about:
* Columnar database: probably explaining it against the usual row based databases makes the concept more digestible
* More examples on the indexing: to make clear when is a good fit and when not
* The order of sentences for the index: the primary index works only on top of a sorted column, so I believe we need to write the prereq first (sorted column) and then what you can do with it (primary index)

docs/products/clickhouse/concepts/indexing.rst
anelook: Changed

docs/products/clickhouse/concepts/indexing.rst
anelook: We have a separate article on it. I can add more, but then I'm duplicating it... Added a clear reference to the separate article. Do you think this makes it more clear?

if not, I can indeed add some more sentences about columnar db

anelook:
Thx Francesco!! helpful comments!
I added some changes to improve content.

Regarding "the primary index works only on top of a sorted column" - this is a tricky point, where I'm  trying to avoid misleading the readers. So I wonder how to explain it to stay technically correct.  This is a territory where I'm not super confident that I'm getting all details correct. But I updated a bit the paragraph about it.

anelook:
Also, decided that I cannot ignore the secondary index and added some short overview on it as well.
This is by no mean deep analysis, I referenced official docs for that

docs/products/clickhouse/concepts/indexing.rst
lornajane: I think the link addresses what is needed here

docs/products/clickhouse/concepts/indexing.rst
lornajane: I am lost as well, so I'm resolving the comment and hoping for the best!

lornajane (approved):
I learned something, thanks!

docs/products/clickhouse/concepts/indexing.rst
lornajane: This is better

docs/products/clickhouse/concepts/indexing.rst
lornajane: I _think_ I understand this now

PR #614, opened 2022-03-03
Describe instance type approach

kevinbowman-aiven:

Remove specifics about which instance types are used, and described the general approach instead.


lornajane (commented):
Minor question on this, We have IOPs in the original title, but the change to "resource capability" seems better. Could we remove IOPs from the title but then keep it (with its definition) in the body of the text so it still shows up when searching? Would that make sense?

kevinbowman-aiven:
Update title, text, and filename to lower the relevance of "IOPS"

lornajane (approved):
I think this is a good explanation without being too involved. Thanks for picking it up!

PR #616, opened 2022-03-03
fix/kafka: sink connectors integrations

laysauchoa:
Some pages point to aiven help io,
but we already have the devportal for them.

This commit changes this replacing to the right links


ftisiot (approved):
LGTM

PR #617, opened 2022-03-03
Added all versions of postgresql extensions

kevinbowman-aiven:

Updated list of PostgreSQL extensions available on Aiven.  List obtained by querying `pg_available_extensions` on each version, and then reformatted into tables on the page.

I'm not sure that this is the best way to present the data, so feel free to suggest alternatives.  Also, I'm not sure how best to keep this up to date, but the previous version was also out of date so at least it's a step forwards.


kevinbowman-aiven:
I just spotted some formatting issues, hang fire on this one...

kevinbowman-aiven:
Formatting issues should be fixed now

kevinbowman-aiven:
Vale now complaining about two things - "geocoding" (which [Wikipedia agrees with](https://en.wikipedia.org/wiki/Address_geocoding)) and "untrusted" (it wants me to use an initial capital, even though it's mid-way through a sentence?)

TibsAtWork:
I'll have a quick look at the vale issues - one moment...

TibsAtWork:
Something of a messy fix for "Untrusted", but we're changing the way we check spelling soon anyway.

kevinbowman-aiven:
Thank you!

TibsAtWork:
I really like the advice on how to find the available plugins.

In the [earlier PR](https://github.com/aiven/devportal/pull/595) that originally expanded this list, @lornajane decided not to include [extensions that are only available up until PG 10](https://github.com/aiven/devportal/pull/595#issuecomment-1049024972), as no-one should be using such versions. That included chkpass, pg_partman, plv8, plcoffee and plls.

I think it might be worth removing them again (are there any others?) and putting a note in the introduction at the top that extensions only usable in PostgreSQL 10 and earlier are not shown.

kevinbowman-aiven:
Hmm that's interesting, I should probably have realised that :D. This PR came from a discussion about a question from someone who wanted to know if Aiven had `pgpartman` which wasn't on the old list, but is available (although not really needed after v10), so it's definitely still generating questions.  Since v10 is still supported, personally I think it's still relevant info, but I'm happy to defer to experience on that.

TibsAtWork:
If the decision is made to keep them, I'd be tempted to put them in their own section (probably at the end), since most people won't be able to use them (so having them in the main part of the page is distracting), and if/when we stop supporting 10, we can easily remove that section.

ftisiot:
why not using a special label (like the beta one) to flag them?
I quite like them being organised by type, and making the EOL PG 10 as a label would make it prominent (and easily allow to identify the ones that we can remove once support for PG10 finishes)

kevinbowman-aiven:
Note that PG10 isn't EOL until 10th November - I've put a reminder in my calendar for that date to review the extensions, I'll happily remove the 10-only ones then.  I have no opinion about how it's formatted, I'm afraid.

TibsAtWork:
I like the idea of a label. 
If it goes that way, then I'd guess a label with text "PG10" and an explanation at the start that items marked with that label only apply to PostgreSQL 10, which will be EOL 2022-11-10.
If we're going to document things that will go away soon(ish), I'd rather be upfront about when that is.

If you want that, then near the top of the file (e..g, after the title) insert:
```
    .. |PG10| replace:: :badge:`PG10,cls=badge-secondary text-black badge-pill`
```
and at each place you want the label, put `|PG10|` - for instance:
```
``chkpass`` |PG10| - https://www.postgresql.org/docs/10/chkpass.html
    Data type for auto-encrypted passwords.  Not available after PostgreSQL 10.
```
(and yes, I've just tested that and it should work).

TibsAtWork:
I quite like that.
I also found the documentation for badges (finally) at https://sphinx-panels.readthedocs.io/en/latest/#link-badges, if you prefer a different style (although I think the grey is not too bad, as it doesn't draw attention as much as a colour).

TibsAtWork (approved):
Thank you very much for doing this

PR #619, opened 2022-03-04
New service configuration values from API

lornajane:

Ran the make commands to update the available configuration for each service from the current live API. Looks like we added prometheus parameters mostly, which seems good to know!


anelook (approved):
Thank you!

PR #621, opened 2022-03-04
Adding tfvars to Vale vocab

dewan-ahmed:

``tfvars`` is a specific file extension for Terraform. Requesting to add this word to Vale's *accept.txt* for builds to pass.

### DIFF
--- a/.github/styles/Vocab/Docs/accept.txt
+++ b/.github/styles/Vocab/Docs/accept.txt
@@ -121,6 +121,7 @@ TaskManager
 Telegraf
 Terraform
 Terraforming
+tfvars
 TimescaleDB
 toolchain
 Transport Layer Security


lornajane:
Where are we using tfvars as a word rather than in code/literal markup? @TibsAtWork will probably chime in but if it's not an English word (or technical term or prooper noun that we use), then it shouldn't be in `accept.txt` most of the time.

TibsAtWork:
I agree with Lorna.
The `accept.txt` file is a very blunt stick, and we'll be changing approach soon anyway, to a proper dictionary approach.
That makes me want to keep the dictionary for actual words, and anyway I think I'd be marking things like this up as
```
``tfvars``
```
in reStructuredText (the equivalent of `tfvars`) to indicate they're a technical artefact (the same way I'd talk about `.txt` files).

dewan-ahmed:
Thanks for the suggestion @lornajane & @TibsAtWork 

I've fixed the original PR using ``tfvars`` and am closing this PR as this is no longer required.

PR #625, opened 2022-03-08
Migrate Help documentation about static IP

cutinha:
Static IP Addresses
    https://help.aiven.io/en/articles/5538233-static-ip-addresses
    Troubleshooting DNS names pointing to Private IPs not Resolving
    https://help.aiven.io/en/articles/5885707-troubleshooting-dns-names-pointing-to-private-ips-not-resolving


docs/platform/howto/static-ip-addresses.rst
```diff
@@ -0,0 +1,108 @@
+Static IP addresses
```
ftisiot, line 1: What are we doing here? Maybe we should add a verb to the title

```suggestion
Use static IP addresses with Aiven services
```

docs/platform/howto/static-ip-addresses.rst
```diff
@@ -0,0 +1,108 @@
...
+===================
+
+**Note: Each static IP addresses has an hourly price ranging from $0.025
```
ftisiot, line 4: We use the Notes in devportal as follows 

```suggestion
.. Note:: 

   Each static IP addresses has an hourly price ranging from $0.025
```

docs/platform/howto/static-ip-addresses.rst
```diff
@@ -0,0 +1,108 @@
...
+to $0.075 depending on region. The feature is recent and
+Aiven's**\ `pricing page <https://aiven.io/pricing>`__\ **doesn't
+include static IPs yet.**
+
+By default an Aiven service will use public IP addresses allocated from
+the cloud provider's shared pool of addresses for the cloud region. As a
+result, Aiven service nodes effectively get a random IP address. Because
+the addresses are shared with all other users of the cloud region, it's
```
ftisiot, line 12: I find the part after "Because" hard to read, could you please try to rephrase it? Also please check that you're mentioning twice the granting access in the same sentence, maybe it's better to split it in two sentences?

docs/platform/howto/static-ip-addresses.rst
```diff
@@ -0,0 +1,108 @@
...
+not feasible to create firewall rules to limit access to/from nodes
+using public IP addresses as the firewall would need to grant access to
+the cloud region's entire IP range granting access to all other users of
```
ftisiot, line 15: Is this basically saying "Aiven can't do the job because we would need to have access to all IP ranges of the cloud region"?

docs/platform/howto/static-ip-addresses.rst
```diff
@@ -0,0 +1,108 @@
...
+that cloud region.
+
+You can however create IP addresses to your project. These addresses
+have a static IP for the lifetime of the address resource, and makes the
```
ftisiot, line 19: what is the lifetime of the address resource? Will it always be the same or not?

docs/platform/howto/static-ip-addresses.rst
```diff
@@ -0,0 +1,108 @@
...
+addresses predictable for firewall rules. After creating the address it
+can be associated with an Aiven service. Once enough IP addresses have
```
ftisiot, line 21: How many IP addresses do I need to associate before turning on the Static IP flag on the service?

docs/platform/howto/static-ip-addresses.rst
```diff
@@ -0,0 +1,108 @@
...
+been associated with a service, you can turn on static IP use on the
+service.
+
+.. _h_ff06143f3e:
```
ftisiot, line 25: if you are planning to reuse the anchors, please create proper names like `.. _platform_howto_setup_static_ip:`

docs/platform/howto/static-ip-addresses.rst
```diff
@@ -0,0 +1,108 @@
...
+
+Setting it up
```
ftisiot, line 27: We use the present in devportal
```suggestion
Reserve static IP addresses
```

docs/platform/howto/static-ip-addresses.rst
```diff
@@ -0,0 +1,108 @@
...
+-------------
+
+| Start by creating IP addresses into the cloud region you intend to use
```
ftisiot, line 30: I see this section being under a "Calculate the number of static IP addresses needed"

docs/platform/howto/static-ip-addresses.rst
```diff
@@ -0,0 +1,108 @@
...
+  them for a service. The amount of IP addresses depends on the service
+  plan. You'll need 2x the number of IP addresses compared to the amount
+  of nodes in the plan. This is so maintenance upgrades and failovers
```
ftisiot, line 33: Instead of "This is so maintenance upgrades and failovers can be handled without using non-static IP addresses"

maybe 

"We request a number of static IPs two times the amount of the cluster nodes, since it guarantees upgrades and failovers to be handled using only static IP addresses and not relying on random ones."

docs/platform/howto/static-ip-addresses.rst
```diff
@@ -0,0 +1,108 @@
...
+  can be handled without using non-static IP addresses. However no more
+  than 6 additional addresses will be used, so for example a 3 node plan
+  will need 2 \* 3 = 6 static IPs, but a 9 node plan will need 9 + 6 =
+  15 static IPs.
+| Static IPs are created to your project with
+
+::
+
+   # avn static-ip create --cloud azure-westeurope
```
ftisiot, line 42: We usually detach the command from the  output, see https://developer.aiven.io/docs/tools/cli/service/privatelink.html for an example

ftisiot: Also it would be good if you could add the cli examples in the documentation under `/docs/tools/cli`

docs/platform/howto/static-ip-addresses.rst
```diff
@@ -0,0 +1,108 @@
...
+   CLOUD_NAME        IP_ADDRESS  SERVICE_NAME  STATE     STATIC_IP_ADDRESS_ID
+   ================  ==========  ============  ========  ====================
+   azure-westeurope  null        null          creating  ip359373e5e56
+
+When the IP address has been provisioned, the state turns to ``created``
```
ftisiot, line 47: How can I check?

docs/platform/howto/static-ip-addresses.rst
ftisiot: Also, recently we agreed that all devportal examples should be based on the Aiven console (with a reference to the cli docs). Is this step achievable in the Console?

ftisiot (changes requested):
Hi @cutinha! Good work! Thanks for the contribution!
I would do the following:
* Move all the examples to be based on the console (instead of the cli)
* Document all the cli commands under /docs/tools/cli with a dedicated `static-ip.rst` file

I also checked the static-ip-addresses doc and found some minor bugs, could you please reiterate on the above and the comments? I'm happy to review a second draft! 

Thanks again!

cutinha:
@ftisiot considering this document is already a published document by a colleague, I was helping this getting migrated here and have done all possible changes I can do for now , any further queries would need input from the original author. thanks

lornajane:
@cutinha Thank you for giving this a try. We will do the rest of the work on this article.

docs/platform/howto/static-ip-addresses.rst
lornajane: It was a LOT more complicated but I figured it out! Explanation of how many IPs are needed added to the doc, and wording made clearer.

lornajane:
Blocking issues at this point are that we need:
 - a clearer concept piece about the static IPs, with some information about when the different states can be moved and how (by CLI, or by service reconfiguration) and what the states mean
 - a review and much better explanation of the private IPs problem in the other article in this PR

A follow up is #683  since the `avn static-ip` commands aren't included on DevPortal so I couldn't link to them. Not required for this PR though, we can add that stuff afterwards.

lornajane:
I have added enough to this PR now that I don't think I still qualify as a reviewer! @laysauchoa or @ftisiot could you take a look at this point?

PR #628, opened 2022-03-10
Add FlowCharting and OpenSearch to Grafana plugin list

FloorD:
Product added support for the FlowCharting panel plugin, and the OpenSearch data source plugin for Grafana. I've updated our reference list.


ftisiot (approved):
LGTM

PR #635, opened 2022-03-14
Article about dump data index from OS cluster

laysauchoa:
Article about dump data index from OS cluster

https://help.aiven.io/en/articles/3280782-moving-indexes-with-elasticsearch-dump-tool

Tested and works as well with OpenSearch (OS) v1.0.0

### DIFF
--- a/docs/products/opensearch/howto/import-opensearch-data-using-elasticsearch-dump.rst
+++ b/docs/products/opensearch/howto/import-opensearch-data-using-elasticsearch-dump.rst
@@ -34,11 +34,11 @@ Prerequisites
 Here are some information you need to collect to copy your data:
 
 OpenSearch cluster:
-* OpenSearch cluster url , for e.g. ``http://opensearch-url:9243/``
+* OpenSearch cluster URL , for e.g. ``http://opensearch-url:9243/``
 
 Aiven for OpenSearch:
 * ``INDEX_NAME``: the name of the index that you aim to backup.
-* ``SERVICE_URI``: OpenSearch service uri. You can find it in Aiven's dashboard.
+* ``SERVICE_URI``: OpenSearch service URI. You can find it in Aiven's dashboard.
 
 
 
@@ -85,7 +85,7 @@ You need to collect this information about your Aiven for OpenSearch cluster and
 
 Aiven for OpenSearch:
 
-* ``SERVICE_URI``: OpenSearch service uri, which you can find in Aiven's dashboard.
+* ``SERVICE_URI``: OpenSearch service URI, which you can find in Aiven's dashboard.
 * ``INDEX_NAME``: the name of the index that you aim to backup.
 
 AWS S3:
--- a/docs/products/opensearch/howto/import-opensearch-data-using-elasticsearch-dump.rst
+++ b/docs/products/opensearch/howto/import-opensearch-data-using-elasticsearch-dump.rst
@@ -1,12 +1,12 @@
 
-Dump OpenSearch index using ``elasticsearch-dump``
-==================================================
+Dump OpenSearch® index using ``elasticsearch-dump``
+===================================================
 
-It is a good practice to perform backups of your OpenSearch data to another storage service. This way you can access your data and restore it if something unexpected happens to it. 
+It is a good practice to perform backups of your OpenSearch® data to another storage service. This way you can access your data and restore it if something unexpected happens to it. 
 
 In this article, you can find out how to dump your OpenSearch data to an:
 
-* :ref:`Aiven OpenSearch <copy-data-from-os-to-os>`
+* :ref:`Aiven for OpenSearch® <copy-data-from-os-to-os>`
 * :ref:`AWS S3 bucket <copy-data-from-os-to-s3>`
 
 To copy the index data, we will be using ``elasticsearch-dump`` `tool <elashttps://github.com/elasticsearch-dump/elasticsearch-dump>`__. You can read the `instructions on GitHub <https://github.com/elasticsearch-dump/elasticsearch-dump/blob/master/README.md>`_ on how to install it. From this library, we will use ``elasticdump`` command to copy the input index data to an specific output. 
--- a/docs/products/opensearch/howto/import-opensearch-data-using-elasticsearch-dump.rst
+++ b/docs/products/opensearch/howto/import-opensearch-data-using-elasticsearch-dump.rst
@@ -72,8 +72,8 @@ When the dump is completed, you can check that the index is available in the Ope
 
 .. _copy-data-from-os-to-s3:
 
-Copying data from Aiven OpenSearch to S3
-----------------------------------------
+Copying data from Aiven OpenSearch to AWS S3
+--------------------------------------------
 
 Prerequisites
 ~~~~~~~~~~~~~
--- a/docs/products/opensearch/howto/import-opensearch-data-using-elasticsearch-dump.rst
+++ b/docs/products/opensearch/howto/import-opensearch-data-using-elasticsearch-dump.rst
@@ -9,7 +9,11 @@ In this article, you can find out how to dump your OpenSearch data to an:
 * :ref:`OpenSearch cluster <copy-data-from-os-to-os>`
 * :ref:`AWS S3 bucket <copy-data-from-os-to-s3>`
 
-We will be using ``elasticsearch-dump`` `tool <elashttps://github.com/elasticsearch-dump/elasticsearch-dump>`__ for these processes. To install the ``elasticsearch-dump``, read the `instructions <https://github.com/elasticsearch-dump/elasticsearch-dump/blob/master/README.md>`_ on GitHub.
+To copy the index data, we will be using ``elasticsearch-dump`` `tool <elashttps://github.com/elasticsearch-dump/elasticsearch-dump>`__. You can read the `instructions on GitHub <https://github.com/elasticsearch-dump/elasticsearch-dump/blob/master/README.md>`_ in how to install it. From this library, we will use ``elasticdump`` command to copy the ``input`` index data to an specific ``output``. 
+
+.. note::
+
+    Make sure to have ``elasticsearch-dump`` `tool <elashttps://github.com/elasticsearch-dump/elasticsearch-dump>`__ installed for the next steps.
 
 .. _copy-data-from-os-to-os:
 
@@ -19,33 +23,52 @@ Copying data from OpenSearch to Aiven OpenSearch
 Prerequisites
 ~~~~~~~~~~~~~
 
-* Aiven for OpenSearch cluster 
-* OpenSearch cluster 
-* ``elasticsearch-dump`` `tool <elashttps://github.com/elasticsearch-dump/elasticsearch-dump>`__ installed
+* OpenSearch cluster as the ``input`` (can be in Aiven or elsewhere)
+* Aiven for OpenSearch cluster as the ``output``
 
-To import the data we will use ``elasticdump`` command. This command works by copying the ``input`` index data to an specific ``output``. The ``input`` and ``ouput`` can be either an OpenSearch URL or a file path (local or remote file storage). In this particular case, we are using both URLs, one from an **OpenSearch cluster** and the other one from **Aiven for OpenSearch cluster**.
+.. note::
+    
+    The ``input`` and ``ouput`` can be either an OpenSearch URL or a file path (local or remote file storage). In this particular case, we are using both URLs, one from an **OpenSearch cluster** and the other one from **Aiven for OpenSearch cluster**. 
+
+
+Here are some information you need to collect to copy your data:
+
+OpenSearch cluster:
+* OpenSearch cluster url , for e.g. ``http://opensearch-url:9243/``
+
+Aiven for OpenSearch:
+* ``INDEX_NAME``: the name of the index that you aim to backup.
+* ``SERVICE_URI``: OpenSearch service uri. You can find it in Aiven's dashboard.
+
+
+
 
 Import mapping
 ~~~~~~~~~~~~~~
 
+The process of defining how a document and the fields are stored and indexed is called mapping. When no data structure is specified, we rely on OpenSearch to automatically detect the fields using dynamic mapping. However, we can set our data mapping before the data is sent:
+
 .. code-block:: shell
 
     elasticdump \
-    --input=http://opensearch-url:9243/index_name \
-    --output=SERVICE_URI/index_name \
+    --input=http://opensearch-url:9243/INDEX_NAME \
+    --output=SERVICE_URI/INDEX_NAME \
     --type=mapping
+
 
 Import data 
 ~~~~~~~~~~~
 
+This is how you can copy your index data from an OpenSearch cluster (can be in Aiven or elsewhere) to an Aiven for OpenSearch one.
+
 .. code-block:: shell
 
     elasticdump \
-    --input=http://opensearch-url:9243/index_name \
-    --output=SERVICE_URI/index_name \
+    --input=http://opensearch-url:9243/INDEX_NAME \
+    --output=SERVICE_URI/INDEX_NAME \
     --type=data
 
-When the dump is completed, you can check that the index is available in the OpenSearch service you send it to. You will able to find it under the **Indexes** tab in your Aiven console.
+When the dump is completed, you can check that the index is available in the OpenSearch service you send it to. You will be able to find it under the **Indexes** tab in your Aiven console.
 
 .. _copy-data-from-os-to-s3:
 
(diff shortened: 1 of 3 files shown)


laysauchoa:
**side information:** I've tested with both OpenSearch services and from OpenSearch to AWS s3 bucket.

docs/products/opensearch/howto/import-opensearch-data-using-elasticsearch-dump.rst
```diff
@@ -0,0 +1,80 @@
+
+Dump OpenSearch index using ``elasticsearch-dump``
```
ftisiot, line 2:
```suggestion
Backup OpenSearch indexes using ``elasticsearch-dump``
```

ftisiot: Or is this for Copying OS indexes?

docs/products/opensearch/howto/import-opensearch-data-using-elasticsearch-dump.rst
```diff
@@ -0,0 +1,80 @@
...
+==================================================
+
+It is a good practice to perform backups of your OpenSearch data to another storage service. This way you can access your data and restore it if something unexpected happens to it. 
+
+In this article, you can find out how to dump your OpenSearch data to an:
+
+* :ref:`OpenSearch cluster <copy-data-from-os-to-os>`
+* :ref:`AWS S3 bucket <copy-data-from-os-to-s3>`
+
+We will be using ``elasticsearch-dump`` `tool <elashttps://github.com/elasticsearch-dump/elasticsearch-dump>`__ for these processes. To install the ``elasticsearch-dump``, read the `instructions <https://github.com/elasticsearch-dump/elasticsearch-dump/blob/master/README.md>`_ on GitHub.
```
ftisiot, line 12: I try to stay away from `this` as much as possible and state clearly what we are doing
```suggestion
To copy the index data, we will be using ``elasticsearch-dump`` `tool <elashttps://github.com/elasticsearch-dump/elasticsearch-dump>`__. To install the ``elasticsearch-dump``, read the `instructions <https://github.com/elasticsearch-dump/elasticsearch-dump/blob/master/README.md>`_ on GitHub.
```

docs/products/opensearch/howto/import-opensearch-data-using-elasticsearch-dump.rst
```diff
@@ -0,0 +1,80 @@
...
+
+.. _copy-data-from-os-to-os:
+
+Copying data from OpenSearch to Aiven OpenSearch
+------------------------------------------------
+
+Prerequisites
+~~~~~~~~~~~~~
+
+* Aiven for OpenSearch cluster
```
ftisiot, line 22:
```suggestion
* Target Aiven for OpenSearch cluster 
```

docs/products/opensearch/howto/import-opensearch-data-using-elasticsearch-dump.rst
```diff
@@ -0,0 +1,80 @@
...
+.. _copy-data-from-os-to-os:
+
+Copying data from OpenSearch to Aiven OpenSearch
+------------------------------------------------
+
+Prerequisites
+~~~~~~~~~~~~~
+
+* Aiven for OpenSearch cluster 
+* OpenSearch cluster
```
ftisiot, line 23:
```suggestion
* Source OpenSearch cluster (can be in Aiven or elsewhere)
```

docs/products/opensearch/howto/import-opensearch-data-using-elasticsearch-dump.rst
```diff
@@ -0,0 +1,80 @@
...
+------------------------------------------------
+
+Prerequisites
+~~~~~~~~~~~~~
+
+* Aiven for OpenSearch cluster 
+* OpenSearch cluster 
+* ``elasticsearch-dump`` `tool <elashttps://github.com/elasticsearch-dump/elasticsearch-dump>`__ installed
+
+To import the data we will use ``elasticdump`` command. This command works by copying the ``input`` index data to an specific ``output``. The ``input`` and ``ouput`` can be either an OpenSearch URL or a file path (local or remote file storage). In this particular case, we are using both URLs, one from an **OpenSearch cluster** and the other one from **Aiven for OpenSearch cluster**.
```
ftisiot, line 26: Why don't we put the first sentence when we introduce ``elasticsearch-dump`` since it's common (and not specific to the OS -> OS migration)

docs/products/opensearch/howto/import-opensearch-data-using-elasticsearch-dump.rst
```diff
@@ -0,0 +1,80 @@
...
+
+Copying data from OpenSearch to Aiven OpenSearch
+------------------------------------------------
+
+Prerequisites
+~~~~~~~~~~~~~
+
+* Aiven for OpenSearch cluster 
+* OpenSearch cluster 
+* ``elasticsearch-dump`` `tool <elashttps://github.com/elasticsearch-dump/elasticsearch-dump>`__ installed
```
ftisiot, line 24: This is already stated, I would remove it

docs/products/opensearch/howto/import-opensearch-data-using-elasticsearch-dump.rst
```diff
@@ -0,0 +1,80 @@
...
+
+To import the data we will use ``elasticdump`` command. This command works by copying the ``input`` index data to an specific ``output``. The ``input`` and ``ouput`` can be either an OpenSearch URL or a file path (local or remote file storage). In this particular case, we are using both URLs, one from an **OpenSearch cluster** and the other one from **Aiven for OpenSearch cluster**.
+
+Import mapping
```
ftisiot, line 28: What does the `import mapping` does? I believe we need to explain what we are doing before showing code. Also can we use UPPER_CASE_PLACEHOLDERS and reference a  prerequisite section with all of them (https://developer.aiven.io/docs/products/kafka/kafka-connect/howto/jdbc-sink.html is a good example)

docs/products/opensearch/howto/import-opensearch-data-using-elasticsearch-dump.rst
```diff
@@ -0,0 +1,80 @@
...
+~~~~~~~~~~~~~~
+
+.. code-block:: shell
+
+    elasticdump \
+    --input=http://opensearch-url:9243/index_name \
+    --output=SERVICE_URI/index_name \
+    --type=mapping
+
+Import data
```
ftisiot, line 38: Same comment as before, we need a bit more wording to tell what this section is doing

docs/products/opensearch/howto/import-opensearch-data-using-elasticsearch-dump.rst
```diff
@@ -0,0 +1,80 @@
...
+~~~~~~~~~~~
+
+.. code-block:: shell
+
+    elasticdump \
+    --input=http://opensearch-url:9243/index_name \
+    --output=SERVICE_URI/index_name \
+    --type=data
+
+When the dump is completed, you can check that the index is available in the OpenSearch service you send it to. You will able to find it under the **Indexes** tab in your Aiven console.
```
ftisiot, line 48:
```suggestion
When the dump is completed, you can check that the index is available in the target Aiven for OpenSearch service. You will able to find it under the **Indexes** tab in your `Aiven Console <https://console.aiven.io/>`_.
```

docs/products/opensearch/howto/import-opensearch-data-using-elasticsearch-dump.rst
```diff
@@ -0,0 +1,80 @@
...
+
+.. _copy-data-from-os-to-s3:
+
+Copying data from Aiven OpenSearch to S3
```
ftisiot, line 52:
```suggestion
Backup data from Aiven OpenSearch to AWS S3
```

docs/products/opensearch/howto/import-opensearch-data-using-elasticsearch-dump.rst
```diff
@@ -0,0 +1,80 @@
...
+----------------------------------------
+
+Prerequisites
+~~~~~~~~~~~~~
+
+* Aiven for OpenSearch cluster
```
ftisiot, line 58:
```suggestion
* Source Aiven for OpenSearch cluster 
```

docs/products/opensearch/howto/import-opensearch-data-using-elasticsearch-dump.rst
```diff
@@ -0,0 +1,80 @@
...
+.. _copy-data-from-os-to-s3:
+
+Copying data from Aiven OpenSearch to S3
+----------------------------------------
+
+Prerequisites
+~~~~~~~~~~~~~
+
+* Aiven for OpenSearch cluster 
+* S3 bucket
```
ftisiot, line 59: What info do we need about the S3 bucket?
```suggestion
* Target S3 bucket
```

docs/products/opensearch/howto/import-opensearch-data-using-elasticsearch-dump.rst
```diff
@@ -0,0 +1,80 @@
...
+* ``elasticsearch-dump`` `tool <elashttps://github.com/elasticsearch-dump/elasticsearch-dump>`__ installed
```
ftisiot, line 60: This is a common prereq (also title of the article) I would remove it

docs/products/opensearch/howto/import-opensearch-data-using-elasticsearch-dump.rst
```diff
@@ -0,0 +1,80 @@
...
+
+    You can find the ``SERVICE_URI`` on Aiven's dashboard.
+
+
+.. code-block:: shell
+
+    elasticdump \
+    --s3AccessKeyId "${access_key_id}" \
+    --s3SecretAccessKey "${access_key_secret}" \
+    --input=SERVICE_URI/index_name --output "s3://${bucket_name}/${file_name}.json"
```
ftisiot, line 77: I would spend time describing the format of the s3 url in more detail and also the access Key and Id (you can find a reference https://developer.aiven.io/docs/products/kafka/kafka-connect/howto/s3-sink-connector-aiven.html)

ftisiot (changes requested):
Good work Laysa! 

I believe we need to add a bit more wording to various sections and align the command parameters to the standards used in other articles.

docs/products/opensearch/howto/import-opensearch-data-using-elasticsearch-dump.rst
laysauchoa: I think they are all correct, so I tried to add the most close to the tool which is dumping the indexes. I wonder what the user would look first... I am not 100% sure about backup without more context. Otherwise, I'd expect the backup to be in the same system.

laysauchoa: dump: copy (stored data) to a different location, especially so as to protect against loss.

laysauchoa: I do not think is indexes is the right word because I am showing how to move one index. Here you can find similar tools or articles:

https://github.com/ianneub/elasticsearch-to-s3
https://devcoops.com/dump-es-index-to-s3/

So dump seems to fit in this case.

docs/products/opensearch/howto/import-opensearch-data-using-elasticsearch-dump.rst
laysauchoa: ok, I added to the upper level cause it is relevant for both.

docs/products/opensearch/howto/import-opensearch-data-using-elasticsearch-dump.rst
laysauchoa: done!

docs/products/opensearch/howto/import-opensearch-data-using-elasticsearch-dump.rst
laysauchoa: sure!

docs/products/opensearch/howto/import-opensearch-data-using-elasticsearch-dump.rst
laysauchoa: in this context, I think they mean the same thing. But I can add AWS name.

docs/products/opensearch/howto/import-opensearch-data-using-elasticsearch-dump.rst
laysauchoa: I've added some general explanation and a link to AWS docs. 

I'd not aim to explain the specifics about AWS credentials because of:

- not the goal of the article. AWS docs do that better already too.
- expect the reader to be familiar with AWS terms

I looked up the aws documentation and I was able to figure out without previous knowledge. If we have more articles that involves dealing with those credentials, we can think about adding an external article for it or continue linking to the right ressources.

laysauchoa:
Thanks for the review @ftisiot. I have added the changes and hope it is more clear. I do not agree to add target and source around the document because for myself, in particular, it would be more confusing to understand. I'd prefer to reuse the commands terms than to add new words where the reader could not immediately relate to. 

Some of the suggestions are replacing word for their synonymous. If you can elaborate why do you feel that they can help to understand more, I can review them again.

docs/products/opensearch/howto/import-opensearch-data-using-elasticsearch-dump.rst
laysauchoa: I will explain why I am not sure this is a good idea in a general comment. So same for the next suggestions.

docs/products/opensearch/howto/import-opensearch-data-using-elasticsearch-dump.rst
```diff
@@ -0,0 +1,119 @@
...
+==================================================
+
+It is a good practice to perform backups of your OpenSearch data to another storage service. This way you can access your data and restore it if something unexpected happens to it. 
+
+In this article, you can find out how to dump your OpenSearch data to an:
+
+* :ref:`Aiven OpenSearch <copy-data-from-os-to-os>`
+* :ref:`AWS S3 bucket <copy-data-from-os-to-s3>`
+
+To copy the index data, we will be using ``elasticsearch-dump`` `tool <elashttps://github.com/elasticsearch-dump/elasticsearch-dump>`__. You can read the `instructions on GitHub <https://github.com/elasticsearch-dump/elasticsearch-dump/blob/master/README.md>`_ in how to install it. From this library, we will use ``elasticdump`` command to copy the ``input`` index data to an specific ``output``.
```
ftisiot, line 12: I would reference ``input`` and ``output``  as code only close to the command itself
```suggestion
To copy the index data, we will be using ``elasticsearch-dump`` `tool <elashttps://github.com/elasticsearch-dump/elasticsearch-dump>`__. You can read the `instructions on GitHub <https://github.com/elasticsearch-dump/elasticsearch-dump/blob/master/README.md>`_ on how to install it. From this library, we will use the ``elasticdump`` command to copy the input index data to a specific output. 
```

docs/products/opensearch/howto/import-opensearch-data-using-elasticsearch-dump.rst
```diff
@@ -0,0 +1,119 @@
...
+
+Here are some information you need to collect to copy your data:
+
+OpenSearch cluster:
+
+* OpenSearch cluster URL , for e.g. ``http://opensearch-url:9243/``
+
+Aiven for OpenSearch:
+
+* ``INDEX_NAME``: the name of the index that you aim to backup.
```
ftisiot, line 42: isn't this the target index name, not the one you aim to backup? Also can i backup an index named `students` into an index named `stud` or do they need to share the same name? 
If it's the first case (names can be different) I would use two different names like `INPUT_INDEX_NAME` and `OUTPUT_INDEX_NAME`

docs/products/opensearch/howto/import-opensearch-data-using-elasticsearch-dump.rst
```diff
@@ -0,0 +1,119 @@
...
+Copying data from Aiven OpenSearch to AWS S3
+--------------------------------------------
+
+Prerequisites
+~~~~~~~~~~~~~
+
+* Aiven for OpenSearch cluster as the ``input``
+* AWS S3 bucket as the ``output``
+
+You need to collect this information about your Aiven for OpenSearch cluster and your AWS service:
```
ftisiot, line 86:
```suggestion
You need to collect the following information about your Aiven for OpenSearch cluster and your AWS service:
```

ftisiot (changes requested):
Great improvement, I like the clarity now that you mentioned `input` and `output`
I added another couple of comments on the latest things worth an additional check

docs/products/opensearch/howto/import-opensearch-data-using-elasticsearch-dump.rst
laysauchoa: Good you mention! I had not tested this but I tested now and you can change the output index name (not necessarily to keep the same from the input one). I will make the changes.

docs/products/opensearch/howto/import-opensearch-data-using-elasticsearch-dump.rst
```diff
@@ -0,0 +1,117 @@
...
+
+Here are some information you need to collect to copy your data:
+
+OpenSearch cluster:
+
+* OpenSearch cluster URL , for e.g. ``http://opensearch-url:9243/``
+
+Aiven for OpenSearch:
+
+* ``INPUT_INDEX_NAME``: the index that you aim to copy from your input source.
```
ftisiot, line 42: Shouldn't the INPUT_INDEX_NAME be in the OpenSearch cluster section?

docs/products/opensearch/howto/import-opensearch-data-using-elasticsearch-dump.rst
```diff
@@ -0,0 +1,117 @@
...
+* Aiven for OpenSearch cluster as the ``output``
+
+.. note::
+    
+    The ``input`` and ``ouput`` can be either an OpenSearch URL or a file path (local or remote file storage). In this particular case, we are using both URLs, one from an **OpenSearch cluster** and the other one from **Aiven for OpenSearch cluster**. 
+
+
+Here are some information you need to collect to copy your data:
+
+OpenSearch cluster:
```
ftisiot, line 36: Can we also state here that this is the `input` OS cluster? And then the output in the following section?

docs/products/opensearch/howto/import-opensearch-data-using-elasticsearch-dump.rst
```diff
@@ -0,0 +1,117 @@
...
+
+* OpenSearch cluster URL , for e.g. ``http://opensearch-url:9243/``
```
ftisiot, line 38: Can we have this as ``INPUT_SERVICE_URI`` and refer to it in the below calls?

ftisiot (changes requested):
Last minor changes for consistency:
- I would refer to input and output when describing the prereqs
- I would make the `INPUT_SERVICE_URI` as variable as well

docs/products/opensearch/howto/import-opensearch-data-using-elasticsearch-dump.rst
laysauchoa: yes, we can! I will add it

docs/products/opensearch/howto/import-opensearch-data-using-elasticsearch-dump.rst
laysauchoa: Makes sense, I will move it

laysauchoa:
Actually, I looked again and it feels like a lot of repetition with the ``input`` and ``output`` words without bringing new information. I feel like this for the reasons:

- the names already indicate whether this is an input or not 
- the above section already indicates which ones are ``output`` and ``input``. The section is very close to this other references. For e.g.:

  * Aiven for OpenSearch cluster as the ``input``
  * AWS S3 bucket as the ``output``
  
  - a quick look in the command could also indicate which ones are referred to the ``input`` and ``output``. 
  
  
  That being said, I have decided to leave this information only in the above reference not the arguments specification. Hope it is okay. I just do not want to repeat too much the same thing. I also want to keep ``SERVICE_URI`` as it is because it is how we normally talk about this parameter and it is easier to relate or to find from the Aiven console if is the same name.

docs/products/opensearch/howto/import-opensearch-data-using-elasticsearch-dump.rst
laysauchoa: I explained in the general comment why I do not think this is necessary. I have actually cleaned up and removed thoe ``input`` and ``output``

docs/products/opensearch/howto/import-opensearch-data-using-elasticsearch-dump.rst
```diff
@@ -0,0 +1,117 @@
...
+~~~~~~~~~~~~~
+
+* Aiven for OpenSearch cluster as the ``input``
+* AWS S3 bucket as the ``output``
+
+You need to collect the following information about your Aiven for OpenSearch cluster and your AWS service:
+
+Aiven for OpenSearch:
+
+* ``SERVICE_URI`: OpenSearch service URI, which you can find in Aiven's dashboard.
```
ftisiot, line 88:
```suggestion
* ``SERVICE_URI``: OpenSearch service URI, which you can find in Aiven's dashboard.
```

ftisiot (commented):
One last small fix and this is ready to go!

ftisiot:
Added a couple of trademarks needed for the first and most prominent OpenSearch mentions

ftisiot (approved):
LGTM

PR #637, opened 2022-03-14
Switch to shorter title decoration

lornajane:

We had a request from the SEO agency to keep our titles under a certain length. As a first step, this reduces the trailing character count from 32 characters to 8. We should also revisit page titles to keep them short (ideal max is 60 chars)


TibsAtWork (approved):
Yes, that sounds sensible - looking at the sphinx docs, we're not losing anything useful by this.

PR #645, opened 2022-03-15
Remove *.rst from filename

dewan-ahmed:

The _toc.yml doesn't need *.rst at the end of `file`. Removing the *.rst references since some members are seeing WARNING from build due to this.


ftisiot (approved):
LGTM

PR #646, opened 2022-03-15
Terraform section layout and getting-started

dewan-ahmed:

This PR brings in the following changes:

1. Terraform section layout including HowTo, Reference, and a Getting Started page.
2. Removing old getting started from the landing page.
3. Adding a new getting-started page.
4. Moving the only HowTo article under a HowTo section.

### DIFF
--- a/docs/tools/terraform/get-started.rst
+++ b/docs/tools/terraform/get-started.rst
@@ -10,7 +10,28 @@ Requirements
 - `Generate an authentication token <https://developer.aiven.io/docs/platform/howto/create_authentication_token.html>`_ on Aiven's console or CLI
 
 
-1. ``provider.tf`` file:
+Configure your project and services
+'''''''''''''''''''''''''''''''''''
+
+Your Terraform file(s) consists of lines for the actual infrastructure as well as required dependencies and configurations. While you can stuff these together in one file, it's ideal to keep those as separate files.
+In this section, you'll learn how to structure a simple Terraform project. 
+
+1.  BYOTF - Bring Your Own Terraform Script. 
+
+``services.tf`` file:
+
+.. code:: bash
+
+   # Your actual Terraform script goes here 
+
+
+2. Consider the following code block similar to declaring a dependency; the *Aiven Terraform Provider* in this case. Within the `required_providers` block, you mention the source of the provider and specify a certain version. 
+The versions here are for example purpose and will change in the future. Following `Aiven Terraform Provider doc <https://registry.terraform.io/providers/aiven/aiven/latest/docs>`_, `api_token` is the only parameter for the provider configuration.
+
+Make sure that you have either the *Administrator* or *Operator* role when creating the API token. When you create a project, you automatically receive the *Administrator* access. 
+For more details, refer to the `Project members and roles page <https://developer.aiven.io/docs/platform/concepts/projects_accounts_access.html#project-members-and-roles>`_.
+
+``provider.tf`` file:
 
 .. code:: bash
 
@@ -28,12 +49,9 @@ Requirements
    }
 
 
-Consider this code block similar to declaring a dependency; the *Aiven Terraform Provider* in this case. Within the *required_providers* block, you mention the source of the provider and specify a certain version. The versions here are for example purpose and will change in the future.
-Following `Aiven Terraform Provider doc <https://registry.terraform.io/providers/aiven/aiven/latest/docs>`_, *api_token* is the only parameter for the provider configuration.
+3. Avoid including sensitive values in configuration files that are under source control. Rather than using sensitive information within this *variables.tf* file, you can use a ``*.tfvars`` file so that Terraform receives the values during runtime.
 
-Make sure the owner of the API Authentication Token has admin permissions in Aiven.
-
-2. ``variables.tf`` file:
+``variables.tf`` file:
 
 .. code:: bash
 
@@ -47,22 +65,19 @@ Make sure the owner of the API Authentication Token has admin permissions in Aiv
       type        = string
    }
 
-Avoid including sensitive values in configuration files that are under source control. Rather than using sensitive information within this *variables.tf* file, you can use a ``*.tfvars`` file so that Terraform receives the values during runtime.
 
-3. ``var-values.tfvars`` file:
+4. This is where you put the actual values for Aiven API token and Aiven console project name. This file is passed to Terraform using the ``-var-file=`` flag.
+
+``var-values.tfvars`` file:
 
 .. code:: bash
 
    aiven_api_token = "<YOUR-AIVEN-AUTHENTICATION-TOKEN-GOES-HERE>"
    project_name = "<YOUR-AIVEN-CONSOLE-PROJECT-NAME-GOES-HERE>"
 
-This is where you put the actual values for Aiven API token and Aiven console project name. This file is passed to Terraform using the ``-var-file=`` flag.
 
-4. ``services.tf`` file:
-
-.. code:: bash
-
-   # Your actual Terraform script goes here 
+Execution
+'''''''''
 
 Create an empty folder and add the above files to that folder. Then execute the following commands in order:
 
@@ -84,8 +99,9 @@ This command creates an execution plan and shows you the resources that will be 
 
 If you're satisfied with ``terraform plan``, you execute ``terraform apply`` command which actually does the task or creating (or modifying) your infrastructure resources. 
 
-Optional
---------
+
+Clean up
+''''''''
 
 If this was a test environment, be sure to delete the resources once you're done to avoid consuming unwanted bills. 
 
@@ -96,3 +112,9 @@ If this was a test environment, be sure to delete the resources once you're done
 .. code:: bash
 
    terraform destroy -var-file=var-values.tfvars
+
+
+Further reference
+'''''''''''''''''
+
+This article outlined a simple Terraform project structure. For a more complex project structure, please refer to the `Terraform Docs <https://www.terraform.io/language/modules/develop/structure>`_. 
--- a/docs/tools/terraform/get-started.rst
+++ b/docs/tools/terraform/get-started.rst
@@ -16,20 +16,34 @@ Configure your project and services
 Your Terraform file(s) consists of lines for the actual infrastructure as well as required dependencies and configurations. While you can stuff these together in one file, it's ideal to keep those as separate files.
 In this section, you'll learn how to structure a simple Terraform project. 
 
-1.  BYOTF - Bring Your Own Terraform Script. 
+1.  The following Terraform script deploys a single-node Redis service. This is a simple example which you can swap out with your own Terraform script(s) or other advanced recipes from :doc:`the Terraform cookbook <reference/cookbook>`.
 
 ``services.tf`` file:
 
 .. code:: bash
 
-   # Your actual Terraform script goes here 
+  # A single-node Redis service
+  
+   resource "aiven_redis" "single-node-aiven-redis" {
+  project                 = var.project_name
+  cloud_name              = "google-northamerica-northeast1"
+  plan                    = "startup-4"
+  service_name            = "gcp-single-node-redis1"
+  maintenance_window_dow  = "monday"
+  maintenance_window_time = "10:00:00"
+
+  redis_user_config {
+    redis_maxmemory_policy = "allkeys-random"
+
+    public_access {
+      redis = true
+    }
+  }
+}
 
 
-2. Consider the following code block similar to declaring a dependency; the *Aiven Terraform Provider* in this case. Within the `required_providers` block, you mention the source of the provider and specify a certain version. 
-The versions here are for example purpose and will change in the future. Following `Aiven Terraform Provider doc <https://registry.terraform.io/providers/aiven/aiven/latest/docs>`_, `api_token` is the only parameter for the provider configuration.
-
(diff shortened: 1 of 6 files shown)


docs/tools/terraform/getting-started.rst
```diff
@@ -0,0 +1,98 @@
...
+This example shows the setup for a sample Terraform project and some useful commands to stand up (and destroy) your Aiven data infrastructure.
+
+Requirements 
+''''''''''''
+- `Download and install Terraform <https://www.terraform.io/downloads.html>`_
+- `Sign up <https://console.aiven.io/signup?utm_source=github&utm_medium=organic&utm_campaign=devportal&utm_content=repo>`_ for Aiven if you haven't already
+- `Generate an authentication token <https://developer.aiven.io/docs/platform/howto/create_authentication_token.html>`_ on Aiven's console or CLI
+
+
+1. ``provider.tf`` file:
```
lornajane, line 13: What action are we doing in this step? Let's explain the what and why, then show the code.

docs/tools/terraform/getting-started.rst
```diff
@@ -0,0 +1,98 @@
...
+         }
+      }
+   }
+
+   provider "aiven" {
+      api_token = var.aiven_api_token
+   }
+
+
+Consider this code block similar to declaring a dependency; the *Aiven Terraform Provider* in this case. Within the *required_providers* block, you mention the source of the provider and specify a certain version. The versions here are for example purpose and will change in the future.
```
lornajane, line 31: I think `required_providers` is a variable and should be code/literal formatted. Also the `api_token` in the next paragraph

docs/tools/terraform/getting-started.rst
```diff
@@ -0,0 +1,98 @@
...
+Following `Aiven Terraform Provider doc <https://registry.terraform.io/providers/aiven/aiven/latest/docs>`_, *api_token* is the only parameter for the provider configuration.
+
+Make sure the owner of the API Authentication Token has admin permissions in Aiven.
```
lornajane, line 34: How do I make sure of this? Admin permissions on what?

docs/tools/terraform/getting-started.rst
```diff
@@ -0,0 +1,98 @@
...
+
+2. ``variables.tf`` file:
```
lornajane, line 36: What is happening in this step? Let's outline what happens here, then show the code.

docs/tools/terraform/getting-started.rst
```diff
@@ -0,0 +1,98 @@
+Set up your first Terraform project
+===================================
+
+This example shows the setup for a sample Terraform project and some useful commands to stand up (and destroy) your Aiven data infrastructure.
```
lornajane, line 4: What sort of project am I setting up? What does it do?

docs/tools/terraform/getting-started.rst
```diff
@@ -0,0 +1,98 @@
...
+      description = "Aiven console API token"
+      type = string
+   }
+
+   variable "project_name" {
+      description = "Aiven console project name"
+      type        = string
+   }
+
+Avoid including sensitive values in configuration files that are under source control. Rather than using sensitive information within this *variables.tf* file, you can use a ``*.tfvars`` file so that Terraform receives the values during runtime.
```
lornajane, line 50: Can we say WHY we don't include the values? I think also this setup declares the expected variables, then defines the actual values in a file so perhaps those could be part of the same step since they're sort of the same task. If either or both of these files should be excluded from source control, that would be good advice to include

docs/tools/terraform/getting-started.rst
```diff
@@ -0,0 +1,98 @@
...
+3. ``var-values.tfvars`` file:
+
+.. code:: bash
+
+   aiven_api_token = "<YOUR-AIVEN-AUTHENTICATION-TOKEN-GOES-HERE>"
+   project_name = "<YOUR-AIVEN-CONSOLE-PROJECT-NAME-GOES-HERE>"
+
+This is where you put the actual values for Aiven API token and Aiven console project name. This file is passed to Terraform using the ``-var-file=`` flag.
+
+4. ``services.tf`` file:
```
lornajane, line 61: This is a bit open-ended for actually getting started for a new person. Let's explain that you can use a number of `*.tf` files to describe your services, and perhaps provide the most simple thing (one Redis service?) in our example.

docs/tools/terraform/getting-started.rst
```diff
@@ -0,0 +1,98 @@
+Set up your first Terraform project
+===================================
+
+This example shows the setup for a sample Terraform project and some useful commands to stand up (and destroy) your Aiven data infrastructure.
+
+Requirements
```
lornajane, line 6: I think I'd call this "Prepare the dependencies"

docs/tools/terraform/getting-started.rst
```diff
@@ -0,0 +1,98 @@
...
+
+This example shows the setup for a sample Terraform project and some useful commands to stand up (and destroy) your Aiven data infrastructure.
+
+Requirements 
+''''''''''''
+- `Download and install Terraform <https://www.terraform.io/downloads.html>`_
+- `Sign up <https://console.aiven.io/signup?utm_source=github&utm_medium=organic&utm_campaign=devportal&utm_content=repo>`_ for Aiven if you haven't already
+- `Generate an authentication token <https://developer.aiven.io/docs/platform/howto/create_authentication_token.html>`_ on Aiven's console or CLI
+
+
```
lornajane, line 12: This could be a section with a heading "Configure your project and services"

docs/tools/terraform/getting-started.rst
```diff
@@ -0,0 +1,98 @@
...
+This is where you put the actual values for Aiven API token and Aiven console project name. This file is passed to Terraform using the ``-var-file=`` flag.
+
+4. ``services.tf`` file:
+
+.. code:: bash
+
+   # Your actual Terraform script goes here 
+
+Create an empty folder and add the above files to that folder. Then execute the following commands in order:
```
lornajane, line 67: The instruction on what to do with the files should probably be included before the files themselves, so the user knows what to do as they meet each one.

This section with the commands in it could have a heading "Build your platform with Terraform" or something like that? I think starting by explaining that the services are declared, and now we'll use Terraform to make reality match the yaml. A user should know why they are running the command before they do it, and have an expectation on what the command will do.

docs/tools/terraform/getting-started.rst
```diff
@@ -0,0 +1,98 @@
...
+
+.. code:: bash
+
+   terraform init 
+
+This command performs several different initialization steps in order to prepare the current working directory for use with Terraform. In our case, this command automatically finds, downloads, and installs the necessary Aiven Terraform Provider plugins.
```
lornajane, line 73: I found the instructions, but they should go before the command, to explain the next step before doing (edit: this suggestion applies to the rest of the article as well)

docs/tools/terraform/getting-started.rst
```diff
@@ -0,0 +1,98 @@
...
+
+This command creates an execution plan and shows you the resources that will be created (or modified) for you. This command does not actually create any resource; this is more like a preview.
+
+.. code:: bash
+
+   terraform apply -var-file=var-values.tfvars
+
+If you're satisfied with ``terraform plan``, you execute ``terraform apply`` command which actually does the task or creating (or modifying) your infrastructure resources. 
+
+Optional
```
lornajane, line 87: I think we could use a more descriptive heading ... how about "Remove services when you are finished" or something?

docs/tools/terraform/getting-started.rst
```diff
@@ -0,0 +1,98 @@
...
+--------
+
+If this was a test environment, be sure to delete the resources once you're done to avoid consuming unwanted bills. 
+
+.. warning::
+
+   Use this command with caution. This will actually delete resources that might have important data.
```
lornajane, line 94: Can we be clearer about how to be more confident about the command? I think the warning is a good one but let's show how to plan the destroy, and use this warning to prompt the user to check that none of the resources listed for destruction are things they wanted to keep.

lornajane (changes requested):
I know this was an adaptation of existing content but I still added the feedback that I think would make this the most useful for our users (and I think we know that this is going to be a pretty high profile resource on the site!), it's all minor and I think the structure changes are good!

docs/tools/terraform/getting-started.rst
dewan-ahmed: I used "Clean up" as the title. Please suggest if you like "Remove services when you are finished" instead.

dewan-ahmed:
Changes via latest commits:

1. Brought instructions block before the code block so that the audience understands what the code blocks will do.
2. Edited section titles based on feedback.
3. code/literal formatted couple of variables.
4. Clarified admin level access for API auth token.
5. Added further instructions so that the audience can be confident with ``terraform destroy``.
6. Further reference for those who'd like to setup a more complex Terraform project.

Still a bit unclear:
1. [What explanation to add on the Terraform project itself?](https://github.com/aiven/devportal/pull/646#discussion_r827904415)
2. If the modified text for admin level access for auth token will suffice
3. If it's appropriate to use **BYOTF - Bring Your Own Terraform Script**

@lornajane thanks for your last review and looking forward to this round of review.

docs/tools/terraform/get-started.rst
```diff
@@ -0,0 +1,126 @@
...
+
+.. code:: bash
+
+   # Your actual Terraform script goes here 
+
+
+2. Consider the following code block similar to declaring a dependency; the *Aiven Terraform Provider* in this case. Within the `required_providers` block, you mention the source of the provider and specify a certain version. 
+The versions here are for example purpose and will change in the future. Following `Aiven Terraform Provider doc <https://registry.terraform.io/providers/aiven/aiven/latest/docs>`_, `api_token` is the only parameter for the provider configuration.
+
+Make sure that you have either the *Administrator* or *Operator* role when creating the API token. When you create a project, you automatically receive the *Administrator* access.
```
lornajane, line 31: This is much better! Can it go with the instructions for generating an API token (you can just link to what we have) right before storing it in the file? I don't think this is context the user needs before they can create the providers file.

docs/tools/terraform/get-started.rst
```diff
@@ -0,0 +1,126 @@
...
+
+``var-values.tfvars`` file:
+
+.. code:: bash
+
+   aiven_api_token = "<YOUR-AIVEN-AUTHENTICATION-TOKEN-GOES-HERE>"
+   project_name = "<YOUR-AIVEN-CONSOLE-PROJECT-NAME-GOES-HERE>"
+
+
+Execution
```
lornajane, line 79: Ideally a verb first, then a subject. So perhaps "Apply the Terraform configuration" ?

docs/tools/terraform/get-started.rst
```diff
@@ -0,0 +1,126 @@
...
+.. warning::
+
+   Use the following command with caution. This will actually delete resources that might have important data.
+
+.. code:: bash
+
+   terraform destroy -var-file=var-values.tfvars
+
+
+Further reference
```
lornajane, line 123: We could also reference the cookbook either here or as a note near the `services.tf` example.

lornajane (changes requested):
Nice! I think for your point 3, we do need to pick an example, just one service even, so that people can use this guide to actually get started. Other than that, some minor comments but we're in good shape.

Let me know if you'd like a hand to rebase the branch.

docs/tools/terraform/get-started.rst
dewan-ahmed: Added a reference to the cookbook near the `services.tf` section.

docs/tools/terraform/get-started.rst
dewan-ahmed: I moved this part to right before storing the token in the file. Regarding adding a link to creating an auth token, we have that link under the "Requirements" section. Is this what you asked? I might have misinterpreted.

dewan-ahmed:
Changes via https://github.com/aiven/devportal/pull/646/commits/66ee4ca0d0c00d6e7c8efe0cdb6eefddfb6872ec:

1. Fixes based on @lornajane's last review.
2. Adding a newline to the docs/platform/concepts/tls-ssl-certificates.rst file which was causing a warning during build

I marked other comments as resolved but left one open which I'm not quite sure about.

Lorna has scheduled a 30mins call for March 22, 2022 to go over rebasing this branch with main (and resolving the conflicts).

dewan-ahmed:
Thanks @lornajane for the help with rebase. I added a few more changes to fix the indentation issue and section title (based on your previous comments).

Currently, on the landing page, we have a shortcut to 'Reference' and 'Cookbook'. Is that Ok? Since we only have 'Cookbook' under the 'Reference' section, I wonder if having both is redundant. Another way to think is we might have non-cookbook content under the 'Reference' section in the future so it might make sense to keep both.

Looking forward to your thoughts on this and any other change that this PR might need.

lornajane:
I pushed some small changes, @dewan-ahmed take a quick look that I haven't done anything awful and then we'll merge if you're happy?

dewan-ahmed:
Thanks @lornajane 

The guide reads much better after your edits.

I have one small question about the phrase "show off" but other than that, I'm very happy. Thanks again.

lornajane (approved):
Super happy with the improved getting started guide and landing page. Thank you!

PR #647, opened 2022-03-16
update bigquery sink link to confluent as wepay is deprecated

encima:
Replace wepay bigquery link with confluent as it is deprecated and we use the confluent one.


lornajane (approved):
Small but useful change, thanks!

PR #649, opened 2022-03-16
Enabled words in the URL to be searched

kevinbowman-aiven:

Have added the "URL" field to the OpenSearch index definition, so any searches can also use that.

### DIFF
--- a/netlify/functions/search/search.js
+++ b/netlify/functions/search/search.js
@@ -57,6 +57,14 @@ const handler = async (event) => {
                   },
                 },
               },
+              {
+                match_phrase_prefix: {
+                  url: {
+                    query: query,
+                    slop: 5,
+                  },
+                },
+              },
             ],
           },
         },


kevinbowman-aiven:
(changing to draft, this PR currently only changes the indexing, but there's also likely a change needed to the Netlify function as well)

kevinbowman-aiven:
I'm pretty sure this will work now, I've tried to recreate the query which the Netlify function sends as best I can.

lornajane:
To test:
 - check the instructions in `SEARCH.md` for testing with a new opensearch service 
 - check that the indexes include the URL 
 - check that the search function works (including terms in the URL) when run locally with `netlify dev`

To deploy:
 - deploy, then after the build, drop the existing index on the live opensearch service
 - then recreate the index and re-index both devportal and help centre using local makefile commands and the production OpenSearch creds
 - check that search works on live

laysauchoa:
the code seems right, did you had the chance to test with the steps mentioned by Lorna, @kevinbowman-aiven?

lornajane:
@laysauchoa one of the reviewers needs to do the testing, I think we can assume that the PR author thinks it works! I commented as a note to either you or me for what we need to look at.

laysauchoa:
alright, I tried to run the tests this morning and I was not able to do the test in my local machine. I may be missing something in how to test it following the instructions. Related to https://github.com/aiven/devportal/issues/681

laysauchoa:
Thank you so much for the PR. I believe is a great feature to have for sure. Here is what I understood about the PR: If the word is included in the URL and not in the content of the file, this still would enable the page to be returned as a valid. Please, correct me if I understood it wrong.

**For example**
the page: https://developer.aiven.io/docs/products/kafka/kafka-connect/concepts/list-of-connector-plugins.html
In the content of this page, there is no mention of the word `plugins`
But the word is included in the URL. **Therefore, the results should include this page.** 

So looking for the word `plugins`. I'd expect the result to have this:

```
{ ...,
"url": "docs/products/kafka/kafka-connect/concepts/list-of-connector-plugins.html",
    "source": "devportal",
    "sort_priority": 1
  },
  "highlight": {
    "title": [
      "@opensearch-dashboards-highlighted-field@List@/opensearch-dashboards-highlighted-field@ @opensearch-dashboards-highlighted-field@of@/opensearch-dashboards-highlighted-field@ available Apache Kafka® Connect connectors - Aiven"
    ],
...}
```

So this is the case that I am testing. I followed the testing steps and looked for the word: `plugins`. 
**Branch:** `search-includes-url`
**Query:** http://localhost:49429/.netlify/functions/search?query=plugins
**Result:** The file result does not contain this URL. You can find my results here: [query=plugins.txt](https://github.com/aiven/devportal/files/8370299/query.plugins.txt)

Did I get something wrong about the final goal of the PR? Maybe you can help me to understand which cases are you testing, or does this case work for you?

laysauchoa:
To sum up

To test:

check the instructions in SEARCH.md for testing with a new opensearch service _(done - works)_
check that the indexes include the URL _(includes the correct url)_
check that the search function works (including terms in the URL) when run locally with netlify dev _(partially-works; URL are included but I found a case where it does not return the URL although word is included there)_

_**I've not deployed because I would like to clarify why my results are different from what I expected.**_

kevinbowman-aiven:
Hi @laysauchoa, yes you're correct, that page should have been in your results.  I've just checked it on a local setup here, and as long as I've built the index using the code from this branch, then when I switch between this branch and the main branch then I do see that page appear/disappear from the results.

A couple of things to check:
* Were you on this branch when you initially created the index, and when spidering the content, and when querying it?  I think if any of those happened when you were on the main branch then it will affect the results
* I can see from the results file you linked that there are 2 pages of results (10 per page), but you've only posted the first page.  If you go to `http://localhost:49429/.netlify/functions/search?query=plugins&page=2` do you see the missing result on that page?

laysauchoa:
@kevinbowman-aiven it works, thank you for the tip. Indeed, it was in second page in my case.

laysauchoa (approved):
🥳 It works so it can be merged!

laysauchoa:
@lornajane it may make sense to have this one too reviewed and merged together: https://github.com/aiven/devportal/pull/687

lornajane:
Thank you very much @kevinbowman-aiven for this improvement and @laysauchoa for taking the time to review!

PR #650, opened 2022-03-16
Use synonyms when searching

kevinbowman-aiven:

This PR adds the capability to specify a list of synonyms to be used when searching.  It adds them at index-creation time, there are a couple of them seeded in the PR so it's obvious where to add more.

Note that this PR clashes with #649 so I've built it on top of that and will leave as draft until that is accepted/rejected (and if the latter, I'll rebuild it on main).

### DIFF
--- a/scripts/synonyms.json
+++ b/scripts/synonyms.json
@@ -1,4 +1,4 @@
 [
      "postgresql, postgres, pg",
-     "kafka, kafak, kfaka"
+     "mirrormaker, mm2"
 ]
--- a/scripts/synonyms.json
+++ b/scripts/synonyms.json
@@ -1,0 +1,4 @@
+[
+     "postgresql, postgres, pg",
+     "kafka, kafak, kfaka"
+]
--- a/SEARCH.md
+++ b/SEARCH.md
@@ -125,6 +125,19 @@ os_client.index(index='devportal',
 
 You might also need to take care of removing documents that no longer exist.
 
+# Synonyms
+
+If you want to alias one search term to another, then you can update the file `scripts/synonyms.json` with comma-separated lists of aliases.  For example:
+```
+[
+     "postgresql, postgres, pg",
+     "kafka, kafak, kfaka"
+]
+```
+Note the lack of a trailing comma on the final term; this file must be valid JSON.
+
+The aliases are used at index-creation time, so if this file changes then the index needs to be re-created before it will take effect.
+
 # Testing changes to search functionality
 
 It seems like Netlify uses the deployed functions rather than the ones in a branch when building a preview, so we need to take care when testing these. A good approach is to use the [Netlify CLI](https://www.netlify.com/products/cli/) locally. For example, to test the search function:
--- a/netlify/functions/search/search.js
+++ b/netlify/functions/search/search.js
@@ -23,6 +23,7 @@ const handler = async (event) => {
       }
     }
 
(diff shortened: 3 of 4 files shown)


anelook (changes requested):
This is a great improvement!

The code itself looks good to me. However, I think we shouldn't put all synonyms imbedded directly in `scripts/create_index.py`. Let's move them to a separate file. Can be a JSON file with analyzer body.
In this way we can keep a clear separation between the code to create an index and the devportal analyzer.

kevinbowman-aiven:
Good point!  I've broken the synonyms out into a different file, as opposed to the whole analyzer, as I think that's a more approachable way for people to add new synonyms (I think there's more likely to be synonym changes in the future than wider analyzer changes).  However, if you think the whole analyzer should be broken out instead, then I'm happy to do that instead, just let me know.

Also possible future enhancement: GitHub action to lint-check the JSON file.

scripts/synonyms.json
```diff
@@ -0,0 +1,4 @@
+[
+     "postgresql, postgres, pg",
+     "kafka, kafak, kfaka"
```
anelook, line 3: should "kafka, kafak, kfaka" better achieved with fuzzy matching?

kevinbowman-aiven: Yes, it absolutely should be :). I wrote this before the fuzzy matching PR went in, which makes this more useful for fundamental aliases rather than typos (like "pg").  I'll remove the kafka one, although I think it's good to seed the list with at least 2 entries so I'll find another one to add in its place.

anelook:
I see your point and agree that keeping just synonyms in a separate file makes it easier for people to add synonyms. 

One question about file location - should we put it next to other vocabulary files we have? 
Also, extra bonus if we add couple of lines into the README how to add the synonyms! :)  

One more question - can I test this somehow locally? Or how did you test this change?

kevinbowman-aiven:
> One question about file location - should we put it next to other vocabulary files we have?

Do you mean in `.github/styles/Vocab/` ?  It feels a little strange to put this into a `.github` directory, but I really don't have a strong feeling about it.  If you'd prefer it to be there, then I'm happy to move it?

> Also, extra bonus if we add couple of lines into the README how to add the synonyms! :)

I've added it to the SEARCH doc - does that get me the bonus points? :)

> One more question - can I test this somehow locally? Or how did you test this change?

Yes, you should be able to follow the steps at the bottom of the SEARCH.md doc ("Testing changes to search functionality") - let me know internally if any of that doesn't work and I'll debug it with you.

scripts/synonyms.json
kevinbowman-aiven: I've changed this for a different synonym example ("mirrormaker, mm2").

TibsAtWork:
On the `.github` directory issue: `vale` stuff (the `.github/styles` directory) is there because that's where the github action for vale says we should put it. In the future, there will be multiple subdirectories (not just `styles`) in  the more specific `.github/vale`.

I'd be tempted to keep the JSON file next to the script, at least for the moment - we can think about project organisation later on as we gain more tooling.

anelook (approved):
Thank you @kevinbowman-aiven! This will be very helpful for our users!

Tested locally - all looks great!
Merging.

lornajane:
Did this PR need the indexes recreating on OpenSearch to take effect? Or does it do its magic when the search runs?

kevinbowman-aiven:
Yes, the analyzer is setup at index-creation time.

PR #674, opened 2022-03-23
Remove netlify-cli dependency to reduce vulnerabilities

lornajane:
Another attempt to improve our dependency situation after reports from Dependabot. This is an alternative to #653, which causes the build to fail and looks like it still leaves some vulnerable dependencies. The netlify-cli package should only be needed for local testing of changes to the netlify functions, and the documentation shows that it should be installed as a global dependency. It isn't needed for most contributors, or for production, so I suggest removing it.


TibsAtWork (approved):
Looks sensible to me

PR #675, opened 2022-03-23
Generate list of clouds from API, include in CLI docs

lornajane:

I often wish I had a reference list of clouds to look at. This adds exactly that functionality, but we'll need to update it from time to time by re-running the script. I would like feedback on this PR but is is **incomplete** until I add `make` commands to run the script and update the include file (following the same pattern as the advanced parameters docs do) so please don't merge.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -83,3 +83,6 @@ service-type-config-grafana:
 service-type-config-influxdb:
 	python "$(SOURCEDIR)/scripts/aiven/service_type_config.py" "influxdb" > includes/config-influxdb.rst
 
+# (Re)Generate cloud listing
+cloud-list:
+	python "$(SOURCEDIR)/scripts/aiven/clouds.py" > includes/clouds-list.rst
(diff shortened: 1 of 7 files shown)


TibsAtWork:
I definitely like the idea of being able to find this from the documentation.

I think I'd like a note saying on what date this list was generated.

I *suspect* we'll get asked to allow sorting by either column, but I would ignore that for now.

lornajane:
Feel free to pick it up if you don't want to wait for me :) I put it here because this is where I looked for it. I am not sure if GitStamp (the thing that does the timestamps) picks up the last-updated date of the include content.

We have geography information, but weirdly not provider information, in the API response, so that's an option. (No, we are not about to do interactive filtering)

ftisiot:
I moved the article to `/platform/reference` and ordered the results by Cloud. @TibsAtWork I believe we can still improve this , maybe splitting the tables by geo-region? Also, still missing the timestamp

ftisiot:
Added the Timestamp and managed to grab the correct Cloud name, also added the entry in the Makefile. Should be ready to review @TibsAtWork and @lornajane.
I'm not sure how to setup the automation tho

lornajane:
I love it (and the automation will have to follow, we have a bunch of these that I just re-run from time to time!!). However I can't approve my own PR so I'm tagging @TibsAtWork for review

TibsAtWork (approved):
This looks excellent to me. Let's go for it.

PR #680, opened 2022-03-25
helparticle: Disable user 2FA setting [HELP-95]

simonna-k:

adding instruction for users how to disable user 2FA setting

if the user contacted us anonymously, we are not able to verify who they are in order to help them with their request. However, in this case we do have a self-service feature where customers can disable 2FA themselves: if they request to reset their password in the login page, 2FA will be disabled for them.


lornajane (approved):
Thank you @simonna-k . I also learned something :)

PR #682, opened 2022-03-28
Datasource integration between two PG services

kevinbowman-aiven:

Added a short doc explaining what the "PostgreSQL" option does under the "datasource" integration type when selected on a PG service.

I'm not wild about the title of the doc, feel free to change it if you can think of something better.


ftisiot (approved):
LGTM

Issue #686, opened 2022-03-29
Enable misspelling words to return relevant results


laysauchoa:
Using a word which is misspelled can happen to bring no result to the query.

For example: looking for the word Apache Kafka with misspelling brings no result

<img width="1470" alt="Screenshot 2022-03-29 at 20 41 14" src="https://user-images.githubusercontent.com/36624597/160682529-c8b6ef8d-3305-48cd-b73c-376a48162254.png">

The solution is to allow some fuzziness in the queries. It is recommended to use the fuzziness parameter in a dynamic way rather than a fixed one.

PR #687, opened 2022-03-29
Enable misspelling words to return relevant results

laysauchoa:
Fixes: #686 

Test:
Screenshot 2022-03-29 at 20 43 45


lornajane (approved):
I love that we're taking good advantage of using our own technologies! This small improvement is excellent, thank you!

PR #688, opened 2022-03-30
Update non-leader-for-partition.rst

davidespo:

Adding clarity about inactive producers. They will not wait until errors to update cluster metadata cache.


ftisiot (approved):
LGTM

Issue #692, opened 2022-04-04
The generated sitemap.xml puts `/en/` into the URLs, so they don't work


TibsAtWork:
We use the sphinx extension `sphinx_sitemap` to generate a sitemap.xml (it gets built in `_build/html/sitemap.xml`).

By default, that includes the `language` setting in the URL, producing things like:
```
<url><loc>https://developer.aiven.io/en/docs/platform/howto/create_authentication_token.html</loc></url>
```

Unfortunately that URL doesn't exist - what we want is:
```
<url><loc>https://developer.aiven.io/docs/platform/howto/create_authentication_token.html</loc></url>
```
without the `/en`.

Looking at the extension documentation (at https://pypi.org/project/sphinx-sitemap/), we need to set `sitemap_url_scheme` explicitly to avoid this.

PR #693, opened 2022-04-04
sitemap.xml links should not include `/en`

TibsAtWork:
Tell the sitemap extension not to put the language (`/en`) into the URLs in the `sitemap.xml` file

**Before** looks like https://developer.aiven.io/sitemap.xml, with instance, entries like:
```
<url>
<loc>https://developer.aiven.io/en/docs/community.html</loc>
</url>
```
(there's no such page)

**After** loses the `/en` and looks like https://deploy-preview-693--developer-aiven-io-preview.netlify.app/sitemap.xml, with entries like:
```
<url>
<loc>https://developer.aiven.io/docs/community.html</loc>
</url>
```
which is a page that exists.

Fix https://github.com/aiven/devportal/issues/692

### DIFF
--- a/conf.py
+++ b/conf.py
@@ -61,6 +61,12 @@ gitstamp_fmt = "%B %Y"
 
 # sitemap config
 html_baseurl = 'https://developer.aiven.io'
+# Since we have `language='en'` set (further down) the URLs in the sitemap will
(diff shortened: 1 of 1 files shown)


anelook (approved):
Tested locally - seeing same results. The issue with the urls is now fixed.

PR #699, opened 2022-04-05
Fix ClickHouse urls

anelook:

ClickHouse updated their interface and reshuffled some of the resources. They also removed mentioning of Yandex metrica and replaced with generic naming


TibsAtWork (approved):
Thanks. Links seem to work.

PR #702, opened 2022-04-05
Change database fork to service fork to provide better relevant search and navigation experience

runwuf:
Replace `database` with `service` for fork instructions.  `Database` does not apply to grafana.  This is to provide better relevant search result and navigation in developer portal when user search on how to fork Aiven services, not limited to `database` to cause confushion.


docs/platform/concepts.rst
```diff
@@ -16,7 +16,7 @@ Learn about some of the key concepts for working with Aiven platform:
   Learn about Aiven's access control, encryption, network security, data privacy and operator access.
 
 
-* :doc:`Database forking </docs/platform/concepts/database-forking>`.
+* :doc:`Service forking </docs/platform/concepts/service-forking>`.
 
   Quickly and safely make a copy of your database for testing, dry run upgrades, or anything else, without affecting your existing service.
```
TibsAtWork, line 21:
```suggestion
  Quickly and safely make a copy of your database or other service for testing, dry run upgrades, or anything else, without affecting your existing service.
```
Perhaps clarify here that it is not just a database (but it's nice, I think, to leave the word database in that text).

TibsAtWork (approved):
These seem sensible changes to me.

The renamed file is in `platform/concepts` which is wildcarded in `_toc.yml`, so no change to that is needed to update the navigation sidebar.

Issue #357, opened 2021-11-23
The `disable_replication_factor_adjustment` feature for OpenSearch is deprecated


ryansammonaiven:
# What's wrong?

The `disable_replication_factor_adjustment` feature for OpenSearch was deprecated in order to ensure a consistent high availability configuration across services. There is no way to disable automatic replication now.

# URL of affected page, and any other information

https://developer.aiven.io/docs/products/opensearch/howto/switch_off_automatic_replication_factor_adjustment

PR #703, opened 2022-04-06
Migrate article about opensearch replication factors

anelook:

Migration of https://help.aiven.io/en/articles/2342117-automatic-adjustment-of-elasticsearch-index-replication-factors

Note, advanced property disable_replication_factor_adjustment is deprecated and in dev-sandbox / dev-advocates no longer available. That's why I excluded the part where we recommend setting this value.

### DIFF
--- a/docs/products/opensearch/reference/index-replication.rst
+++ b/docs/products/opensearch/reference/index-replication.rst
@@ -1,7 +1,7 @@
 Automatic adjustment of replication factors
 ===========================================
 
-Aiven for OpenSearch automatically adjusts index replication factors
+Aiven for OpenSearch® automatically adjusts index replication factors
 to ensure data availability and service functionality.
 
 When is the replication factor automatically adjusted?
--- a/docs/products/opensearch/reference/index-replication.rst
+++ b/docs/products/opensearch/reference/index-replication.rst
@@ -4,33 +4,13 @@ Automatic adjustment of replication factors
 Aiven for OpenSearch automatically adjusts index replication factors
 to ensure data availability and service functionality.
 
-When replication factor is automatically adjusted?
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+When is the replication factor automatically adjusted?
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
--  If ``number_of_replicas`` is too large for current cluster size, it
-   is automatically lowered to maximum possible value (number of nodes
-   on the cluster - 1).
+The maximum value for ``number_of_replicas`` is the number of nodes in the cluster - 1, as it is not possible to replicate indexes (shards) to a larger number of nodes than exist on the cluster
 
--  If ``number_of_replicas`` is 0 on a multi-node cluster, it is
-   automatically increased to 1.
+For example, for a 3-node cluster, the maximum ``number_of_replicas`` is 2, which means all shards on the index are replicated to all 3 nodes.
 
--  If ``number_of_replicas`` is anywhere between 1 and maximum value, it
-   is not automatically adjusted.
-
-Lowering replication factor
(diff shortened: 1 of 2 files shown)


lornajane:
See also: #357

docs/products/opensearch/reference/index-replication.rst
```diff
@@ -0,0 +1,36 @@
+Automatic adjustment of replication factors
+===========================================
+
+Aiven for OpenSearch automatically adjusts index replication factors
+to ensure data availability and service functionality.
+
+When replication factor is automatically adjusted?
```
TibsAtWork, line 7: It was like this in the original, but if I was writing this from scratch, I'd put the "`is`" after the "`When`" - that is, "`When is ...`"

docs/products/opensearch/reference/index-replication.rst
```diff
@@ -0,0 +1,36 @@
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+-  If ``number_of_replicas`` is too large for current cluster size, it
+   is automatically lowered to maximum possible value (number of nodes
+   on the cluster - 1).
+
+-  If ``number_of_replicas`` is 0 on a multi-node cluster, it is
+   automatically increased to 1.
+
+-  If ``number_of_replicas`` is anywhere between 1 and maximum value, it
```
TibsAtWork, line 17: Minor point, but since we're editing stuff, "`the maximum`"

docs/products/opensearch/reference/index-replication.rst
```diff
@@ -0,0 +1,36 @@
...
+
+Aiven for OpenSearch automatically adjusts index replication factors
+to ensure data availability and service functionality.
+
+When replication factor is automatically adjusted?
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+-  If ``number_of_replicas`` is too large for current cluster size, it
+   is automatically lowered to maximum possible value (number of nodes
+   on the cluster - 1).
```
TibsAtWork, line 12: I would rephrase as something like:
```
The maximum value for ``number_of_replicas`` is the number of nodes in the cluster - 1. 
```
and then carry on with the example from below:
```
For example, for a 3-node cluster, the maximum ``number_of_replicas`` is 2, which means all shards on the index are replicated to all 3 nodes.
```
and finish with
```
If ``number_of_replicas`` is too large, it will automatically be reduced to the maximum value.
```

docs/products/opensearch/reference/index-replication.rst
```diff
@@ -0,0 +1,36 @@
...
+Lowering replication factor
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+When replication factor ( ``number_of_replicas`` ) is set to larger than
+size of the cluster, ``number_of_replicas`` is automatically lowered, as
+it is not possible to replicate indexes (shards) to a larger number of
+nodes than what exists on the cluster. Do note number replication factor
+is ``number_of_replicas`` + 1. For example, for three-node cluster,
+maximum ``number_of_replicas`` is 2, which means all shards on the index
+are replicated to all three nodes.
```
TibsAtWork, line 29: If the example is moved earlier, then I think this is all said by the previous section, so can be removed.

docs/products/opensearch/reference/index-replication.rst
```diff
@@ -0,0 +1,36 @@
...
+
+Increasing replication factor
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+For multi-node clusters, Aiven for OpenSearch automatically increases
+``number_of_replicas`` to 1, if it is set to 0. This is to ensure no
+data-loss occurs if one node is lost.
```
TibsAtWork, line 36: Again, this was already said in the list in the earlier section.

TibsAtWork (commented):
I *think* the original text repeated itself a bit much, so it could be shortened.

Unless you feel there's value in saying the max / min points in different ways, so people don't miss them.

PR #704, opened 2022-04-06
Add more ® and other fixes

TibsAtWork:
* Add missing ® and ™*, add some "for" (Aiven XXX -> Aiven for XXX)
* Correct "Hashicorp" in `accept.txt` to be "HashiCorp", and also fix in
  docs
* Correct a `pg` that should be `jq`
* Fix a `:doc:` link (remove trailing underscore)
* Add acknowledgement of Elasticsearch in a file where it is implicitly
  referenced

(These are some fixes I found while working on https://github.com/aiven/devportal/pull/510)

**NOTE** this appears to be failing on `make linkcheck` because of the "moving links" at clickhouse.com. It passes `make clean; make html` and `make spell`.


ftisiot (approved):
LGTM

Issue #705, opened 2022-04-06
Remove reference to PostgreSQL `file_fdw` extension


TibsAtWork:
The PostgreSQL extension `file_fdw` provides arbitrary reads from PostgreSQL, which has bad security implications.
It is also not clear that a user can actually make use of it in a practical sense (since a user cannot access files on the server itself).

Remove mention of it from the documentation at https://developer.aiven.io/docs/products/postgresql/reference/list-of-extensions

It is assumed that it will also be removed from the supported extensions (because otherwise the user will be able to rediscover it for themselves, and also because at some point in the future we might want to automate scanning for plugins - see https://github.com/aiven/devportal/issues/444)

PR #706, opened 2022-04-06
Remove reference to PostgreSQL extension `file_fdw`

TibsAtWork:
Remove reference to PostgreSQL extension `file_fdw` in the list of PostgreSQL extensions

It's questionable whether a user can actually make use of this extension, and using it has security implications. Remove the description so users don't expect it to be supported.

Fix: https://github.com/aiven/devportal/issues/705


ftisiot (approved):
LGTM

PR #710, opened 2022-04-06
Change database fork to service fork to provide better relevant search and navigation experience

runwuf:
Replace `database` with `service` for fork instructions. Database does not apply to grafana. This is to provide better relevant search result and navigation in developer portal when user search on how to fork Aiven services, not limited to database to cause confusion.


ftisiot:
The problem i see is that when compiling the website we get an error

`../devportal/docs/platform/concepts/database-forking.rst: WARNING: document isn't included in any toctree`

TibsAtWork:
@ftisiot That does seem like a nice way to handle the redirection of the database-forking file to service-forking.
Does it need a comment explaining what it's doing, or is it sufficiently obvious to someone who understands a little bit of javascript and DOM?

TibsAtWork (approved):
This all seems to work, including the redirection.

Issue #660, opened 2022-03-21
Service Migration Check


laysauchoa:
https://help.aiven.io/en/articles/5622913-service-migration-check

PR #715, opened 2022-04-07
mysql/howto: service migration check

laysauchoa:
https://help.aiven.io/en/articles/5622913-service-migration-check

Fixes: https://github.com/aiven/devportal/issues/660

### DIFF
--- a/docs/products/mysql/howto/do-check-service-migration.rst
+++ b/docs/products/mysql/howto/do-check-service-migration.rst
@@ -1,0 +1,75 @@
+Service migration check
+=======================
+
+In this article, we show how to find potential errors before starting your database migration process. 
+This can be done by using either the `Aiven CLI <https://github.com/aiven/aiven-client>`_  or `Aiven REST API <https://api.aiven.io/doc/#section/Introduction>`_. 
+
+When migrating a database to Aiven, you may find errors such as:
+
+.. code-block:: shell
+
+    {
+        "migration": {
+            "error": "Migration process failed",
+            "method": "",
+            "seconds_behind_master": null,
+            "source_active": true,
+            "status": "done"
+        },
+        "migration_detail": []
+    }
+    -----Response End-----
+    STATUS METHOD ERROR
+    ====== ====== ========================
+    done Migration process failed
+
+
+
+To avoid these types of failures during the migration process we recommend you run some checks in advance. In this article, we will show how you can run those checks for your MySQL migration process.
+
+Aiven CLI
+#########
+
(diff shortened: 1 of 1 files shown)


TibsAtWork:
It's a faithful migration of the original article, but I have a couple of questions:

1. (Minor complaint) It's not at all obvious to me what is meant to have gone wrong in the migration process. This is true of the original text. Is the original example "simplified" to leave information out of the error response? It looks like it's just saying *something* went wrong.

2. (More complicated) The original article makes it clear that this might apply to any database migration, not just MySQL, and that it is using MySQL as an example. So do we need to have this article for each database, or should we keep it in a general place (not sure where) and link to it from each database's migration page.

laysauchoa:
> It's a faithful migration of the original article, but I have a couple of questions:
> 
> 1. (Minor complaint) It's not at all obvious to me what is meant to have gone wrong in the migration process. This is true of the original text. Is the original example "simplified" to leave information out of the error response? It looks like it's just saying _something_ went wrong.
> 2. (More complicated) The original article makes it clear that this might apply to any database migration, not just MySQL, and that it is using MySQL as an example. So do we need to have this article for each database, or should we keep it in a general place (not sure where) and link to it from each database's migration page.

1) I have performed the migration checks but I do not know how to trigger migration errors. I will check if this is a general example of an error or a specific one.
2) Some of the parts of this article are only applied to MySQL such as the use of the filter, so imo this does not fit as a general article. I asked @Aiqin-Aiven about it and seems like postgress also supports migration check. Maybe we can have an issue about it to be investigated whether should be expanded to others services or not according to the priority, wdyt?

TibsAtWork:
Thanks!
For (1) I wouldn't worry too much - the idea of "there might be an error" is the important thing,

For (2) that sounds good - I think I agree that an issue for PG would be the way to go.

Let me know if you do want to do anything about (1), otherwise I shall approve it as it is.

laysauchoa:
Thanks @TibsAtWork for the comment. 

(1) I think so too, but I am in checking with more people and maybe I'll add extra info later if something pops up :)
(2) I have added an issue for postgresql but probably we will be focusing on the migration first. In any case, it is good to keep it in mind in case this content is requested by more people. fyi, issue: https://github.com/aiven/devportal/issues/724

TibsAtWork:
@laysauchoa  Let me know when you're done checking for (1) / happy for me to approve, and I'll get back to this.

laysauchoa:
All good we can merge this one! @TibsAtWork

PR #716, opened 2022-04-07
Change ClickHouse urls to follow their docs

anelook:
ClickHouse reverted changes to the documentation (where there was a change in file structure). So I'm changing the urls as well.


laysauchoa (approved):
lgtm!

PR #717, opened 2022-04-08
Add language to literalinclude

mdedetrich:
Something I noticed when working on devport is that some of the `literalinclude` are missing the `language` sections.

### DIFF
--- a/docs/products/postgresql/howto/connect-node.rst
+++ b/docs/products/postgresql/howto/connect-node.rst
@@ -30,7 +30,8 @@ Code
 
 Add the following to ``index.js`` and replace the placeholder with the PostgreSQL URI:
 
-.. literalinclude:: /code/products/postgresql/connect.js
+.. literalinclude:: /code/products/postgresql/connect.js\
+   :language: javascript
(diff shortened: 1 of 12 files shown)


docs/products/postgresql/howto/connect-node.rst
```diff
@@ -30,7 +30,8 @@ Code
 
 Add the following to ``index.js`` and replace the placeholder with the PostgreSQL URI:
 
-.. literalinclude:: /code/products/postgresql/connect.js
+.. literalinclude:: /code/products/postgresql/connect.js\
```
TibsAtWork, line 33:
```suggestion
.. literalinclude:: /code/products/postgresql/connect.js
```
(extra `\` at the end of the line, stops it showing the code)

TibsAtWork (commented):
Thanks for this - it does often seem to make a good guess as to the language, but it's definitely much better to be explicit about it.

I've also added this to my list of "things it would be nice to check for"

TibsAtWork (approved):
Excellent.

Issue #665, opened 2022-03-21
What is the maximum number of connections for MySQL?


laysauchoa:
https://help.aiven.io/en/articles/3288602-what-is-the-maximum-number-of-connections-for-mysql

PR #718, opened 2022-04-08
MySQL/add: how to estimate max number of connections

laysauchoa:
From the original article,
some estimations have changed, so it is updated here:

https://help.aiven.io/en/articles/3288602-what-is-the-maximum-number-of-connections-for-mysql

Fixes: https://github.com/aiven/devportal/issues/665

### DIFF
--- a/docs/products/mysql/howto/max-number-of-connections.rst
+++ b/docs/products/mysql/howto/max-number-of-connections.rst
@@ -1,0 +1,29 @@
+Calculate the maximum number of connections for MySQL
+=====================================================
+
+
+The number of simultaneous connections in Aiven for MySQL depends on the usable memory on the server. The usable memory is the total memory on the node minus the operating system and management overhead. This overhead is currently estimated as 350 MiB (0.341797 GiB). 
+
+It can be estimated as:
+
+.. math::
+
+    {usable\_memory} = {total\_memory\_on\_the\_node - management\_overhead} 
+
+.. note::
+    
+
+To calculate the maximum number of connections, we need to check the value of the usable memory. The usable memory is rounded to the nearest GiB. For plans under 4 GiB of usable memory, you can estimate as:
+
+.. math::
+    
+   {max\_number\_of\_connections} = usable\_memory\_in\_GiB\times 75 + {extra\_connection}
+
+
+For plans higher or equal to 4 GiB of usable memory, you can calculate as:
+
+.. math::
+    
+   {max\_number\_of\_connections} = usable\_memory\_in\_GiB\times 100 + {extra\_connection}
+
+Independent of the plan, an ``extra_connection`` with a value of ``1`` will be added for the system process.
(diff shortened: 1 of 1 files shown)


docs/products/mysql/howto/max-number-of-connections.rst
```diff
@@ -0,0 +1,29 @@
+Calculate the maximum number of connections for MySQL
+=====================================================
+
+
+The number of simultaneous connections in Aiven for MySQL depends on the usable memory on the server. The usable memory is the total memory on the node minus the operating system and management overhead. This overhead is currently estimated as 350 MiB (0.341797 GiB).
```
TibsAtWork, line 5: Is it useful to go to six digits after the decimal point?
```
>>> 350.0 / 1024
0.341796875
```
so that's already rounding a bit. Not sure how much precision is needed, but is `approximately 0.34` good enough?

docs/products/mysql/howto/max-number-of-connections.rst
```diff
@@ -0,0 +1,29 @@
...
+It can be estimated as:
+
+.. math::
+
+    {usable\_memory} = {total\_memory\_on\_the\_node - management\_overhead} 
+
+.. note::
+    
+
+To calculate the maximum number of connections, we need to check the value of the usable memory. The usable memory is rounded to the nearest GiB. For plans under 4 GiB of usable memory, you can estimate as:
```
TibsAtWork, line 16: I think this should explicitly say `rounded down` (which is what the *help* article said)

TibsAtWork (commented):
I like the new style of the article.

But I would still quite like a couple of worked examples at the end - one for the "my RAM is 4GiB" and the other for a higher case.

Is there anything useful (to the user) that can be said about *why* things are different above/below 4GiB?

docs/products/mysql/howto/max-number-of-connections.rst
laysauchoa: I think an approximation is enough. In the end, it will be rounded to the nearest integer. I also agree that some examples would be good, so I have added. It is not clear why the changes were done for those two plans but the explanation is that plans with +4Gib have less memory-constrained compared to the other one.

TibsAtWork: 👍

docs/products/mysql/howto/max-number-of-connections.rst
```diff
@@ -0,0 +1,52 @@
...
+.. math::
+
+    {usable\_memory} = {total\_memory\_on\_the\_node - management\_overhead} 
+
+To calculate the maximum number of connections, we need to check the value of the usable memory. For this estimation, the usable memory value is **rounded down** to the nearest integer value of GiB. 
+
+**For plans under 4 GiB of usable memory**, you can estimate as
+
+.. note::
+    Independent of the plan, an ``extra_connection`` with a value of ``1`` will be added for the system process.
```
TibsAtWork, line 17: Can we move this note above the `**For plans` line? That would make it feel more general.

Otherwise, this seems perfect to me.

docs/products/mysql/howto/max-number-of-connections.rst
TibsAtWork: The examples make this nice and clear - thank you.

docs/products/mysql/howto/max-number-of-connections.rst
laysauchoa: yes, we can. I also think it makes more sense tbh. Thank you!

TibsAtWork (approved):
Thank you. And many compliments on your maths layout as well!

Issue #652, opened 2022-03-18
Improve the titles in the `_toc.yml` file (navigation sidebar)


TibsAtWork:
## The problem

The [`_toc.yml`](https://github.com/aiven/devportal/blob/main/_toc.yml) file is used to generate the navigation sidebar.

Some of the page titles in the `_toc.yml` file are specified explicitly in the file, some default to using the (top level) heading of the page itself.

Some of those default titles are quite long, and they may contain words that are not needed in the context of the sidebar (for instance, it may be clear what the product is from the tree structure, but the page heading will want to be explicit).

Also, adding ® marks to many top level titles adds visual clutter to the sidebar. We would rather keep the sidebar simple to read.

## The solution

It is *probably* sensible to just specify page titles explicitly in the `_toc.yml` file for all pages.

It is definitely needed for longer titles, or titles that contain ® or other marks.

Also, add a section on "creating a new page" to the [CONTRIBUTING](https://github.com/aiven/devportal/blob/main/CONTRIBUTING.rst) guide, and explain what needs doing to keep the `_toc.yml` file up to date.

TibsAtWork:
See also https://github.com/aiven/devportal/issues/580 which is partially related

Issue #694, opened 2022-04-04
SEO: Some titles are too long and cut off in Google search


anelook:
# What's wrong?

SEO analysis indicated a numerous pages where the title is cut in google search results.

<img width="647" alt="Screenshot 2022-04-04 at 13 41 27" src="https://user-images.githubusercontent.com/4600541/161536706-0223dcc2-3766-4edf-8ff0-14d8b96fcf21.png">

The titles themselves are not the culprit. The length of the title is increased because we add "Aiven Developer documentation" at the end.

Should we change this and remove "Aiven Developer documentation"?

Alternatively, we might want to keep "Aiven Developer documentation". This gives the documentation some identity and recognisability. Additionally, it makes people aware that there is such thing as "Aiven Developer documentation" when they scroll the results, even if from time to time the results are cut.  

# URL of affected page, and any other information
Some examples:
https://developer.aiven.io/docs/products/flink/concepts/kafka-connectors
https://developer.aiven.io/docs/products/flink/howto/connect-kafka
https://developer.aiven.io/docs/products/flink/howto/connect-opensearch

lornajane:
Check the live site, this was raised recently and I changed the trailing text to just " - Aiven". Some of our titles will be too long, because the topics are specific and complex, and the full product names are long. But for that example, my guess is that it hasn't been reindexed yet since I changed the trailing text

lornajane:
This doesn't seem to be a showstopper currently, so closing.

PR #720, opened 2022-04-16
Tidy up titles, both on pages and in navigation

### LINKED ISSUE #580: Improve the heading text when searching for articles
Follow up from Helena Meißner findings:

Improve the heading text when searching for articles.
E.g. adding the service name instead of "Aiven Developer documentation"

MySQL for example but as not using that pattern consistently

Search related

Issue
screenshot_2022-02-16_at_14 10 32

Working example:
screenshot_2022-02-16_at_14 13 28

lornajane:
Fixes #580. Also improves #652 although there is probably more to do on that one.

No page URLs were harmed in the making of this PR. The aim was to have:
 - Descriptive page titles if they are viewed in isolation, e.g. in search results, or when landing directly on a page from a referral link
 - Consistent (but short and without trademark decoration) titles in all the product sections for the navigation bar

It's a lot of changes, hopefully it's not too much to review at once, I couldn't think of a logical way to split this.

### DIFF
--- a/_toc.yml
+++ b/_toc.yml
@@ -90,7 +90,9 @@ entries:
     title: Apache Kafka
     entries:
       - file: docs/products/kafka/getting-started
+        title: Get started
       - file: docs/products/kafka/howto/fake-sample-data
+        title: Sample data generator
       - file: docs/products/kafka/concepts
         entries:
           - file: docs/products/kafka/concepts/acl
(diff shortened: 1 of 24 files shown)


anelook (approved):
Looked through all changes and they look good to me.
The changed titles are more descriptive now. Note, this will affect the length in (google)search results and part of the title can be cut (https://github.com/aiven/devportal/issues/694). we should be mindful about long titles.

PR #723, opened 2022-04-19
Migrate MySQL memory usage

laysauchoa:
Add Concepts article on MySQL part.


docs/products/mysql/concepts/mysql-memory-usage.rst
```diff
@@ -0,0 +1,16 @@
+MySQL memory usage
+==================
+
+MySQL InnoDB buffer pools are **pre-allocated** and shown as **used memory** even without user activity or stored data in the database. Because of this pre-allocated memory, it's expected high memory usage within all Aiven MySQL plans.
```
ftisiot, line 4:
```suggestion
`MySQL InnoDB buffer pools <https://dev.mysql.com/doc/refman/8.0/en/innodb-buffer-pool.html>`_ are **pre-allocated** and shown as **used memory** even without user activity or stored data in the database. Because of this pre-allocated memory, you can expect a high memory usage within all Aiven MySQL plans. 
```

docs/products/mysql/concepts/mysql-memory-usage.rst
```diff
@@ -0,0 +1,16 @@
...
+
+MySQL InnoDB buffer pools are **pre-allocated** and shown as **used memory** even without user activity or stored data in the database. Because of this pre-allocated memory, it's expected high memory usage within all Aiven MySQL plans. 
+
+.. important::
+    
+    The high memory usage on MySQL Aiven services does not indicate any particular issue. 
+
+InnoDB buffer pools are set to values between 15% to 80% of total available memory. Smaller plans use a smaller percentage of total memory for buffer pools because the operating system and management infrastructure components and other MySQL features may consume a considerable portion of the available memory. A sufficient amount of free memory is reserved to run extra utilities while backing up the database.
+
+The MySQL documentation references buffer pool sizing explicitly:
```
ftisiot, line 12: Do we have a reference for this? I would put also the link to the documentation stating the fact

docs/products/mysql/concepts/mysql-memory-usage.rst
```diff
@@ -0,0 +1,16 @@
+MySQL memory usage
```
ftisiot, line 1: should this be 
```suggestion
MySQL high memory usage
```

docs/products/mysql/concepts/mysql-memory-usage.rst
```diff
@@ -0,0 +1,16 @@
...
+    
+    The high memory usage on MySQL Aiven services does not indicate any particular issue. 
+
+InnoDB buffer pools are set to values between 15% to 80% of total available memory. Smaller plans use a smaller percentage of total memory for buffer pools because the operating system and management infrastructure components and other MySQL features may consume a considerable portion of the available memory. A sufficient amount of free memory is reserved to run extra utilities while backing up the database.
+
+The MySQL documentation references buffer pool sizing explicitly:
+
+    A buffer pool is the memory area that holds cached InnoDB data for both tables and indexes. A larger buffer pool requires less disk I/O to access the same table data more than once. On a dedicated database server, you might set the buffer pool size to 80% of the machine's physical memory size.
+
+All nodes also have some base CPU consumption due to management overhead (about 5% of a single CPU core).
```
ftisiot, line 16: This is written as last comment, but it's a relevant information that maybe we can include a bit more towards the end

ftisiot (changes requested):
Hi! Useful article to have in devportal. 

I have a doubt regarding the format since as of now it is:
* InnoDB buffer pools pre-allocated 
* We allocate 15%-80%
* Reference from mySQL docs
* Ah by the way we add extra 5% in management overhead

I would probably give the title "Understand MySQL high memory usage" and structure it as follows:

* Intro about seeing high usage in Mysql and the statement that is normal (people don't have to worry)
* Start taking down memory usage in pieces
1) Management overhead 5%
2) InnoDB buffer pools 15%-80%
* Explanation why this happens
* Reference to MySQL docs
* Additional reference to Aiven keeping some free memory for backups

laysauchoa:
> This article is a bit more general. The reason for that is too much details could expose business logic.

Some of the suggestions can be applied like links and change of titles. However, I have discussed internally and more explanations should not be added in the external documentation.

See my comment: **This article is a bit more general. The reason for that is too much details could expose business logic.**

ftisiot:
Yep, I agree with that, my suggestion is more towards reshaping the existing content to fit a logical information flow. We don't need to add any new information

docs/products/mysql/concepts/mysql-memory-usage.rst
laysauchoa: As it is the same link and they are very close maybe I will add only once

laysauchoa:
> Yep, I agree with that, my suggestion is more towards reshaping the existing content to fit a logical information flow. We don't need to add any new information

I agree. I change the structure a bit, so hope it is more clear. Feel free to add any minor thing if it is needed it.

ftisiot (approved):
LGTM

PR #741, opened 2022-04-20
Add vale easy install command

NotoriousPyro:
Added a quick command to get the latest Vale from tar.gz and then untar it to $HOME/bin

This is because I don't want to use brew or snap to install it, and their doco only seems to have a simple example that uses the current dir instead of tmp dirs and $HOME/bin: https://vale.sh/docs/vale-cli/installation/#github-releases

I'm not sure if this is more suitable elsewhere.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -50,7 +50,10 @@ If the spellchecker is rejecting words that are valid (such as technology terms)
 
 Installing Vale
 """""""""""""""
-On Fedora (F34), you can install Vale easily (without a package manager) using the following:::
+
+The `Vale installation page <https://docs.errata.ai/vale/install>`_ has instructions for all platforms including docker; this will be updated if the approach changes between versions.
+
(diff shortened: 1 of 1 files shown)


NotoriousPyro:
Might be more suitable as a target in https://github.com/aiven/devportal/blob/main/Makefile

Thoughts?

NotoriousPyro:
Might be more suitable as a target in https://github.com/aiven/devportal/blob/main/Makefile

Thoughts?

lornajane:
I am a little nervous of encouraging people to run commands and pipe things to curl and so on. So let's definitely not hide the details away; people should be aware of what they are running. We can also link to the project installation instructions so that people always refer to the current process recommended by the project.

NotoriousPyro:
Agreed. Will also add `-v` to `cp` to make it verbose so that people know what's being copied from where.

lornajane (approved):
Thanks very much!

PR #743, opened 2022-04-20
Test file for tutorial with Tibs

ewe-gra:
It's weird because my structured text didn't work.


laysauchoa:
@ewe-gra there is a small typo about the extension of your file. It should be `.rst` and it is written as `.rts`. If you change that and push again, it will work :)

ewe-gra:
This was a testing session. Closing this PR.

PR #754, opened 2022-04-21
Migrate REST API endpoint access information

anavasiliuk:
Migrated https://help.aiven.io/en/articles/2026734-elasticsearch-rest-api-limited-access-endpoints.


lornajane:
Do we know _why_ only some endpoints are available? I think a bit more explanation could be useful here if we can find out a bit more background.

lornajane (approved):
This is in good shape, thanks!

PR #755, opened 2022-04-21
Fix extra symbol added in commands

laysauchoa:
According to the README of devportal:

Do not include a `$` before a command that the user should run,
because it will get copied into the user's clipboard and cause
the command to fail (this has been a common standard in the past).

For this reason, I'd like to remove from all the commands the symbol.


lornajane (approved):
A quality of life improvement indeed. Thanks @laysauchoa :)

PR #757, opened 2022-04-21
Add some structure on MySQL code samples

laysauchoa:
Follow what is being done in other repo, I'd like to have the sample codes in one place specific. Also, I want to re-write articles who do not follow this structure. I am starting with the connection from the command line here.


lornajane (approved):
This is nice! A couple of alterations so minor that it was quicker to make them myself :)

PR #763, opened 2022-05-04
Update filename reference to fix broken link for Redis.

dewan-ahmed:

Based on a previous PR, the Redis getting started guide filename changed which caused two broken links for https://developer.aiven.io/docs/products/redis/howto/migrate-aiven-redis.html page and https://developer.aiven.io/docs/products/redis/index.html page. This PR fixes those broken references.


laysauchoa (approved):
thanks for fixing it!

Issue #765, opened 2022-05-05
Bump to Python 3.7 or latest


laysauchoa:
Python 3.6 stopped getting security updates in December 2021, we should try to migrate to Python 3.7 or the latest Python.

PR #766, opened 2022-05-06
Move to Python 3.7

laysauchoa:
Python 3.6 stopped getting security updates in December 2021,
this commit migrates to Python 3.7 our workflows.

fixes: https://github.com/aiven/devportal/issues/765


TibsAtWork (approved):
Thank you for doing this.
Assuming it's all safe because it didn't complain (about itself)!

Issue #677, opened 2022-03-24
Document: How to enable Aiven password


laysauchoa:
Trying to login with this without having a Aiven password does not work

```
avn user login <user@aiven.io>
```

This is the error message:

```
❯ avn user login [user-email]()
[user-emal]()'s Aiven password:
ERROR	command failed: Error: {"errors":[{"message":"Authentication failed","status":403}],"message":"Authentication failed"}
```
I believe this documentation is missing in dev portal. I was referring to this [blog](https://aiven.io/blog/command-line-magic-with-the-aiven-cli).

Here are some steps that needs to be taken to make it work:

- Enable Aiven password

<img width="1207" alt="Screenshot 2022-03-24 at 13 42 25" src="https://user-images.githubusercontent.com/36624597/159920868-11639a6a-7fbe-4093-bdb4-5da4fadf047b.png">

- Check if password works properly in the login

<img width="1207" alt="Screenshot 2022-03-24 at 13 44 15" src="https://user-images.githubusercontent.com/36624597/159920886-72057b81-3490-4e6e-a374-3f956b6fa6f1.png">

As suggestion from @ftisiot this article should be in: **How to create a Aiven Password** and linked to **How to login with Aiven CLI**.

laysauchoa:
Motivation had changed a bit: This is needed if the user wants to authenticate and log in directly without using any 3rd parties by using Aiven username and password.

Still blog should be fixed to follow the token access instead of Aiven password access.

TibsAtWork:
The CLI blog post is due to be replaced by a new one.

PR #768, opened 2022-05-06
How to enable Aiven password

laysauchoa:
This is needed if the user wants to authenticate and log in directly without using any 3rd parties by using Aiven username and password.

fixes:https://github.com/aiven/devportal/issues/677

### DIFF
--- a/_toc.yml
+++ b/_toc.yml
@@ -24,6 +24,7 @@ entries:
       - file: docs/platform/howto/create_new_service
       - file: docs/platform/howto/create_new_service_user
       - file: docs/platform/howto/create_authentication_token
+      - file: docs/platform/howto/enable-aiven-password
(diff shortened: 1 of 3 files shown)


lornajane:
I would recommend that we instead advise users to use a token to log in with CLI

laysauchoa:
Thanks, I'll clarify the use of this feature first.

TibsAtWork:
Yes, I was thinking that was something important to talk about. And I'd already written "Discuss pros and cons of using password vs token." in the outline to remind me.

laysauchoa:
I've added a note on the cli tools article about it and focused on the case that the user wants to authenticate and log in directly without using any 3rd parties by using Aiven username and password. Thanks again!

TibsAtWork (approved):
Thank you for sorting this out.

PR #769, opened 2022-05-06
amend opensearch dashboard alerting

ebeta-aiven:
The article has been amended to reflect the fact that it is currently not possible to authenticate a sender account when creating an email destination in Opensearch alerts. That is, we do not offer a way to set username and password in Opensearch keystore. To do so one needs to run the command ./bin/opensearch-keystore in all the nodes and we do not expose a way to achieve the same result.

I have mentioned this aspect among the important notes.


runwuf (approved):

lornajane:
Thanks @ebeta-aiven , thanks @runwuf !

PR #774, opened 2022-05-09
Minor updates spotted while using the CLI docs

lornajane:

A tiny bit of reformatting and some command updates, I was using the docs for something and picked up a few things so here they are! Also bonus reformatting of an email address.


ftisiot (approved):
LGTM

Issue #292, opened 2021-10-25
PostgreSQL and Google Data Studio


lornajane:
Migrate existing content: https://help.aiven.io/en/articles/1661836-integrating-google-data-studio-with-aiven-for-postgresql

PR #778, opened 2022-05-10
Migrate PostgreSQL and Google Data Studios page

TibsAtWork:
Old page: https://help.aiven.io/en/articles/1661836-integrating-google-data-studio-with-aiven-for-postgresql

* Removed diagrams (two were now wrong anyway)
* Updated directions to current steps
* Checked it all worked

Fixes: https://github.com/aiven/devportal/issues/292

### DIFF
--- a/.github/vale/styles/Aiven/capitalization_headings.yml
+++ b/.github/vale/styles/Aiven/capitalization_headings.yml
@@ -23,6 +23,7 @@ exceptions:
   - Go
   - Google BigQuery
   - Google Cloud Storage
+  - Google Data Studio
   - Grafana
   - Hobbyist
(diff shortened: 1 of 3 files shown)


ftisiot:
Hi @TibsAtWork, good work! I just checked and the  client authentication is not a needed step and adds a bit of unnecessary complexity IMHO, should we take it out?

TibsAtWork:
Confirmed (and thanks again) - it only needs the `ca.pem` file (but it does need that, presumably because we have `SSLmode require` on our PG service).  Should be ready for re-review now, @ftisiot

ftisiot (approved):
Great work!

PR #779, opened 2022-05-11
Update Cassandra icon

lornajane:
We were using an outdated image for the cassandra icon, so use the new one instead!


laysauchoa (approved):
looks good to me! Thanks @lornajane for changing it

Issue #782, opened 2022-05-11
Migrate Python examples for connecting to Apache Kafka


TibsAtWork:
Migrate article about connecting to and using Kafka with Python

https://help.aiven.io/en/articles/5343895-python-examples-for-testing-aiven-for-apache-kafka

- [x] I plan to open a pull request for this

PR #836, opened 2022-05-16
Migrate help for Kafka Python examples

TibsAtWork:
Migrate the "how to connect" part of
https://help.aiven.io/en/articles/5343895-python-examples-for-testing-aiven-for-apache-kafka

I have removed the "example" code, as it's very hard to make good general example code, and I think it's better for the user to look elsewhere. This follows what we've done for the "connect with XXX" pages for other products. In the process, I've made it closer in style to (for instance) the MySQL page.

I've merged the paired SASL examples, as I don't think we need to provide examples that only differ in the `sasl_mechanism`.

Fixes: https://github.com/aiven/devportal/issues/782

### DIFF
--- a/docs/products/kafka/howto/connect-with-python.rst
+++ b/docs/products/kafka/howto/connect-with-python.rst
@@ -91,7 +91,7 @@ Variable              Description
                         partition
 =================     =============================================================
 
-For more information on ``auto_offset_reset``, see the Kafka documentation on [auto.offset.rest](https://kafka.apache.org/documentation/#consumerconfigs_auto.offset.reset) and [Consumer Position](https://kafka.apache.org/documentation/#consumerconfigs_auto.offset.reset)
+For more information on ``auto_offset_reset``, see the Kafka documentation on `auto.offset.rest <https://kafka.apache.org/documentation/#consumerconfigs_auto.offset.reset>`_ and `Consumer Position <https://kafka.apache.org/documentation/#consumerconfigs_auto.offset.reset>`_.
 
 
 Connect a producer
--- a/docs/products/kafka/howto/connect-with-python.rst
+++ b/docs/products/kafka/howto/connect-with-python.rst
@@ -1,8 +1,10 @@
 Connect to Aiven for Apache Kafka® with Python
 ==============================================
 
-These examples show how to connect to an Aiven for Apache Kafka® service using the
-`kafka-python <https://pypi.org/project/kafka-python/>`__ library, as either a producer or consumer.
+.. |kafka-python| replace:: ``kafka-python``
+.. _`kafka-python`: https://pypi.org/project/kafka-python/
+
+These examples show how to connect to an Aiven for Apache Kafka® service using the |kafka-python|_ library, as either a producer or consumer.
 
 .. note:: The examples given here provide different options for the different authentication
    methods. For more information on the supported methods, see `our article on Kafka
@@ -11,7 +13,7 @@ These examples show how to connect to an Aiven for Apache Kafka® service using 
 Pre-requisites
 --------------
 
-Install the Python  `kafka-python <https://pypi.org/project/kafka-python/>`__ library:
+Install the Python |kafka-python|_ library:
 
 .. code:: bash
 
--- a/docs/products/kafka/howto/connect-with-python.rst
+++ b/docs/products/kafka/howto/connect-with-python.rst
@@ -95,7 +95,10 @@ Variable              Description
                         partition
 =================     =============================================================
 
-For more information on ``auto_offset_reset``, see the Kafka documentation on `auto.offset.reset <https://kafka.apache.org/documentation/#consumerconfigs_auto.offset.reset>`_ and `Consumer Position <https://kafka.apache.org/documentation/#consumerconfigs_auto.offset.reset>`_.
+For more information on ``auto_offset_reset``, see the Kafka documentation on
+`auto.offset.reset <https://kafka.apache.org/documentation/#consumerconfigs_auto.offset.reset>`_
+and
+`Consumer Position <https://kafka.apache.org/documentation/#design_consumerposition>`_.
 
 
 Connect a producer
--- a/docs/products/kafka/howto/connect-with-python.rst
+++ b/docs/products/kafka/howto/connect-with-python.rst
@@ -91,7 +91,7 @@ Using SASL authentication
 
          from kafka import KafkaProducer
 
-         # Choose an appropriate SASL mechanism
+         # Choose an appropriate SASL mechanism, for instance:
          SASL_MECHANISM = 'SCRAM-SHA-256'
 
          producer = KafkaProducer(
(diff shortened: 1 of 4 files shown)


ftisiot:
@TibsAtWork could you please check why Vale is complaining about this one?

TibsAtWork:
@ftisiot Fixed with reStructuredText magic.

docs/products/kafka/howto/connect-with-python.rst
```diff
@@ -0,0 +1,154 @@
...
+
+Using SSL authentication
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. code:: python
+
+        from kafka import KafkaConsumer
+
+        consumer = KafkaConsumer(
+            "demo-topic",
```
ftisiot, line 122: Should we say that this should be the topic you want to read from?

docs/products/kafka/howto/connect-with-python.rst
```diff
@@ -0,0 +1,154 @@
...
+            auto_offset_reset="earliest",
```
ftisiot, line 123: Also this is an important parameter, can we describe it a bit more?

docs/products/kafka/howto/connect-with-python.rst
```diff
@@ -0,0 +1,154 @@
...
+            bootstrap_servers=f"{HOST}:{SSL_PORT}",
+            client_id = CONSUMER_CLIENT_ID,
+            group_id = CONSUMER_GROUP_ID,
+            security_protocol="SSL",
+            ssl_cafile="ca.pem",
```
ftisiot, line 128: Should we state that this should point to the file? Including the folder if needed?

ftisiot (changes requested):
Hi @TibsAtWork ! This is solid work! 
I would perhaps expand the examples a bit more in order to make them "standalone" by:
* discussing additional parameters like the topic name and the `auto_offset_reset`
* showing  the code to produce/consume (as of now we're only defining how to connect)?

Happy to discuss more if each of the points doesn't make sense

TibsAtWork:
Thanks @ftisiot for the useful comments - I shall follow them up.

I'd taken out the actual produce/consume code because there are a variety of different ways to do it, and the typical ways one shows in examples are not (I think) like those one would use in production. Also, other (sort of similar) articles for other products do just limit themselves to connection.

On the other hand, it is true that the KafkaConsumer documentation doesn't make it obvious it can be used as an iterator!

I could add a "simple demo usage" section to the end if you think it would be worthwhile.

docs/products/kafka/howto/connect-with-python.rst
TibsAtWork: Thanks for spotting that - this used to be introduced in text that got removed.
I'm adding it to the `Variables` section further up.

docs/products/kafka/howto/connect-with-python.rst
TibsAtWork: I'm adding this to the `Variables` as well, with links into the Kafka documentation.

I don't know if it's also worth linking to the medium article https://medium.com/lydtech-consulting/kafka-consumer-auto-offset-reset-d3962bad2665 which seems to have a good description?

docs/products/kafka/howto/connect-with-python.rst
TibsAtWork: I've added a note to the end of the `Pre-requisites` section - let me know if you think that's enough.

TibsAtWork:
I'm waiting for a fix on `main` to rebase this to fix the Sphinx error report, but in terms of actual text I think this is ready for another look, @ftisiot. I've got comments on your earlier comments somewhere above this...

docs/products/kafka/howto/connect-with-python.rst
```diff
@@ -0,0 +1,177 @@
...
+
+* If you are going to connect with SSL authentication:
+
+  * In the *Connection information* section:
+
+    #. If **Authentication Method** is shown, choose **Client Certificate**
+    #. Next to *Access Key*, click **Download** and save the ``service.key`` file.
+    #. Next to *Access Certificate*, click **Download** and save the ``service.cert`` file.
+    #. Next to *CA Certificate*, click **Download** and save the ``ca.pem`` file.
+
```
ftisiot, line 35: Can we add a tip about obtaining the same with the cli https://developer.aiven.io/docs/tools/cli/service/user.html#avn-service-user-creds-download ?

docs/products/kafka/howto/connect-with-python.rst
```diff
@@ -0,0 +1,177 @@
...
+                      Allowed values are:
+
+                      * ``latest`` - consume from the end of the topic partition.
+                        This is the default.
+                      * ``earliest`` - consume from the beginning of the topic
+                        partition
+=================     =============================================================
+
+For more information on ``auto_offset_reset``, see the Kafka documentation on `auto.offset.rest <https://kafka.apache.org/documentation/#consumerconfigs_auto.offset.reset>`_ and `Consumer Position <https://kafka.apache.org/documentation/#consumerconfigs_auto.offset.reset>`_.
```
ftisiot, line 94: Small bug :D 

```suggestion
For more information on ``auto_offset_reset``, see the Kafka documentation on `auto.offset.reset <https://kafka.apache.org/documentation/#consumerconfigs_auto.offset.reset>`_ and `Consumer Position <https://kafka.apache.org/documentation/#consumerconfigs_auto.offset.reset>`_.
```

ftisiot (changes requested):
Almost good to go!

There's a small bug and a tip regarding avn cli I would add, Rest is perfect!

docs/products/kafka/howto/connect-with-python.rst
TibsAtWork: Oh yes, good point. Thanks.

docs/products/kafka/howto/connect-with-python.rst
TibsAtWork: Ooh, well spotted!

TibsAtWork:
I don't understand the two linkcheck failures (in **Sphinx / prose (pull_request)**.
It seem to think that https://kafka.apache.org/documentation/#consumerconfigs_auto.offset.reset and for https://kafka.apache.org/documentation/#design_consumerposition are broken,  but both work if I try them (including clicking on the relevant text in the preview). I *assume* is is some problem with the `linkcheck` program that sphinx uses.

It *might* be related to https://github.com/sphinx-doc/sphinx/issues/9255, although I'm not convinced.

TibsAtWork:
I'll investigate a little more, but we might have to live with it, or perhaps mark those as links to ignore in the `conf.py`

TibsAtWork:
I've adopted the "ignore those links" approach, and raised https://github.com/aiven/devportal/issues/894 so we don't lose this problem.

ftisiot (approved):
LGTM

PR #854, opened 2022-05-18
add DR scenario explanation and process

encima:
This could go under a new `support` folder but added to `concepts` here. Add Disaster Recovery explanation, how to request it and briefly how it works (maybe more technical details should be added?)

### DIFF
--- a/.github/vale/dicts/aiven.dic
+++ b/.github/vale/dicts/aiven.dic
@@ -26,7 +26,6 @@ datastore
 dashboards
 Debezium
 Dev
-Disaster Recovery
 Distro
 Docker
 dockerized
--- a/.github/vale/dicts/aiven.dic
+++ b/.github/vale/dicts/aiven.dic
@@ -26,6 +26,7 @@ datastore
 dashboards
 Debezium
 Dev
+Disaster Recovery
 Distro
 Docker
 dockerized
--- a/.github/vale/styles/Aiven/capitalization_headings.yml
+++ b/.github/vale/styles/Aiven/capitalization_headings.yml
@@ -16,6 +16,7 @@ exceptions:
   - Connect
   - Dashboards
   - Dev
+  - Disaster Recovery
   - Docker
   - Elasticsearch
   - Flink
--- a/_toc.yml
+++ b/_toc.yml
@@ -20,6 +20,7 @@ entries:
       - file: docs/platform/concepts/static-ips
       - file: docs/platform/concepts/tls-ssl-certificates
       - file: docs/platform/concepts/byoa
+      - file: docs/platform/concepts/disaster-recovery-test-scenarios
     - file: docs/platform/howto
       title: HowTo
       entries:
--- a/docs/platform/concepts.rst
+++ b/docs/platform/concepts.rst
@@ -34,3 +34,7 @@ Learn about some of the key concepts for working with Aiven platform:
 * :doc:`Bring your own account (BYOA) </docs/platform/concepts/byoa>`.
 
   BYOA is an optional setup feature that allows you to purchase your Aiven services through your existing Cloud Provider.
+
+* :doc:`Disaster Recovery (DR) testing scenarios </docs/platform/concepts/disaster-recovery-test-scenarios>`.
+  
+  DR testing scenarios are simulations of disaster scenarios run for enterprise support customers by Aiven operators.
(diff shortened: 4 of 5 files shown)


encima:
The Aiven for PostgreSQL linting but I assume we should not add the ® as it implies the Aiven naming is trademarked with PG.
The `aiven` recommendation is within an email but unsure how to ignore this one 🤦‍♀️

TibsAtWork:
I do like this article, and it seems a reasonable place to put it to me.

I've made a couple of fixes:

* It isn't necessary to add `Disaster Recovery` to the dictionary, as both `disaster` and `recovery` are in the default dictionary (and the capitalisation is ignored). I'm not actually sure how a "word" with a space in it would be used, either.
* The underlining for a reStructuredText title always needs to be at least as long as the title text. I've fixed that for the top title.
* `Aiven for PosgreSQL®` is the correct form - the `®` is taken to belong to PostgreSQL, and the use of "for" is enough to indicate we're not claiming any ownership of the PostgreSQL name.
* The email address problem is down to a weirdness of how the linting is done, and the only workaround we've got (which I've used) is to specify `support@Aiven.io` as the email "text". This looks odd to me, but is a perfectly valid form of our email address (and will work).

In order for the document to appear in the right place in the navigation, it also needs adding to the `_toc.yml` file at the top-level of the source tree. If you want me to help with that, please do ask

encima:
> I do like this article, and it seems a reasonable place to put it to me.
> 
> I've made some fixes:
> 
> * It isn't necessary to add `Disaster Recovery` to the dictionary, as both `disaster` and `recovery` are in the default dictionary (and the capitalisation is ignored). I'm not actually sure how a "word" with a space in it would be used, either.
> * The underlining for a reStructuredText title always needs to be at least as long as the title text. I've fixed that for the top title.
Thank you!
> * `Aiven for PosgreSQL®` is the correct form - the `®` is taken to belong to PostgreSQL, and the use of "for" is enough to indicate we're not claiming any ownership of the PostgreSQL name.

Hmm, that seems to me like Aiven holds the trademark but I have no idea whether that is true!

> * The email address problem is down to a weirdness of how the linting is done, and the only workaround we've got (which I've used) is to specify `support@Aiven.io` as the email "text". This looks odd to me, but is a perfectly valid form of our email address (and will work).

Couldn't an exception be added to the linter for `@aiven` instead of `aiven`? 

> In order for the document to appear in the right place in the navigation, it also needs adding to the `_toc.yml` file at the top-level of the source tree. If you want me to help with that, please do ask

Thanks! I have done that now. Cheers for the quick turnaround!

TibsAtWork:
The "Aiven for XXX<mark>" approach is the official way to do it - contact me outside github if you want more (too much!) detail 😄 

The email address linting is a good point. I had thought it was a problem with no solution (given our toolchain), but a bit more investigation suggests it might be fixable with some work on our linting rules. Unfortunately that's not going to happen now, but it is now on my backlog to investigate further.

TibsAtWork:
Apologies - nearly there, but I'd forgotten that `docs/platform/concepts.rst` is the one case (I think) where the summary page actually has content that is written by hand. You can see this at https://deploy-preview-854--developer-aiven-io-preview.netlify.app/docs/platform/concepts.html - the sidebar has your new page, but it's missing from the main page text.

Could I get you to add a summary at the end, similar to the others already there. and then I'll approve?

(I think the new page gives just the right amount of information, by the way - more technical details are always tempting, but then they may go out-of-date, and also are probably best explained in a proper dialogue with the customer once they've contacted support.)

encima:
> Apologies - nearly there, but I'd forgotten that `docs/platform/concepts.rst` is the one case (I think) where the summary page actually has content that is written by hand. You can see this at https://deploy-preview-854--developer-aiven-io-preview.netlify.app/docs/platform/concepts.html - the sidebar has your new page, but it's missing from the main page text.
> 
> Could I get you to add a summary at the end, similar to the others already there. and then I'll approve?
> 
> (I think the new page gives just the right amount of information, by the way - more technical details are always tempting, but then they may go out-of-date, and also are probably best explained in a proper dialogue with the customer once they've contacted support.)

Thanks for the feedback! That is a good point about leaving out specifics. If more are requested then we can always add later.

Added now and hopefully it makes sense/passes Vale!

TibsAtWork:
It passes `make html` (apart from one known issue, not yours) and `make spell` on my machine, as well as passing CI.

PR #856, opened 2022-05-18
Fixed whitespace to remove non-printable chars

### LINKED ISSUE #855: Prometheus config example contains non-printable characters
# What's wrong?

The `scrape_configs` examples for Prometheus do not copy correctly when using the button in the top-right. Rogue characters end up in the `prometheus.yml` configuration and then the file cannot be parsed (example: `C2A0`, UTF-8 non-breaking space)

https://developer.aiven.io/docs/platform/howto/integrations/prometheus-metrics.html

screenshot example

dbeech:
Fixes #855


dbeech:
I have tested the copy button for both code examples from the deploy preview site and both now work as expected

lornajane (approved):
Small fix but mighty!

lornajane:
Thank you for both the bug and the fix :)

Issue #841, opened 2022-05-16
Platform overview still lists Elasticsearch as an Aiven service


TibsAtWork:
# What's wrong?

If you go to https://developer.aiven.io/docs/platform/index.html you can see that `Elasticsearch` is still listed 
as an Aiven product.

Since it is no longer possible to create a new Elasticsearch service using the Aiven console, this should probably be removed from the list on this page.

PR #862, opened 2022-05-19
Removed the Elasticsearch reference.

divdby0:

Removed the Elasticsearch service references from the Aiven platform definition page since it is no longer available as a service.

Fixes: https://github.com/aiven/devportal/issues/841


TibsAtWork:
I've updated the Conversation/Description to say "Fixes:" and the link to the issue, which should mean that when we approve and merge the PR, it also closes the issue as fixed.

TibsAtWork (approved):
Thanks you very much - a good first issue to fix!

PR #866, opened 2022-05-20
Improve venv instructions

juha-aiven:
Simplify the venv create command.
Clarify the .gitignore-related note.
Add the exact command to activate the venv.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -15,11 +15,13 @@ Local Development
 
 We recommend using a virtual environment like `venv <https://docs.python.org/3/library/venv.html>`_::
 
-    python3 -m venv /path/to/new/virtual/environment
+    python3 -m venv venv
 
-If you call it something other than `venv` or `env`, make sure to add your preferred name to `<.gitignore>`_.
+If you call the directory (the last parameter above) something other than `venv` or `env`, make sure to add the directory name to `<.gitignore>`_.
 
-Activate your virtual environment using the `activate` script for your environment.
+Activate your virtual environment using the `activate` script for your environment::
+
+    source venv/bin/activate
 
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -15,11 +15,13 @@ Local Development
 
 We recommend using a virtual environment like `venv <https://docs.python.org/3/library/venv.html>`_::
 
-    python3 -m venv /path/to/new/virtual/environment
+    python3 -m venv venv
 
-If you call it something other than `venv` or `env`, make sure to add your preferred name to `<.gitignore>`_.
+If you call the directory (the last parameter above) something other than `venv` or `env`, make sure to add the directory name to `<.gitignore>`_.
```
TibsAtWork, line 20: This is definitely better than the previous version, but we'd rather not encourage people to add their own `venv` alternative names to the project `.gitignore` file. I wonder if we should just drop this line - especially now you've made the "create a venv" command more specific (which is good).

juha-aiven: I think we might as well drop the line about `.gitignore`. The way the instructions describe setting up `venv` works, but that's naturally the only way to set things up. If the user wants to be clever and do things otherwise, he'll need to figure out things himself.

juha-aiven: Pushed a commit that removed the mention re `.gitignore`.

juha-aiven:
Any ideas why the checks are failing?

TibsAtWork:
Not sure - nothing obvious to me. Trying to run them again.

TibsAtWork:
I can't seem to get it to re-run those.
It is probably simplest if you close and then re-open the PR, to force the checks again.

PR #867, opened 2022-05-20
Karapace security

juha-aiven:
Add documentation on Karapace Schema Registry and Karapace REST Proxy security.

Note this documentation is for an unreleased feature.

### DIFF
--- a/_toc.yml
+++ b/_toc.yml
@@ -120,6 +120,7 @@ entries:
       - file: docs/products/kafka/concepts
         entries:
           - file: docs/products/kafka/concepts/acl
+          - file: docs/products/kafka/concepts/karapace-security
           - file: docs/products/kafka/concepts/log-compaction
           - file: docs/products/kafka/concepts/partition-segments
           - file: docs/products/kafka/concepts/horizontal-vertical-scaling
--- a/docs/products/kafka/concepts/karapace-security.rst
+++ b/docs/products/kafka/concepts/karapace-security.rst
@@ -1,0 +1,21 @@
+Karapace security
+=================
+
+By default Karapace Schema Registry and Karapace REST Proxy security is limited to authentication. All authenticated users have full access to both. All users configured in the service have access. You can find the URL for both of these services in the Console. The URL embeds a username and password.
+
+Karapace Schema Registry
+------------------------
+
+For more granular access control user authorization can be enabled in Karapace Schema Registry by enabling the flag `karapace_schema_registry_acl`. Authorization is only supported for Karapace Schema Registry, not for Confluent Schema Registry.
+
+Karapace Schema Registry access control lists (ACL) is stored in Aiven platform and Aiven platform forwards the configured ACL to Karapace Schema Registry. Aiven Client can be used to manage entries int the ACL. Currently there's no UI support to manage the ACL.
+
+When Karapace Schema Registry receives an HTTP request from an HTTP client, the incoming HTTP request contains username and credentials. Karapace Schema Registry authenticates the user. Karapace Schema Registry verifies that the user has access to the requested resource based on the ACL. Some endpoints do also filtering of the returned content based on the ACL.
+
+Karapace REST Proxy
+-------------------
+
+For more granular access control user authorization can be enabled in Karapace REST Proxy by enabling the flag `karapace_rest_proxy_acl`.
+
+When Karapace REST Proxy receives an HTTP request from an HTTP client, the incoming HTTP request contains username and credentials. Karapace REST Proxy uses these credentials when it connects to Aiven for Apache Kafka® and authenticates the connection. Kafka® uses access control lists (ACL) and will either allow or deny the request and return the response to Karapace REST Proxy. Karapace REST Proxy in turn forwards the response to the HTTP client.
+


lornajane:
Thanks for bringing this! I will admit that I'm confused so far though, do you think we could see some examples of changing the ACLs for both the REST API and the schema registry? And perhaps some information about what happens if access is denied in either of those contexts. Also is this feature so that a user who can't access the schema for a message shouldn't be able to access the message, is that the idea? I think we should explain what this is for as well, if there are one or two clear problems it solves.

juha-aiven:
> do you think we could see some examples of changing the ACLs for both the REST API and the schema registry?

I added an example how to add an ACL entry in this PR. This is for Schema Registry only, REST API authz we don't have yet.

> And perhaps some information about what happens if access is denied in either of those contexts. 

Added explanation of the error code.

> Also is this feature so that a user who can't access the schema for a message shouldn't be able to access the message, is that the idea? I think we should explain what this is for as well, if there are one or two clear problems it solves.

I'm not sure if I'm following. The Kafka topic ACL controls access to messages themselves in Kafka. Schema Registry ACL controls the access to the schemas etc stored in Schema Registry. So these are related, but not connected. Both ACLs are managed separately.

ftisiot:
Thanks @juha-aiven for the contribution, I'm currently reorganising the content to follow the devportal structure, I'll push a commit soon with the new org and then I'll tag you for more details to the sections I feel missing

ftisiot (changes requested):
Hi @juha-aiven good work! I reviewed the article and split it following the devportal structure.
Could you please check the various articles if they make sense?

Also it would be good to add:
* a list of Karapace endpoints and the need for read or write options
* A better description of what the `Global:` does... and what options can be associated with it? I didn't fully got it

ftisiot:
I believe this is ready to be reviewed, since I've been involved a lot I would hope to have an external reviewer.

I'm a bit concerned that the Karapace endpoints are not well explained, but that might be a joint effort between devportal and karapace docs. But mainly we can tackle it as with an additional PR and have this initial PR go live in devportal as it it since the feature should go live soon

docs/products/kafka/concepts/schema-registry-authorization.rst
```diff
@@ -0,0 +1,103 @@
...
+      
+  * ``*`` matching any characters
+  * ``?`` matching a single character
+
+The algorithm to determine if a request has the necessary permissions is the following:
+
+* Iterate over ACL entries in the order
+* Check if the current requesting user matches the ACL entry and the resource in the ACL entry matches the accessed resource
+* If it matches, use that ACL entry. Check whether the operation in the ACL entry matches the accessed operation, based on that grant or deny access
+* If no ACL entry is found, deny access, with HTTP response having status code 401
```
tvainika, line 53: This is not how this has been implemented. The implementation grants access if any of ACL entries matches, and allows access. The order does not matter.

docs/products/kafka/concepts/schema-registry-authorization.rst
```diff
@@ -0,0 +1,103 @@
...
+
+Endpoints and required ACLs
+---------------------------
+
+Karapace schema registry offers a series of endpoints to:
+
+* Get/set the global configurations: ``/config``
+* Get/set schema subjects and associated details: ``/subject``
```
tvainika, line 61: There is also `/compatibility/subjects/<subject:path>/versions/<version:path>`, `/schemas/ids/<schema_id:path>/versions`, and `/schemas/ids/<schema_id:path>` endpoints related to authorization.

docs/products/kafka/concepts/schema-registry-authorization.rst
```diff
@@ -0,0 +1,103 @@
...
+    - ``schema_registry_read``
+    - ``Config:``
+    - Read access for ``user_1`` to the global compatibility configuration
+  * - ``user_1``
+    - ``schema_registry_read``
+    - ``Subject: s1``
+    - Read access for ``user_1`` to the subject ``s1``. When filtering results in list endpoints, only return the results related to subjects ``s1``.
+  * - ``user_1``
+    - ``schema_registry_write``
+    - ``Subject: s1``
```
tvainika, line 88:
```suggestion
    - ``Subject:s1``
```

docs/products/kafka/concepts/schema-registry-authorization.rst
```diff
@@ -0,0 +1,103 @@
...
+    - operation
+    - resource
+    - explanation
+  * - ``user_1``
+    - ``schema_registry_read``
+    - ``Config:``
+    - Read access for ``user_1`` to the global compatibility configuration
+  * - ``user_1``
+    - ``schema_registry_read``
+    - ``Subject: s1``
```
tvainika, line 84:
```suggestion
    - ``Subject:s1``
```

docs/products/kafka/concepts/schema-registry-authorization.rst
```diff
@@ -0,0 +1,103 @@
...
+    - Read access for ``user_1`` to the subject ``s1``. When filtering results in list endpoints, only return the results related to subjects ``s1``.
+  * - ``user_1``
+    - ``schema_registry_write``
+    - ``Subject: s1``
+    - Write access for ``user_1`` to the subject ``s1``. Write implies read access, see the above row.
+  * - ``user_readonly*``
+    - ``schema_registry_read``
+    - ``Subject: s*``
```
tvainika, line 92:
```suggestion
    - ``Subject:s*``
```

docs/products/kafka/concepts/schema-registry-authorization.rst
```diff
@@ -0,0 +1,103 @@
...
+    - Read access for users with prefix ``user_readonly`` to the subjects with prefix ``s``. When filtering results, only return entities for subjects with prefix ``s``.
+  * - ``user_write*``
+    - ``schema_registry_write``
+    - ``Subject: s*``
```
tvainika, line 96:
```suggestion
    - ``Subject:s*``
```

docs/products/kafka/howto/enable-schema-registry-authorization.rst
```diff
@@ -0,0 +1,26 @@
+Enable Karapace schema registry authorization
+=============================================
+
+Schema registry authorization feature enabled in :doc:`Karapace schema registry <../howto/enable-karapace>`  allows you to both authenticate the user, and additionally grant or deny access to individual `Karapace schema registry REST API endpoints <https://github.com/aiven/karapace>`_ and filters the content the endpoints return.
+
+Enable Karapace schema registry authorization via Aiven CLI
+-----------------------------------------------------------
+
```
tvainika, line 8: Should we mention this requires newest aiven CLI too?

docs/tools/cli/service/schema-registry-acl.rst
```diff
@@ -0,0 +1,85 @@
...
+    - The resource to grant access to, can be in the form of ``Config:`` or ``Subject:<subject>``, accepts * and ? as wildcard characters accepts ``*`` and ``?`` as wildcard characters. More information can be found in the :ref:`dedicated page <karapace_schema_registry_acls>`.
+  * - ``--username``
+    - The username pattern: accepts ``*`` and ``?`` as wildcard characters
+
+**Example:** Add an ACLs for users with username starting with ``userAB`` to write (``schema_registry_write``) subjects having name starting with ``s123`` in the service ``kafka-doc``.
+
+::
+
+  avn service schema-registry-acl-add kafka-doc \
+    --username userAB*                          \
```
tvainika, line 37: How about commands here and all over, if this is copy-pasted to local shell and cwd contains file named userABC, this will cause unexpected behavior? How about quoting `'userAB*'`?

docs/tools/cli/service/schema-registry-acl.rst
```diff
@@ -0,0 +1,85 @@
...
+  * - Parameter
+    - Information
+  * - ``service_name``
+    - The name of the service
+
+**Example:** List the ACLs defined for a service named ``kafka-doc``.
+
+::
+
+  avn service schema-registry-acl-list kafka-doc
```
tvainika, line 84: How about adding example output here?

tvainika (changes requested):
Some misunderstandings found, and maybe some other useful suggestions

tvainika (approved):

TibsAtWork (approved):
* Added a few ® marks
* Split apart references to "Aiven for Apache Kafka® Karapace" to make it clearer that Karapace is a separate thing.
* Made some minor changes to the english

I believe I've not changed any *meaning* of anything, and thus this can be approved.

Issue #797, opened 2022-05-13
Migrate article: Aiven support details


lornajane:
Migrate the content from this article to fit the documentation portal format https://help.aiven.io/en/articles/868101

(Assign this issue to yourself before you start, and remember to mention the issue number when you open your pull request)

encima:
Support page has been put on the main site: https://aiven.io/support-services  now so maybe we do not need this?

KevinBrolly:
I believe this one can be closed as we no longer need https://help.aiven.io/en/articles/868101 and the https://aiven.io/support-services will suffice, the article currently redirects to there.

PR #869, opened 2022-05-20
Add support details page

FloorD:
Closes Migrate article: Aiven support details #797

### DIFF
--- a/_toc.yml
+++ b/_toc.yml
@@ -585,8 +585,8 @@ entries:
   - file: docs/community
     title: Community
     entries:
+    - file: docs/aiven-support-details
     - file: docs/community/documentation/index
       entries:
       - glob: docs/community/documentation/tips-tricks/*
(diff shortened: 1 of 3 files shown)


lornajane:
@FloorD Assigning back to you to check the build errors so that someone can review (but shout if you need a hand!). The broken link is the one on the main page, check this on your local setup?

laysauchoa:
Hi, do you need help to fix the build? also, are you able to run it locally? let me know if I can help :)

lornajane:
Rebased to resolve the navigation conflict BUT I don't think that "community" is the right place for the support details. This could be a top-level platform section link perhaps? I'd also consider adding it at the bottom of the main page or something like that so people can find the info....

lornajane:
We don't need this, the support pages have moved to https://aiven.io/support-services so the page no longer needs migrating.

PR #875, opened 2022-05-23
Improve venv instructions

juha-aiven:
Replaces PR #866 that had strange issues in the checks.

Simplify the venv create command.
Clarify the .gitignore-related note.
Add the exact command to activate the venv.


TibsAtWork (approved):
Thank you for fixing this (and for coping with the strange way the checks were failing)

Issue #802, opened 2022-05-13
Migrate article: Selecting the cloud and region


lornajane:
Migrate the content from this article to fit the documentation portal format https://help.aiven.io/en/articles/489585

(Assign this issue to yourself before you start, and remember to mention the issue number when you open your pull request)

PR #880, opened 2022-05-25
Migrate migrate to cloud/region article

laysauchoa:
This article has been partically
already migrate, so I am moving the
extra information to there.

It brings new info about the migration itself,
but not about the migration steps.

Fixes: #802


ftisiot (approved):
LGTM

Issue #298, opened 2021-10-25
PostgreSQL best practices checklist


lornajane:
Let's replace this article https://help.aiven.io/en/articles/4738831-basic-postgresql-best-practices with more detailed information on HOW to do the things mentioned (link to them if they should be their own articles) and add any more tips that spring to mind. More than just a migration task.

lornajane:
I don't agree that this should be closed. The original article linked here also had useful content about connection pooling and read-only replicas. We would need to bring that over before adding the redirect and marking this issue as closed (the content won't be accessible once the redirect is in place).

ftisiot:
I believe the correct redirect for this one might be https://developer.aiven.io/docs/products/postgresql/howto/optimize-pg-slow-queries.html

lornajane:
I think the connection pooling content is pretty much covered, but some of this read replica content would be good to bring. I'll PR this

lornajane:
Closed in #896

PR #896, opened 2022-06-01
Add best practices notes to existing article

lornajane:

Picked up the rest of the source issue at #298 so we don't lose anything when we add the redirect.


ftisiot (approved):
LGTM

PR #897, opened 2022-06-03
Add section: authenticated email destination

ebeta-aiven:
Following the fix https://aiven.atlassian.net/browse/MA-1206, the
article Create alerts with OpenSearch Dashboards needs to be amended.

- Add an additional procedure for creating email destination with authenticated sender. There is is a preparatory step to be executed in console which the user is not expected to be aware of.
- Remove an existing note, which states that this feature is not supported.


safa-topal (approved):

lornajane:
Thanks @ebeta-aiven :)

PR #904, opened 2022-06-07
mysql: add php example

sluhur:
code/products/mysql/connect.php
	docs/products/mysql/howto/connect-with-php.rst

### DIFF
--- a/_toc.yml
+++ b/_toc.yml
@@ -485,6 +485,7 @@ entries:
               - file: docs/products/mysql/howto/connect-from-cli
               - file: docs/products/mysql/howto/connect-with-python
               - file: docs/products/mysql/howto/connect-with-java
+              - file: docs/products/mysql/howto/connect-with-php
           - file: docs/products/mysql/howto/create-database
           - file: docs/products/mysql/howto/connect-from-mysql-workbench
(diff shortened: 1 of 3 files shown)


lornajane:
Could you add some context about why this change is needed? Does it fix a bug, add a feature, or something else?

laysauchoa:
I could not find a php example for mysql, could it be that you forgot to commit the file?

sluhur:
> Could you add some context about why this change is needed? Does it fix a bug, add a feature, or something else?

This is an example on how to connect to MySQL using php. We have example on how to connect to PostgreSQL but not MySQL.

sluhur:
> 
I have re-commit the files again.

laysauchoa:
Hi, this seems good @sluhur! Thanks for adding this example. @lornajane would be possible to help us to review? I'm not familiar with php code.

lornajane (changes requested):
This code sample doesn't work for me, and looks like a straight copy of the postgresql one which I'm not sure I'd quite expect since the SSL handling is probably different - did you test the code when creating this? Could you take another look please?

lornajane (approved):

lornajane:
Thanks! I made some minor changes to make it a bit more "PHP" but this is a nice addition.

PR #909, opened 2022-06-07
Pride logo update

ftisiot:

New pride logo for devportal!


ftisiot:
Yep will do that, waiting for @FloorD to check with the team and act on merge

FloorD (approved):
Built locally and LGTM!

laysauchoa:
**Timeline**
Just coming back about the timeline for where the image would be support, I've found this here: https://aiven.io/blog/celebrating-our-colorful-cast so we should just point out that timeline is 11 months from now and in case, it should change someone can take care of it.

**Tab image:**
Screenshot 2022-06-08 at 11 27 17

Is there a reason why the tab image was not replaced as well? I think it would be nice to actually support full mode in this case. What do you think?

**Tests:**
I've tested this with both mode: dark and light.

Screenshot 2022-06-08 at 11 12 36
Screenshot 2022-06-08 at 11 25 25

laysauchoa:
Should we delete this branch for now?

PR #912, opened 2022-06-08
Update Redis maxclients for maximum connections

runwuf:
Updated Redis documentation to state this number is an estimated number based on memory avaiable, using CLI example to see the exact maximum connections number.


laysauchoa:
This is a nice addition and I've tested

✦ ❯ echo "info" | redis-cli -u $REDIS | grep maxclients
Warning: Using a password with '-a' or '-u' option on the command line interface may not be safe.
maxclients:15688

laysauchoa:
This is a nice addition and I've tested

✦ ❯ echo "info" | redis-cli -u $REDIS | grep maxclients
Warning: Using a password with '-a' or '-u' option on the command line interface may not be safe.
maxclients:15688

PR #926, opened 2022-06-09
Improving wording

laysauchoa:
Better to have explicit things
because it would help search engines
to find this article


ftisiot (approved):
LGTM

PR #927, opened 2022-06-09
Fixing canonical for index pages

ftisiot:

We have been reported that devportal index pages create canonical links that are not working therefore being asked to remove the `index` part from the canonical. This PR is an attempt to do it

### DIFF
--- a/_templates/base.html
+++ b/_templates/base.html
@@ -3,7 +3,11 @@
 {%- block extrahead %}
 {{ super() }}
 
-<link rel="canonical" href="https://developer.aiven.io/{{ pagename }}" />
+
+<link rel="canonical" href="https://developer.aiven.io/{% if '{}'.format(pagename).split('/')[-1] == 'index' %}{{ pagename | replace('/index','') }}{% else %}{{ pagename }}{% endif %}" />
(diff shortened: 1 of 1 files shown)


anelook:
This is cool, Francesco!! I see no redirects!

One thing - can I suggest that we document a bit better the pieces and the logic behind the changed line? So that if 3 months we still remember the reasoning behind it 

I'm trying to process this line, and it seem to be a mix of javascript with something else I'm not really familiar with (is it python?) :D -
{% if '{}'.format(pagename).split('/')[-1] == 'index' %}{{ pagename | replace('/index','') }}{% else %}{{ pagename }}{% endif %}

I'm also wondering about the risks of how IE and other browsers will read/process this line. Is there a risk that it is browser dependent functions?

ftisiot:
Yeah I'm going to add a comment about the rest of the code as well :D 
Regarding the browser compatibility, the python functions are standard but also, I'm really not sure where this is executed (Is it server side since it's python?)

lornajane:
I'm closing this until we have a better handle on the problem, we need to discuss the requirements, and how we can produce the data that's expected rather than patching it in one place.

PR #931, opened 2022-06-09
How to enable Karapace for Kafka

lornajane:

We had a question from a customer about whether they had to run Karapace themselves, because they couldn't find what they needed. Adding a document that explains that Karapace is already included with a Kafka service, covers how to turn it on, and signposts the main project documentation (which itself is a work in progress, but we'll keep most Karapace docs there rather than here).


ftisiot (approved):
LGTM!

Issue #864, opened 2022-05-20
Terraform Recipe Request - Kafka service with custom configurations


dewan-ahmed:
# What should we add to our content collections?

Create a new Terraform Cookbook Recipe for Kafka with Kafka Connect services with custom configurations

# Suggested instructions for this task

1. Review [Terraform basics](https://www.terraform.io/intro)
2. Review [setting up first Aiven Terraform project](https://developer.aiven.io/docs/tools/terraform/get-started.html)
3. Run the [Terraform code](https://github.com/aiven/aiven-examples/blob/master/terraform/kafka/kafka.tf) to create a Kafka service with custom configuration. Feel free to do some research over specific custom configurations that are being used in the industry.
4. Check out [the existing recipes ](https://developer.aiven.io/docs/tools/terraform/reference/cookbook.html) and create a PR for the new recipe.

sluhur:
https://github.com/aiven/devportal/pull/941

PR #941, opened 2022-06-14
Kafka service with custom configurations

sluhur:

Resolves https://github.com/aiven/devportal/issues/864

### DIFF
--- a/_toc.yml
+++ b/_toc.yml
@@ -125,7 +125,7 @@ entries:
             - file: docs/tools/terraform/reference/cookbook/kafka-connect-terraform-recipe
               title: Apache Kafka and OpenSearch  
             - file: docs/tools/terraform/reference/cookbook/multicloud-postgresql-recipe
-              title: Multicloud PostgreSQL 
+              title: Multicloud PostgreSQL
             - file: docs/tools/terraform/reference/cookbook/kafka-flink-integration-recipe
               title: Apache Kafka and Apache Flink
             - file: docs/tools/terraform/reference/cookbook/kafka-mirrormaker-recipe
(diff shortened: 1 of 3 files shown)


sluhur:
> 

@dewan-ahmed I've done second commit and now the file is there.

sluhur:
The vale check is suggesting using Kafka instead of kafka but the kafka used everywhere even in the config.

ftisiot:
We usually solve this by leaving all the references to Kafka as Apache Kafka (capitals included).
If there is a parameter like kafka_user, we write it as code like ``kafka_user``. This makes vale happy :D

dewan-ahmed:
Hi @sluhur I've added a few fixes via https://github.com/aiven/devportal/pull/941/commits/0430e52b3995fcd4da938971a4f286a15b487595 and https://github.com/aiven/devportal/pull/941/commits/f5aad80a6c56aac8673a498c5c74e3a3e27d312e:

1. Fixed the _toc.yml conflict with ``main``
2. Fixed the Vale error
3. Added a Mermaid diagram
4. Added a few lines to the introduction
5. Added Terrraform syntax highlighting and formatting
6. Removed the additional ``variables.tf`` file and included a note to to the reader to check out the getting started guide (following the same pattern as the other recipes)

I'm adding few comments/questions in the review. Thanks and looking forward to another awesome recipe.

dewan-ahmed (commented):
Hi @sluhur 

1. Could you please check the Mermaid diagram if it makes sense? This is a draft diagram which I'll improve based on your feedback. What is the connector name here? What does it connect to? Do we need to show those details? From the console, it seems that we have a Connect service but no connector added. Without the connector, would the reader be able to test/validate the recipe?
Screen Shot 2022-06-19 at 2 19 37 AM

Could you please describe all the custom configurations within the recipe (may be right after the recipe)? There are tons so please trim the ones that we're not describing. Could we create a Kafka topic as well in this recipe? It'll be nice to add a few lines of instructions on how the user can test/validate the setup. For example, they might produce some data to some system and those data might end up on a Kafka topic via the connector. There can be two Kafka users and based on ACL, one of them can view the message and the other cannot. There can be a similar example for sink topic as well. We can choose either or. I don't know which source/sink connector we're using here so providing details on this will be nice.

2. I made a few changes within the terraform file. I removed the commented out lines and changed some of the resource names to follow previous recipe naming. What is the purpose of the ``output`` block here? Could we remove this? If not, could you please add a few lines on how the reader might be using the ``output``?

Thanks and looking forward to the next revision.

dewan-ahmed (approved):
LGTM.

Great work @sluhur 🎉

Issue #444, opened 2022-01-04
Add generated `List of advanced parameters`


laysauchoa:
This list can be added using the following command:

- [x] flink: https://github.com/aiven/devportal/issues/446
- [x] grafana: https://github.com/aiven/devportal/issues/448
- [x] kafka
- [x] m3db: https://github.com/aiven/devportal/issues/447
- [x] mysql: https://github.com/aiven/devportal/pull/415
- [x] opensearch: https://github.com/aiven/devportal/issues/445
- [x] postresql
- [x] redis: https://github.com/aiven/devportal/issues/449

This can be done:
`make service-type-config-<service_name>` and adding the right paths on the tree.

mhoffm-aiven:
it would probably be a good idea to add some periodic github action that generates docs and compares git diffs  to see if we have added/removed some parameters in aiven-core

lornajane:
@mhoffm-aiven I would love this! Do you know if there's a cron action or something?

ftisiot:
Just adding a comment on PostgreSQL here, IIRC what we did there was a one-off experience rather than using the new tool, should we align the process also for PG?

lornajane:
@ftisiot yes, at some point for PG

mhoffm-aiven:
> @mhoffm-aiven I would love this! Do you know if there's a cron action or something?

there is: https://docs.github.com/en/actions/learn-github-actions/events-that-trigger-workflows#scheduled-events

example from our terraform provider nightly acceptance tests: https://github.com/aiven/terraform-provider-aiven/blob/534912d9b9243c02a1c178aa5d16a433f11642ec/.github/workflows/acceptance-tests.yml#L5

laysauchoa:
All the PRs have been merged. Can this be closed or should leave it open for discussions purpose?

lornajane:
Let's leave it open and figure out the automation piece

PR #945, opened 2022-06-14
Automatic generate advanced params files

laysauchoa:
**Motivation** 
We have it used the aiven/scripts to automate documentation. This is a next step where we can run that time to time and keep our documentation update for those automated docs. 

I have decided to add automatic docs in one single workflow to avoid PR noise. However, I can split in two workflows if you think is better (cloud and service advanced params). The PR does not automatically merges, so there will be a manual review to check it out.

**Frequency:**
Not sure how do you feel about the frequency here. I think once per week is okay. But open to suggestions.

**Example:**
Find an example of the automation here: https://github.com/aiven/devportal/pull/952

Fixes: https://github.com/aiven/devportal/issues/444


lornajane:
Nice! AND our python code benefitted from you looking at it, I knew it was a mess and this is much better :)

If we can, I think it would be good to have separate PRs for each of the `make` commands, so that if things change in multiple products and one can't be merged, the other one isn't blocked - and they might need different reviewers if we need to get a second opinion from someone who has worked on a specific area.

Makefile
```diff
@@ -54,35 +54,37 @@ create-feedback-table:
...
 service-type-config-grafana:
-	python "$(SOURCEDIR)/scripts/aiven/service_type_config.py" "grafana" > includes/config-grafana.rst
+	python "$(SOURCEDIR)/scripts/aiven/service_type_config.py" "grafana" "includes/config-grafana.rst"
 
 service-type-config-influxdb:
-	python "$(SOURCEDIR)/scripts/aiven/service_type_config.py" "influxdb" > includes/config-influxdb.rst
+	python "$(SOURCEDIR)/scripts/aiven/service_type_config.py" "influxdb" "includes/config-influxdb.rst"
 
+service-type-config-all:
+	python "$(SOURCEDIR)/scripts/aiven/service_type_config_all.py"
```
mhoffm-aiven, line 87: I wonder if this is really needed, we could after all just use make and use the other commands as dependencies

laysauchoa: not needed, fixed

mhoffm-aiven (approved):
lgtm

laysauchoa:
Hi @lornajane would be possible to review this PR before I merge it? Also, we need to be aware that the individual PR will be open on Tuesday for each service included. 

Obs: Although, we can still fetch in the document information about elasticsearch, I've not added a workflow for this service as we no longer support it.

lornajane (approved):
Can we give each job a separate name? It's not a showstopper, and I'm approving and merging anyway but if you could follow up and edit the names, it would help to understand in the logs if one fails.

Thanks for setting this up!

PR #950, opened 2022-06-15
Migrate CLI tips-and-tricks help article

dragonlancewu:
Migrated CLI tips-and-tricks help article to developer portal


runwuf:
should this `docs/tools/cli/tips-and-tricks.rst` linked from `docs/tools/cli.rst`?

laysauchoa:
I found this youtube video for aiven CLI. Maybe is another nice resource to add here: https://www.youtube.com/watch?v=nf3PPn5w6K8

TibsAtWork:
Thanks for looking at this - contributions are very welcome!
And whilst I'm suggesting below that we not migrate this article, we wouldn't know to think about that if you hadn't started this PR, so this is still very valuable.

I believe it's probably best to hold off on this PR, for several reasons:

* We're currently working on a new blog post on using the `avn` CLI command, which will be more an end-to-end usage story. At that point we'll stop linking back to the old post.
* We don't really have a "tips and tricks" category in our documentation. (We follow the [Diátaxis framework](https://diataxis.fr/), so this would need to be reframed in those terms.)
* There's only really one "tip" here, and I believe we try to encourage people to use tokens for signing in from the command line. Otherwise, there is already https://developer.aiven.io/docs/platform/howto/enable-aiven-password.html which more or less covers this case, I think.

I *think* it might be appropriate to redirect https://help.aiven.io/en/articles/3074410-aiven-command-line-interface-tips-and-tricks to https://developer.aiven.io/docs/tools/cli.html - this is an action done outside the PR process.

laysauchoa:
I agree about the redirection idea. Maybe would be the best approach here. should we help the contributor to add this to the redirection process?

TibsAtWork:
I think sorting out the redirection is on us (DevRel), as it's something we'd normally do as part of the PR approval process, and it's mostly bookkeeping.

dragonlancewu:
@TibsAtWork and @laysauchoa Thank you for your feedback.

PR #951, opened 2022-06-15
Suggest adding `Fixes <issue>` text in a PR

TibsAtWork:
It's not very obvious that we can link a PR to the associated issue, and that if we do that, the issue will be close when the PR is merged.

Add some text to the CONTRIBUTING file to suggest this.

**Question:** is this one complication too many in our instructions?

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -47,6 +47,8 @@ The instructions for setting up local development are in the ``README``.
 
 2. Open a pull request, use the template so we know what it is and why it's a great addition. If you need to publish this quickly for a reason, mention that as well.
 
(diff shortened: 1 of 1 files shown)


ftisiot (approved):
LGTM!

TibsAtWork:
In this instance, I wanted to add it because it wasn't something the people I was working with knew about.
Given it makes it a lot easier to track back to the original issue, and also that it automatically closes the issue when the PR is merged, it seemed worth calling out.
(and in this case, we also had an issue that no-one had remembered to close, despite the PR being merged some while back)

But we could maybe remove this once we've got better training resources in place?

PR #953, opened 2022-06-15
Migrate Billing Groups help article

encima:

Migrated #826 from help.aiven.io. Removed the screenshots and cleaned up some of the text to highlight the primary billing group

### DIFF
--- a/_toc.yml
+++ b/_toc.yml
@@ -15,8 +15,6 @@ entries:
         title: Billing
         entries:
           - file: docs/platform/concepts/tax-information
-          - file: docs/platform/concepts/billing-groups
-          - file: docs/platform/concepts/use-billing-groups
       - file: docs/platform/concepts/beta_services
       - file: docs/platform/concepts/cloud-security
       - file: docs/platform/concepts/logs-metrics-alerts
@@ -26,6 +24,7 @@ entries:
       - file: docs/platform/concepts/static-ips
       - file: docs/platform/concepts/tls-ssl-certificates
       - file: docs/platform/concepts/byoa
+      - file: docs/platform/concepts/billing-groups
       - file: docs/platform/concepts/enhanced-compliance-env
       - file: docs/platform/concepts/disaster-recovery-test-scenarios
     - file: docs/platform/howto
@@ -63,7 +62,10 @@ entries:
           - file: docs/platform/howto/monitoring-services
           - file: docs/platform/howto/integrations/prometheus-metrics
           - file: docs/platform/howto/integrations/datadog-increase-metrics-limit
-          
+      - file: docs/platform/howto/list-billing
+        title: Billing management
+        entries:
+          - file: docs/platform/howto/use-billing-groups
       
     - file: docs/platform/reference
       title: Reference
--- a/_toc.yml
+++ b/_toc.yml
@@ -15,6 +15,8 @@ entries:
         title: Billing
         entries:
           - file: docs/platform/concepts/tax-information
+          - file: docs/platform/concepts/billing-groups
+          - file: docs/platform/concepts/use-billing-groups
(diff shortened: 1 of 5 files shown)


ftisiot:
Hi @encima good work, I split the article in concepts and howto, can you please check if it makes sense?

_toc.yml
```diff
@@ -61,6 +62,10 @@ entries:
           - file: docs/platform/howto/monitoring-services
           - file: docs/platform/howto/integrations/prometheus-metrics
           - file: docs/platform/howto/integrations/datadog-increase-metrics-limit
+      - file: docs/platform/howto/list-billing
```
laysauchoa, line 65: We have this section for billing as well on the concepts side. So maybe we need to adopt just one certain term for that.

      - file: docs/platform/concepts/list-billing
        title: Billing

laysauchoa:
ping @ftisiot and @lornajane

ftisiot:
Yeah probably have a section dedicated to billing somewhere is a nice thing, since it'll probably be one of the recurring themes for support requests, and keeping all the resources "together" would help people navigate. WDYT @lornajane ?

lornajane:
In line with what we've done in other sections, think about WHO needs to access groups of articles, and that might help? I would also recommend that we check how many articles we think we will have in this area - if it's ten or a hundred, that might help us understand what sort of structure we need. @ewe-gra might have advice for us

ewe-gra:
In Help Center we got 10 billing articles (https://help.aiven.io/en/collections/341199-billing) - not suggesting we keep as it is, would need to think about it. However, I do think that we should keep Billing under Billing (and it exists now in the docs (Platform/Concepts/Billing). If we think about the user: whoever sets up the billing, wants to know all about it from one place -- BILLING. 

Additionally, I wouldn't overcomplicate with too many sections under Billing. If I understood correctly, our Billing teams are working on improvements. As long as we explain what users need to do to change name, add details, create billing groups (current docs on Help Center do that) I wouldn't overcomplicate it if perhaps the whole "account" concept improves in Aiven. 

@encima I read your contributions under in both sections (How to & Concept):
- Will you also add the step by step how to create a billing group? 
- I'd bring all your billing articles to Platform/Concept/Billing...

laysauchoa:
I'd suggest to have all the billing stuffs on: (Platform/Billing) section. And as @ewe-gra said, no sub sections for now

ftisiot:
I like the suggestion, but there's a problem... the current `Billing` section exists under Concepts, while one of the articles is clearly an `HowTo`. I believe we kept them separate across all devportal, not sure we should do a distinction here.

ewe-gra:
Lorna and I chatted that we should have Billing as a separate section under Platform but not Concept. However, I'd still bring all billing articles in DevPortal under Billing right now (and then make the navigation change).

ftisiot:
👍

ftisiot:
@ewe-gra @laysauchoa can you please check the organisation now follows what we have been discussing?

lornajane:
The article about managing groups is a HowTo so it should be in that section rather than under Concepts. I completely agree that we have some information architecture problems in this section and I suggest that we:
* open a new issue to review the information architecture in the Platform section. My suggestion: instead of top-level Concepts and HowTo, do another level of topics/groups under the Platform section before splitting to concepts/howto, this needs some review of the full set of current and expected content which is why I think it should be separate from this PR rather than blocking work that has already been done.
* Update this PR to have the task-based article in the HowTo section for consistency, and get the PR published if everything else looks good. Ensure that there are good links between these two articles that will be separated a bit in the navigation until we sort out the structure.

ftisiot:
The Platform reorg can now be tracked and discussed in #1050

The latest push should put all articles in the temporary correct place.

laysauchoa (approved):
looks good!

Issue #828, opened 2022-05-13
Migrate article: Remote syslog integration


lornajane:
Migrate the content from this article to fit the documentation portal format https://help.aiven.io/en/articles/2933115

(Assign this issue to yourself before you start, and remember to mention the issue number when you open your pull request)

PR #955, opened 2022-06-15
Moving rsyslog help article

z0r1k:

Closes https://github.com/aiven/devportal/issues/828


lornajane:
I moved the DataDog stuff already as well and I would be so happy to get this content moved! Great to see this work

lornajane:
I think it would be even better to have focused articles per integration, as I did for DataDog. I see a lot of questions about this where people haven't been able to find our resources, or think that we don't support the integration they are looking for. We can keep a generic rsyslog article as well, in case anyone needs something we haven't specifically documented.

Reviewers: as with all the docs, we'll need to test all the steps before we merge ...

runwuf:
Got it @lornajane
For the integrations in this article, I can confirm datadog and coralogix integration works, I have validated these two integration in the past.

TibsAtWork:
Our fault for not being specific on this, but in general we prefer to rebase a branch on main rather than merge main into a branch.
This needs a little bit more collaboration between people, but gives a cleaner history.
We're very happy to do it for you if you don't want to fight git.

runwuf:
I will work with @dragonlancewu to rebase this on main.

runwuf:
@TibsAtWork - we rebased from the main and it looks good now.
@z0r1k - I've overwritten your merged main commit so I can do a clean rebase, you'll have to reset your local branch with remote to keep them in-sync.

docs/integrations/rsyslog/index.rst
```diff
@@ -0,0 +1,226 @@
...
+::
+
+   avn service integration-endpoint-create --project your-project \
+       -d loggly -t rsyslog \
+       -c server=syslog.collection.XX.sumologic.com \
+       -c port=6514 \
+       -c format=rfc5424 -c tls=true \
+       -c sd='collector-token-string@NNNNN'
+
+Datadog
```
laysauchoa, line 158: I believe this should be part of the datadog section instead of embedded here and here maybe a link to there, what do you think?

docs/integrations/rsyslog/index.rst
```diff
@@ -0,0 +1,226 @@
...
+
+Add rsyslog integration to service
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+This can be configured in the Aiven Console by navigating to the target
+service overview page and then scrolling down to the Manage Integrations
+button.
+
+
+.. image:: /images/integrations/remote-syslog-integration_image2.png
```
laysauchoa, line 81: I'd suggest to remove this image and just add leave it with instructions. You can see step 5 here: https://developer.aiven.io/docs/integrations/datadog/datadog-metrics.html

docs/integrations/rsyslog/index.rst
runwuf: I kept this section as it is using `avn` cli to add the service integration but added a link to the datadog integration
```:doc:`Further details on send metrics and logs to Datadog </docs/integrations/datadog/index>```
Let me know if you have better suggestions.

File: docs/integrations/rsyslog/index.rst (hunk 2 above)
Comment (runwuf) [reply docs/integrations/rsyslog/index.rst]: Removed this one as per suggestion.  I have also removed the other old screenshots and replaced with new screenshots from new Aiven console UI.

File: docs/integrations/rsyslog/index.rst [hunk 3]
```diff
@@ -0,0 +1,238 @@
...
+
+For coralogix integration you need to use custom format with logline
+
+::
+
+   avn service integration-endpoint-create --project your-project \
+   -d coralogix -t rsyslog \
+   -c server=syslogserver.coralogix.us -c port=5142 \
+   -c tls=false -c format=custom \
+   -c logline="{\"fields\": {\"private_key\":\"YOUR_CORALOGIX_KEY\",\"company_id\":\"YOUR_COMPANY_ID\",\"app_name\":\"%app-name%\",\"subsystem_name\":\"programname\"},\"message\": {\"message\":\"%msg%\",\"program_name\":\"programname\",\"pri_text\":\"%pri%\",\"hostname\":\"%HOSTNAME%\"}}"
```
TibsAtWork, line 213: Is `programname` meant to be replaced by specific text?

docs/integrations/rsyslog/index.rst
```diff
@@ -0,0 +1,238 @@
...
+
+-  ``syslogserver.coralogix.us``
+
+-  ``syslogserver.app.coralogix.in``
+
+LogDNA
+~~~~~~
+
+For LogDNA syslog integration, you would need to use custom format with
+logline. Please note that there are no backslashes around ``key`` value
```
TibsAtWork, line 230: Would the user expect backslashes around `key`?

TibsAtWork (commented):
I've done some reformatting of the text in the "Some formatting changes" commit.

I still need to work out what to do about the various product names (to stop the linting from complaining that they are unrecognised words). These are warnings (so not stopping the build), but I'd like to get them fixed. I'll *probably* add them to the dictionary - that's definitely my job, not yours 😄

I suspect it will also be a good idea to link to product pages in the text, but let's address that separately.

Let me know if you want help with the file reorganisation.

docs/integrations/rsyslog/index.rst
```diff
@@ -0,0 +1,229 @@
+Remote syslog integration
+=========================
+
+In addition to using Aiven for Elasticsearch to store the logs from your
```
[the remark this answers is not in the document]
TibsAtWork: ..I fixed it

docs/integrations/rsyslog/index.rst
runwuf: Removed this to avoid confushion.

docs/integrations/rsyslog/index.rst
runwuf: Nice catch, it should have been a variable `% programname%`

runwuf:
@TibsAtWork thanks for your review, please let me know what else needs to be fixed.

TibsAtWork:
I'll sort out the product names and then look over the article again.

docs/integrations/rsyslog.rst
```diff
@@ -0,0 +1,238 @@
...
+
+:doc:`Further details on send metrics and logs to Datadog </docs/integrations/datadog/index>`
+
+
+New Relic
+~~~~~~~~~
+
+You will also need a custom logline format for NewRelic Syslog
+integration. This is so you can prepend your `NewRelic Insights Insert
+Key <https://docs.newrelic.com/docs/apis/intro-apis/new-relic-api-keys/>`__
```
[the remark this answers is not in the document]
runwuf: Thanks for pointing this out @TibsAtWork , I have updated to use license key instead.

runwuf:
Thanks for all your work @TibsAtWork !
I've added a small ToC section on top of the page as I think it's useful to see what integrations are in this article without having to scroll through the page (I attempted to create links to anchor on the same file from _toc.yml but it looks like not supported).  I've also rearranged integrations in alphabetical orders.  What do you think?

TibsAtWork:
We shouldn't need the extra table of contents, because there should already be one at the right hand side:

Screenshot 2022-07-13 at 10 47 30

That goes away if the browser window is narrow enough (as does the main sidebar on the left, eventually).
Otherwise, if you're not seeing the "normal" right hand side ToC, let us know, as we probably need to fix that.

TibsAtWork:
I've made the following changes:

* Correct the link in the rsyslog overview page
* Remove the explicit ToC again, since there should automatically be one at the right of the page
* Add `alt` text to the images to make the vale linter happy
* Make the intro sentence for each integration more consistent, which also removes the "name/link standing by itself" I'd introduced
* Indent all the code samples the same - I went with 4 spaces rather than none, as that seemed to be more common
* Change the link to the `datadog.html` page to make vale happy

@runwuf I think this may now be good to go, for editorial purposes.

Have all of the examples been checked to make sure they're accurate / "work on this machine as well"?

docs/integrations/rsyslog.rst
```diff
@@ -0,0 +1,261 @@
...
+       -c tls=true -c format=custom \
+       -c logline='DATADOG_API_KEY <%pri%>1 %timestamp:::date-rfc3339% %HOSTNAME% %app-name% - - - %msg%'
+
+.. Note::
+   If you want to use Datadog EU environment, the service address is:
+   
+   * ``server``: ``tcp-intake.logs.datadoghq.eu``
+   * ``port``: ``443``
+
+.. Ideally the following link would use a ``:doc:`` role - for instance:
```
lornajane, line 163: I'm not sure what this is doing here but I think we're removing the section anyway so perhaps it doesn't matter

lornajane (dismissed):

lornajane:
I'd still really like these as separate articles, on the basis that most users will be implementing only one at any one time, even if we think of them from the back end as a single feature. However I don't want to block progress, so if we merge as-is, please open a followup issue to create dedicated guides for each platform, as we did for Datadog.

docs/integrations/rsyslog.rst
```diff
@@ -0,0 +1,261 @@
...
+need to use one of the following Syslog Endpoints for ``server``:
+
+-  ``syslogserver.coralogix.com``
+
+-  ``syslogserver.coralogix.us``
+
+-  ``syslogserver.app.coralogix.in``
+
+
+Datadog
```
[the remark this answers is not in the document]
TibsAtWork: @runwuf can I leave that to you to sort out?

TibsAtWork:
I agree with @lornajane about treating this as a directory of specific files in the future, which will actually be easier with the new ways we organise things. However I also think it's worth doing this "quick fix" now, and also it will actually make the redirection from the original help easier to use.

I can raise an issue for more work once we've merged.

runwuf:
@TibsAtWork  I had no idea of the existence of the toc on the right hand side until today!!! because I have always been using a narrow browser window... 🤦‍♂️
btw, I didn't invent `:class: this-will-duplicate-information-and-it-is-still-useful-here ` when I didn't define this class, it shows up as an alert on the page.

I can confirm Coralogix (I wrote this config) and Datadog works.

The following I have not tested personally, I'll drop a note to our SA team see if others have tested any of them:
- Loggly
- Mezmo (LogDNA)
- New Relic
- Papertrail
- Sumo Logic

runwuf:
@TibsAtWork I validated and updated ALL of these rsyslog integration configs.

TibsAtWork:
@runwuf Thanks very much for all the work on this!

I've moved the "for this remote syslog service you need *these* special considerations" text around a bit so all of the entries have a similar layout, with the setup information before the example (a colleague explained to me the other day that this makes it much more likely people will notice them). I've also done some more reStructuredText tidying, and made the vale linter happy by changing how a link is done.

Basically I think this is now all done, and shall approve and merge it.

I've raised issue https://github.com/aiven/devportal/issues/1148 to look at splitting it into sub-pages later on.

TibsAtWork (approved):
Approved

PR #958, opened 2022-06-15
Clarify this article excludes pg_upgrade and keep alive time for old primary server

runwuf:
* this procedure does not include major version upgrade with ``pg_upgrade``
* states keep alive time for the old primary before terminating (confirmed in core repo code).


ftisiot (approved):
LGTM

PR #965, opened 2022-06-17
Use netlify redirects file for renames

### LINKED ISSUE #565: Remove Flink Real time alerting solution CLI

We have a Real time alerting solution in the console already (https://developer.aiven.io/docs/products/flink/howto/real-time-alerting-solution.html)

The CLI is "another" way to accomplish the same and doesn't really add lots of value. We should remove and add redirects to avoid problems

[ ] I plan to open a pull request for this

### LINKED ISSUE #882: Handle redirects
If we rename or move a page, the URL will change. We should handle this with some sort of redirect rules, such as on Netlify we'd be able to take this approach: https://docs.netlify.com/routing/redirects/

lornajane:

If we need to rename a file, or delete a file, we didn't have a way of redirecting the URL.
This introduces [Netlify redirects](https://docs.netlify.com/routing/redirects/) so that we can have source and destination URLs and the redirects are handed at the webserver routing level. Checked with marketing ops and cloudflare supports something similar so this won't cause pain in our re-hosting project that is coming up.
Fixes #882  and fixes #565 (since it was a good example of what is also blocked by this missing feature).

It also adds some cheeky redirects for product names so you URLs like https://developer.aiven.io/redis will work ;)


anelook (approved):
@lornajane This is amazing! I wrongly assumed that the redirects is stored and handled through Netlify interface.

Issue #998, opened 2022-06-20
datadog: fix the links between articles


laysauchoa:
This articles [1] [2] has the wrong links [3], so you should redirect to the devportal information about datadog instead:
https://developer.aiven.io/docs/integrations/datadog/index.html

Pages affected: 
[1]: https://developer.aiven.io/docs/platform/howto/integrations/prometheus-metrics.html
[2]: https://developer.aiven.io/docs/platform/howto/integrations/datadog-increase-metrics-limit.html
[3]: https://help.aiven.io/en/articles/1759208-getting-started-with-datadog

PR #1001, opened 2022-06-20
datadog: replace links

laysauchoa:
the links are outdated since we have
already content in devportal regarding the
"getting started with datadog"

Fixes: #998


dewan-ahmed (approved):
I tested both links and they direct to the developer.aiven.io Datadog page now.

I cross-referenced with the migration spreadsheet and the swapping of the pages look correct.

Thanks Laysa.

Issue #934, opened 2022-06-10
SEO related: canonical tags of 44 URLs need to be replaced 


edithyeghiazaryan:
**The issue**: The canonical tags of 44 URLs are pointing to redirected pages. As a result, search engines fail to crawl and index the pages properly. https://docs.google.com/spreadsheets/d/13ILZY5Hh3CyuwX3qNlAn-hyEpr-HN0iDMHUL9FBzr74/edit#gid=0

**The solution**: To replace the current canonical URLs with self-referencing canonical tags.

_If additional info is needed, please let me know._

**Resources to look at:** 

**1. Previous discussions on slack:** https://aiven-io.slack.com/archives/C01SY3QU54P/p1654697321714689

Also, Angeline suggests: _"we can use Netlify redirect for the issue with index.html (row 23-44) 
But there is also this sitemap that has index.html and .html in all places https://developer.aiven.io/sitemap.xml -> this can also be configure i assume"_

**2.Educational useful materials:**

- Canonical tags: https://www.semrush.com/blog/canonical-url-guide/?kw=&cmp=DE_SRCH_DSA_Blog_Core_BU_EN&label=dsa_pagefeed&Network=g&Device=c&utm_content=515771357185&kwid=dsa-1057183191115&cmpid=11794204941&agpid=114887190419&BU=Core&extid=23661292104&adpos=&gclid=CjwKCAjw14uVBhBEEiwAaufYx7izGDIDsuLlNGvFyz9l24yrht5c5sqRK1JtFhkiLJCA4aONTJZHmBoCLMMQAvD_BwE

- Redirection - https://djtechblog.com/seo/redirect-default-index-html-to-root-seo-basics/#:~:text=In%20simple%20words%20we%20should,should%20always%20redirect%20to%20root

anelook:
@edithyeghiazaryan, one clarifying question -  "fail to crawl and index the pages properly" - what do we mean by "properly" here? are they ignored completely go google search engine?

edithyeghiazaryan:
@anelook 
Olena, thank you for the question. From the crawlers’ perspective, these pages are canonicalized to non-indexable URLs, which is a paradox for them. As a result, the rankings, impressions, and clicks will drop for these pages if they haven’t dropped already.

[...]

PR #1032, opened 2022-06-24
SEO: Rename index.html with corresponding pages

anelook:
This PR renames all index.html pages (with the exception of https://developer.aiven.io/index.html) with a name of the folder where they are located.
For example, https://developer.aiven.io/docs/products/flink/index -> https://developer.aiven.io/docs/products/flink/flink

To make sure that existing external links to our index.html pages continue working, new redirects rules are added into file _redirects. 

All internal mentions of index.html were replaced as well.

Fixes https://github.com/aiven/devportal/issues/934


lornajane:
I think this is probably the best solution of the available options, thank you for looking into this and making all the changes! How do you feel about moving the files up one level? The `docs/community` section already does things this way (and avoided being flagged in the SEO recommendations). So there we have `docs/community.rst` which is its landing page, available on https://. And `docs/community/*` which is all the content inside that section. This avoids the repeated values in the URLs. Most of the other subdirectories also work this way, such as all the howto and concepts pages which have pages and same-named directories.

anelook:
@lornajane , I agree and moved landing pages one level up, so there is no duplications in the urls.

anelook:
Closed. replaced with https://github.com/aiven/devportal/pull/1113

PR #1036, opened 2022-06-26
Sphinx design

### LINKED ISSUE #730: Migrate to sphinx-design
We're using [sphinx-panels](https://sphinx-panels.readthedocs.io/en/latest/) for our landing page and a few other nice touches on the site. However there's a successor to this which is [sphinx-design](https://sphinx-design.readthedocs.io/en/sbt-theme/index.html) which is designed to be easier to style and manage.

Remove the old plugin and add the new one, migrate all our uses of panels.

lornajane:

We use a plugin called [sphinx-panels](https://sphinx-panels.readthedocs.io/en/latest/) however it is no longer maintained and the recommendation is to use [sphinx-design](https://sphinx-design.readthedocs.io/en/furo-theme/) instead. This PR implements the change, and attempts to fix all the panels and badges that were in use on the site.

Fixes #730


laysauchoa:
Do you think would be a good idea to keep just one style for those selection button/cards for now? I like the one for Redis and for the Aiven community. I found the red selection of a button with red fonts less accessible.

lornajane:
The badges are things like the "beta" indicators we use on the ClickHouse and Flink pages, or the preview indicators here https://deploy-preview-1036--developer-aiven-io-preview.netlify.app/docs/products/kafka/kafka-connect/concepts/list-of-connector-plugins.html

Thank you so much for the feedback on the hover!! Of course I _can't_ hover so I didn't see this at all, this is super helpful, and I agree that the community page is a better version, I'll push similar changes for the main landing page.

For the cards without buttons, I don't think those should be clickable cards - they're more like fancy link containers than the action-based cards we have. Although, maybe we can loop in a UX expert for some advice on this while we are revisiting the controls anyway.

lornajane:
(also in case anyone is interested, the badges use rst's substitution syntax, so the "beta" and "preview" badges are only defined in one place, then re-used wherevere you need them)

index.rst
```diff
@@ -25,131 +25,168 @@ What can we help you with today?
...
+            :align: right
+            :color: primary
+            :outline:
 
-    .. link-button:: docs/products/flink/index
-        :type: ref
-        :text: Read more
-        :classes: stretched-link
-    ---
+            Read more
```
anelook, line 95: What do you think if we remove "Read more" buttons completely? Now that with the new approach the whole panel became clickable, it is already gives a hint where the user will be redirected.

anelook: Screenshot 2022-07-03 at 21 45 00

anelook: Unless we must have that button to have a linkable panel. But in that case - would be better if the border color is not red.

anelook (changes requested):
Thank you for making this change! It is more reusable and compact.
The looks changed a bit, and one thing that now looks a bit off to me - the border color of the button "Read more". I'd either get rid of the button if we can, or change its border color .

index.rst
lornajane: Thanks! This is really good feedback. We do need buttons (cards aren't usually a clickable UX element, and there's no visual clue, it's just a nice enhancement that you don't actually need to hit the button exactly). I agree that it looks a bit funky, and I might wait until we have #1037 with all its colour disasters in place before I finalise this.

lornajane:
Switching to draft as we will need to merge the theme change and rebase before we can tackle this

index.rst
lornajane: To be entirely correct, the button should be there, and the panel should _not_ be clickable since it's really a display thing, and the user might not expect it to be UI control. Anyway I've kept the clickable panel on the cards like this that have only a single button. Other cards have other content, such as a list of links, so having a single click destination on the panel makes no sense anyway (we probably should get some actual UX advice on this at some point, but right now I'm interested in upgrading the plugin).

I think it looks OK so I'll resolve and see if anyone else raises it still, then we can revisit.

ftisiot:
I would agree with @anelook here, and, for the sake of consistency, get rid of the "Read More" button in the main page, since we are highlighting the whole cell

lornajane:
There's a problem with the buttons, I can't click them (no idea how I'm only seeing this now). However I _also_ can't click the cards at all, which makes sense since they aren't a UI element - I will remove the link from the cards, we must have a button or a link so that users understand the thing is clickable, I tried adding the link to the card when there was only one .... but I don't think it makes sense now we've talked about it.

laysauchoa:
- The badges have changed. I prefer the previous one as looks more modern.
new

- Cards take me nowhere as you commented before.
Screenshot 2022-07-28 at 09 39 04

- I found the font here a bit off. Because in this context, looks different of all other fonts. But I guess is related to the theme?

Screenshot 2022-07-28 at 17 51 50

lornajane:
* Fixed the cards by removing their links and just keeping the buttons
* The "pill" styles aren't supported on the badges yet, so I'm afraid I can't fix that
* I don't understand what's wrong with the fonts, can you explain again @laysauchoa ? Is one of the styles using a different font (it looks OK to me but then it would even if every style used a different font so I appreciate the help on this one!)

laysauchoa:
> * Fixed the cards by removing their links and just keeping the buttons
> * The "pill" styles aren't supported on the badges yet, so I'm afraid I can't fix that
> * I don't understand what's wrong with the fonts, can you explain again @laysauchoa ? Is one of the styles using a different font (it looks OK to me but then it would even if every style used a different font so I appreciate the help on this one!)

I guess the other observations are more from a personal taste and not blockers. I will take a look again tomorrow. Are there anything that I can help to check and maybe I miss in the review? Let me know!

lornajane:
I don't have any other concerns, but see if you can find anything else before we go ahead!

laysauchoa:
Almost there! 😉

- I like the index button ``Read more`` more now. However, it gives 404 error when I click.
- Also, in the index page. I noticed that there is misalignment between left and right buttons. I do not think is intentional.

Screenshot 2022-07-29 at 07 53 00

I tested other pages and did not find any other issues.

index.rst
```diff
@@ -25,131 +25,155 @@ What can we help you with today?
...
-    +++
+        |icon-clickhouse| **ClickHouse** A highly scalable, open source database that uses a column-oriented structure. :bdg-secondary:`beta`
 
-    .. link-button:: docs/products/clickhouse
-        :type: ref
-        :text: Read more
-        :classes: stretched-link
+        .. button-link:: docs/products/clickhouse
+            :align: right
+            :color: secondary
```
anelook, line 173: Here we have "secondary", unlike other cases where color is set to "primary". In Light mode there is no differece. However, on hover in dark mode - the secondary looks way better to me! it has better contrast of the text. I'd prefer if we set secondary for others as well

conf.py
```diff
@@ -101,19 +101,21 @@
...
         "color-background-hover": "#c60443",
         "color-background-border": "#e1e1e3",
         "color-highlighted-background": "#1c1c2f",
-        "color-inline-code-background": "#787885",
+        "color-inline-code-background": "#747481",
         "color-sidebar-background": "#f7f7fa",
         "color-sidebar-background-border": "#e1e1e3",
         "color-sidebar-search-background": "#fff",
+        "sd-color-card-background": "#f7f7fa",
+        "sd-color-primary": "#4a4b57",
```
anelook, line 118: This value ends up also used for dark mode. But for dark mode the contrast gets too low. We need to add something brighter and to add it for "dark_css_variables".
For example, when looking at the search input border, in the dark mode we use there `--color-foreground-border: #e1e1e3; `. For me that color  plays similar role to button border, that's why we can use it for the button as well.
Screenshot 2022-08-03 at 12 02 40

anelook (changes requested):
I like that the new design makes it more compact and aligns the cards color with the menu.

I think @laysauchoa already noted the issues, but I added some more details on the colors, in case it is helpful.

anelook:
> 

I didn't spot any 404s, @laysauchoa did all links give you the error, or only some?

laysauchoa:
@anelook locally, I get the error 404, but I see that the previous is working correct on netlifly. I have some local issues then ^^

anelook:
@laysauchoa, I see what you mean now. 

Locally we don't have the server side which will open pages without .html ending. Links from cards lead to pages like http://127.0.0.1:8000/docs/products/postgresql/ and they when running locally return 404. With Netlify it is taken care by redirects and pretty urls functionality.

laysauchoa:
> @laysauchoa, I see what you mean now.
> 
> Locally we don't have the server side which will open pages without .html ending. Links from cards lead to pages like http://127.0.0.1:8000/docs/products/postgresql/ and they when running locally return 404. With Netlify it is taken care by redirects and pretty urls functionality.

Yeah, good point. The current version from `main` when I run locally takes me to ``http://127.0.0.1:8000/docs/products/postgresql.html``. Do we declare the redirects here somewhere? or where the URL changes happens in the PR?

anelook:
I've updated the color for  primary buttons in dark mode and in my opinion it looks better and the contrast is really good. But happy to hear other opinion.

anelook:
I also looked at the reason behind 404 errors which @laysauchoa spotted (true eagle eyes!🤩). The reason is that we switched from link-buttons to button-links. I can’t find good documentation on this, but these elements are rendered differently. 

Locally it helps when I add ".html", however, this breaks our link-check.

 I tried couple of workarounds but didn't find a good solution. Any opinion if this is an issue for us? I don't believe we use those buttons often locally.

lornajane:
I definitely don't think we can block this for a dev setup problem, because it's already blocking a couple of other key things. To me, this looks really good, but I can't merge my own PR. @ftisiot or @TibsAtWork , would you care to chime in?

ftisiot:
Will look into this this morning

ftisiot (approved):
This LGTM!

ftisiot:
@anelook I believe I can dismiss your review here and merge?

anelook (approved):

laysauchoa:
Hi, I believe the boxes on the landing pages were not aligned correct, so I create this https://github.com/aiven/devportal/issues/1277

Issue #515, opened 2022-01-27
Diagram display issue using dark mode


NotoriousPyro:
# What's wrong?

The two diagrams on these pages appear broken (difficult to see dark grey lines on black) when displayed on developer.aiven.io:
![image](https://user-images.githubusercontent.com/706132/151349556-55fe8085-5235-45f5-b415-bd79aa694ded.png)
![image](https://user-images.githubusercontent.com/706132/151349574-f4457f72-722d-4b9c-a258-c7d278bacb58.png)

# URL of affected page, and any other information
https://github.com/aiven/devportal/blob/a5331f4f0887862bc689f146474a8de443a0b955/docs/products/postgresql/concepts/pg-connection-pooling.rst

https://developer.aiven.io/docs/products/postgresql/concepts/pg-connection-pooling.html

Appears to be due to this CSS styling:
![image](https://user-images.githubusercontent.com/706132/151350034-b7007c48-b865-4095-9ab2-84d6f356fc43.png)

When changed to white, the lines are visible:
![image](https://user-images.githubusercontent.com/706132/151350101-5299927e-6d81-4efc-8f98-3e7ee3b0e243.png)

PR #1037, opened 2022-06-26
Update to newest version of Furo Theme (2022.06.21)

### LINKED ISSUE #184: Add a button to switch between normal/dark mode
# What's wrong?
The website detects automatically the color interface of my operational systems and uses it. This is great, since I personally prefer dark mode.

However, sometimes I'd like to switch to the "normal lighter theme", specially on articles with more images, like [PostegreSQL Connection pooling](https://developer.aiven.io/docs/products/postgresql/concepts/pg-connection-pooling.html), since the images are a bit hard to see on dark mode.

Is there a button for that? If not, is it possible to add one?

lornajane:

Keeping up to date with our dependent theme; this gives toggles for light/dark mode (so it fixes #184) , a jump-to-top button, and the ability to open the right hand bar in mobile/small screen mode to use the secondary navigation.

This required a LOT of hand-editing of the page templates (we are based on their templates, but there were quite a lot of changes since), and the way the light/dark CSS setup is handled was also completely changed in this release. We need to visually inspect that everything still looks good (or at least, as good as it did before), since there's a lot of scope for problems. On the plus side, you get a dark mode toggle so it's easier to check things!

Will probably conflict horribly with #1036 but I'll rebase whichever one didn't merge first.


laysauchoa:
@ftisiot I think the issue with the mermaid diagrams is that there is no support for dark mode from what I understood. I added some links here: https://github.com/aiven/devportal/issues/515

lornajane:
@laysauchoa I really don't know what to do with the German month names! It uses https://pypi.org/project/sphinx-gitstamp/ so there's a good chance the existing site does this also ... maybe we can fix this separately? I might need some technical support ;)

_static/css/aiven.css
```diff
@@ -44,169 +20,29 @@
...
 
 p {
-  color: var(--secondaryColor);
+  color: var(--color-content-foreground);
   font-family: Inter;
 }
 
 a {
-  color: var(--link) !important;
+  color: var(--color-link);
```
anelook, line 41: For dark mode this line sets the color to be red by default, which is different than befor. 

I see it references --color-link: #ff3554; While in the old version in was  --linkDark: #d2d2d6;
The hover colors also changed to --color-brand-content: #368ce2; while it was --linkDark: #d2d2d6; before (that's where the !important was enforcing it to be same color as non-hover)

anelook: Screenshot 2022-07-03 at 21 20 19

_static/css/aiven.css
```diff
@@ -1,32 +1,8 @@
 /* Assigning color variables */
```
anelook, line 1: --linkActive is no longer red, in the dark mode the hover on the menu item loses red side line. Based on other changes it should be probably changed with --color-link.

.sidebar-tree .reference:hover {
  background: none;
  border-left: 1px solid var(--linkActive);
  font-family: Inter;
}
.sidebar-tree .current-page > .reference {
  background: none !important;
  border-left: 1px solid var(--linkActive);
  color: var(--linkActive) !important;
  font-weight: 400 !important;
}

anelook: Additionally, there is 1px movement on hover, which is not needed. This might be related to the fact that the border-left is gone

anelook (changes requested):
There is sadly still an issue with the colors - in the dark mode all links are red. 
Also I noticed that the behaviour of hover on the menu items is different - the side line is gone and replaced with the  slight movement of the menu item (but maybe it is intended this way!)

laysauchoa:
> @laysauchoa I really don't know what to do with the German month names! It uses https://pypi.org/project/sphinx-gitstamp/ so there's a good chance the existing site does this also ... maybe we can fix this separately? I might need some technical support ;)

Thanks, this does not seem very relevant to fix, in my opinion. Just something different from the previous version.

_static/css/aiven.css
lornajane: This helps, think I fixed it now.

_static/css/aiven.css
lornajane: I didn't see the hover! I think I have this looking better but let me know what you think (when I push the commit...)

anelook (commented):
I like new changes! The dark mode highlights on hover are now white and not red. 
The issue with jumpy menu items is gone. The hover on cards is fixed.

I looked at the changes in Chrome/FireFox and Safari - all looks consistent. There are minor difference from the current version, but they look good to me.

Several things which might worth noting and checking if we agree that these are acceptable:

1. The hover effect on "copy" button is gone. But from what I see these are not our overwrites, but rather the change in the schema.

it comes from _build/_static/copybutton.css  (from furo assets _copybutton.sass)
.highlight button.copybtn {

where the border value was previously set to 
border: 1px solidvar(--color-background-border);

bun in the new version it is 
border: none;

2. The contrast difference between link color and hover color in dark mode is low - people with eye-sight issues might have difficulty noticing the change. It  does looks visually better for me comparing to the red color, but maybe we want to make it more noticeable.

3. In the light mode the color of menu links is now consistent with the rest of the links we have - and by default they are red. 

Overall, as a user of dark mode  - I find the visual changes to be pleasant to the eye. Can't judge much the light mode, since I'm not using it consistently.

lornajane:
Thanks @anelook! The copybutton is a really good spot, and in fact I fixed this by updating the plugin that does this rather than fiddling with anything on our side. Link colours should look better in the nav, I'm expecting some brand refreshes anyway so we will probably revisit the colours. As long as you don't think it's actually problematic, I think we are OK for now. Do you think we are close to ready?

ftisiot:
I didn't see this in a while and I like the changes!

A few minor things: 
* Note, Tips and Warnings are not as visible in dark mode as in the bright one
image
* Similar (can be seen in the above image) code highlight is not so visible
* Anchors are a different display (# vs P)
image

lornajane:
The anchor character is something that changed in the upstream theme so it's just ... different now. If you don't hate it, we'll keep it! I'll take a look at the other two things

ftisiot:
Sorry, didn't add the old for comparison. I believe the code now is way less evident than before.
Before it was in dark grey with a light grey background, now is in black and the background color is almost invisible
image

lornajane:
I actually quite liked it! Try the change I just pushed though, does this look better to you?

ftisiot:
Yep much better now! I'll have another deep look at this tomorrow morning!

anelook:
@lornajane, thank you for fixing the code snippets hover and other things! 

Overall, I love the current look of dark mode (I'll leave Francesco to judge the light one ;) ), I think the color set looks a bit brighter, not so dull ( though I'm not a designer :D )
All in all, I think there are no visible breaking parts or visible inconsistencies. All else we can improve later if needed. Hopefully we'll also get some feedback from others.
My only preference would be to keep as much as possible the "built-in" styles, since often they are done for a reason.

anelook (approved):
I do think it is good to be merged. Maybe some final glance from @ftisiot is good to have, because I see thatFrancesco wanted to give it another try tomorrow. And in any case better to merge during working hours, it's quite a big change.
I approve just in case.

ftisiot:
Found some more Red links, pages are
* https://deploy-preview-1037--developer-aiven-io-preview.netlify.app/docs/products/mysql/reference.html 
* https://deploy-preview-1037--developer-aiven-io-preview.netlify.app/docs/products/postgresql/concepts.html
 
probably we want to fix them for consistency?

image

Once this is fixed, the rest looks ok!

lornajane:
The tables of contents, got it. Is this version better?

PR #1042, opened 2022-06-27
added note on how to promote a read-replica to master via an API call

as-aiven:

As we had Decathlon asking on how to promote a read-replica to master via an API call, thought it might be nice to add it as a note to where we already show how we do this via Aiven Console


ftisiot (approved):
LGTM

PR #1044, opened 2022-06-27
Changed from a specific version (version 11) to all versions of PostgreSQL.

deanholbrough:
Currently the documentation states JIT is disabled for a specific version of PostgreSQL (version 11). This update changes the statement to all of the versions.


lornajane:
@deanholbrough could you take another look at this PR please? There's a LOT of change here, and I don't think it's what you intended. Check that your main branch is up to date before you make a new branch and add changes to it (I'm guessing at what happened here) - and please ask if you could use some help to figure this out!

PR #1045, opened 2022-06-27
Updated to include all versions of PostgreSQL, not just version 11

deanholbrough:
Currently the documentation states JIT is disabled for a specific version of PostgreSQL (version 11). This update changes the statement to all of the versions.


ftisiot (approved):
LGTM

PR #1090, opened 2022-06-29
added note regarding os migration might lead to os dashboard downtime…

as-aiven:
added note regarding os migration might lead to os dashboard downtime if nodes are of older version


as-aiven:
The note might not be in the best location/page, feel free to adjust, I appreciate feedback here

laysauchoa:
This seems very informative. However, I'd like to propose a more compact version of this:

_The Dashboards from Aiven for OpenSearch is not compatible across minor versions. Because of this, if existing current nodes of your service instance are of an older version, you can expect downtime during migration or plan changes._

I'd say that this should be a small section or a warning note in the OpenSearch Dashboards index (whatever you think is better), then you can link to where migrations, change plan are mentioned.


You can change the style from .. note to ..warning, for example:

```
.. warning ::

    Write info here.
```

You can then link in the relevant places like this:
.. seealso::

    You can find information about the OpenSearch Dashboards compatibility in :doc: <path>

Let me know what you think. This seems like an important information and we would like that the user can easily find it.

lornajane (approved):
Thanks for adding this @as-aiven , it looks good now

PR #1098, opened 2022-06-29
Remove automation for influx db

laysauchoa:
Currently, we do not have a landing page for this
service in the devportal. The automation runs weekly
and for now, seems like a wise decision to remove the
automation that generated advanced parameters for this
service.

We can add later when landing page exists.


ftisiot (approved):
LGTM

PR #1102, opened 2022-06-30
Adding a warning on token revocation.

dewan-ahmed:

I came to know that disabling 2FA would actually revoke all of your existing tokens. I think our docs need to show a warning to the reader for this.


laysauchoa (approved):
@dewan-ahmed thanks for adding this. I think is a great info for 2fa!

PR #1107, opened 2022-07-01
Fixed pool name documentation

palmieric:
As it is written right now in the doc, it is not clear to the customer that the pool name must be equal to the database name or pgbouncer is not capable to connect to the database server.
image
image


ftisiot (approved):
LGTM! Thanks!

Issue #938, opened 2022-06-13
Redis is now ®, not ™


TibsAtWork:
Redis has become a registered mark, rather than just trademarked.

1. Update the vale rule for Redis, and the corresponding tests for the rules
2. Replace all of the occurrences of `Redis™*` with `Redis®*`
3. Update the footer text acknowledging registration marks to say that `*Redis is a registered trademark of Redis Ltd. and the Redis box logo is a mark of Redis Ltd. Any rights therein are reserved to Redis Ltd.` (I think this just needs insertion of the word `registered` into the existing text).

laysauchoa:
Would be possible to help me to link the official information about this, @TibsAtWork?

TibsAtWork:
Apologies - I should have done that earlier. Thanks for pointing this out.

https://redis.io/docs/about/trademark/ is the relevant page at redis.io

The section on "General Use Information" (section 6) is most relevant here - it clarifies that we should be using `®*` (in clause i) and what the acknowledgement text should be (in clause ii)

PR #1108, opened 2022-07-01
Replace `Redis™*` with `Redis®*`

TibsAtWork:
* Change the vale rule that checks for `Redis™*` to check for `Redis®*`
* Update the tests for that rule
* Replace all the occurrences of `Redis™*` with `Redis®*`
* Update the Redis acknowledgement text to include the word "registered"

The change to the vale rule was simpler than I expected - it's now very similar to the other "needs an `®`" rules.

Fixes https://github.com/aiven/devportal/issues/938


laysauchoa (approved):
lgtm, good job!

Issue #658, opened 2022-03-21
Getting started with MySQLx


laysauchoa:
https://help.aiven.io/en/articles/5131813-getting-started-with-mysqlx

laysauchoa:
@dragonlancewu @runwuf this seems like an interesting article to migrate because you can learn a bit of python client for mysql besides just the migration.

laysauchoa:
Running this example gives an error but going with the option2) fixes the error

```
ImportError                               Traceback (most recent call last)
Input In [2], in <module>
----> 1 import mysqlx

File ~/.pyenv/versions/3.8.0/envs/devportal/lib/python3.8/site-packages/mysqlx/__init__.py:43, in <module>
     40 except ImportError:
     41     JSONDecodeError = ValueError
---> 43 from .connection import Client, Session
     44 from .constants import (Auth, LockContention, OPENSSL_CS_NAMES, SSLMode,
     45                         TLS_VERSIONS, TLS_CIPHER_SUITES)
     46 from .crud import Schema, Collection, Table, View

File ~/.pyenv/versions/3.8.0/envs/devportal/lib/python3.8/site-packages/mysqlx/connection.py:79, in <module>
     76 # pylint: disable=W0622
     77 from .errors import (InterfaceError, NotSupportedError, OperationalError,
     78                      PoolError, ProgrammingError, TimeoutError)
---> 79 from .crud import Schema
     80 from .constants import SSLMode, Auth, COMPRESSION_ALGORITHMS
     81 from .helpers import escape, get_item_or_attr, iani_to_openssl_cs_name

File ~/.pyenv/versions/3.8.0/envs/devportal/lib/python3.8/site-packages/mysqlx/crud.py:39, in <module>
     37 from .errors import NotSupportedError, OperationalError, ProgrammingError
     38 from .helpers import deprecated, escape, quote_identifier
---> 39 from .statement import (FindStatement, AddStatement, RemoveStatement,
     40                         ModifyStatement, SelectStatement, InsertStatement,
     41                         DeleteStatement, UpdateStatement,
     42                         CreateCollectionIndexStatement)
     45 _COUNT_VIEWS_QUERY = ("SELECT COUNT(*) FROM information_schema.views "
     46                       "WHERE table_schema = '{0}' AND table_name = '{1}'")
     47 _COUNT_TABLES_QUERY = ("SELECT COUNT(*) FROM information_schema.tables "
     48                        "WHERE table_schema = '{0}' AND table_name = '{1}'")

[...]

PR #1120, opened 2022-07-06
MySQLx/Python: add sample code

laysauchoa:
Add sample code for MySQLx code
using Python.

Fixes https://github.com/aiven/devportal/issues/658


docs/products/mysql/howto/connect-from-mysqlx.rst
```diff
@@ -0,0 +1,59 @@
+Connect to MySQL using MySQLx with Python
+=========================================
+
+Enabling MySQLx protocol support allows customers to use their MySQL instance as a document store. This example shows how to connect to your Aiven for MySQL instance using MySQLx protocol.
```
harshini-rangaswamy, line 4:
```suggestion
Enabling MySQLx protocol support allows customers to use their MySQL instance as a document store. This example shows how to connect to your Aiven for MySQL instance using MySQLx protocol. This is test. 
```

harshini-rangaswamy (commented):
All looks good.

docs/products/mysql/howto/connect-from-mysqlx.rst
```diff
@@ -0,0 +1,59 @@
+Connect to MySQL using MySQLx with Python
```
anelook, line 1: The title says "to", but the file name is "from", let's make it same

docs/products/mysql/howto/connect-from-mysqlx.rst
```diff
@@ -0,0 +1,59 @@
...
+  .. code:: shell
+
+    export PROTOCOL_BUFFERS_PYTHON_IMPLEMENTATION=python
+
+Code
+''''
+
+Add the following to ``main.py`` and replace the placeholders with values for your project:
+
+.. literalinclude:: /code/products/mysql/connect.py
```
anelook, line 46: are you sure that you're including the right code? This is different from what is added as part of this PR

docs/products/mysql/howto/connect-from-mysqlx.rst
```diff
@@ -0,0 +1,59 @@
...
+Prerequisites
+'''''''''''''
+
+* Python 3.7 or later
+  
+* A ``mysqlx`` python library installed::
+     
+     pip install mysql-connector-python
+
+* An Aiven account with a service running. Click on **MySQLx** tab under **Connection information**, and copy ``SERVICE_URI`` and ``Port`` information.
```
anelook, line 33: I think it make sense to add PORT as additional variable as well. and specify that it is port for mysqlx, since it is different from default one

code/products/mysql/connect-mysqlx.py
```diff
@@ -0,0 +1,27 @@
+import mysqlx
+
+MYSQLX_PASSWORD = "user-password"
+SERVICE_URI = "aiven-service-URI"
+
+connection_data = (
+    f"mysqlx://avnadmin:{MYSQLX_PASSWORD}@{SERVICE_URI}/defaultdb?ssl-mode=REQUIRED"
```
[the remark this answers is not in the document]
anelook: Does SERVICE_URI include port here? I put another comment below, but I think it is worth adding MYSQLX_PORT separately as a variable to avoid double meaning and confusion.

anelook (changes requested):
Nicely organised and useful article! 

I found a couple of issues, so didn't test it, since the referenced code is different than in PR. Wanted to clarify that before testing.

code/products/mysql/connect-mysqlx.py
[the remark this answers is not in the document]
laysauchoa: The SERVICE_URI for MySQL service already includes the port. I believe it would be confusing for the user if referred with the port. Because in the UI, you get already in this format. For example:

`mysql://avnadmin:whatever@mysql-name-dev-project.aivencloud.com:PORT/defaultdb?ssl-mode=REQUIRED`

docs/products/mysql/howto/connect-from-mysqlx.rst
laysauchoa: I'm not sure this is correct or not, maybe @TibsAtWork can help.

docs/products/mysql/howto/connect-from-mysqlx.rst
laysauchoa: I'm not sure because when you test, you will see that the `SERVICE_URI` already includes the `PORT`. I think is very confusing to rename things different of the UI and do the split of the string ourselves etc.

docs/products/mysql/howto/connect-from-mysqlx.rst
laysauchoa: oh, no! this is wrong, thanks, I corrrected

code/products/mysql/connect-mysqlx.py
```diff
@@ -0,0 +1,26 @@
+import mysqlx
+
+MYSQLX_PASSWORD = "user-password"
+SERVICE_URI = "aiven-service-URI"
+MYSQLX_USER = "mysqlx-user"
+
+connection_data = f"mysqlx://{MYSQLX_USER}:{MYSQLX_PASSWORD}@{SERVICE_URI}/defaultdb?ssl-mode=REQUIRED"
```
laysauchoa, line 7: @anelook I added the user here as you said it could be different, so it makes sense. Thank you for the suggestion!

docs/products/mysql/howto/connect-from-mysqlx.rst
laysauchoa: I think I got it now, you mean file names. Replaced :)

docs/products/mysql/howto/connect-using-mysqlx-with-python.rst
```diff
@@ -0,0 +1,61 @@
...
+Enabling MySQLx protocol support allows customers to use their MySQL instance as a document store. This example shows how to connect to your Aiven for MySQL instance using MySQLx protocol.
+
+.. note::
+
+    MySQL initially provided support for X-DevAPI (MySQLx) in v5.7.12 as an optional extension that can be installed by the user. On the MySQL v8.0+, the X-DevAPI is supported by default.
+
+Variables
+'''''''''
+
+.. list-table::
```
anelook, line 13: This table didn't render properly

docs/products/mysql/howto/connect-using-mysqlx-with-python.rst
```diff
@@ -0,0 +1,61 @@
...
+   :widths: 20 80
+   :header-rows: 1
+
+   * - Variable
+     - Description
+   * - ``SERVICE_URI``
+     - Connection information about your Aiven for MySQLx service.
+   * - ``MYSQLX_PASSWORD``
+     - Password for ``avnadmin`` user.
```
anelook, line 22: I'd remove "avnadmin", since the name of the user can be different. Can be "Password for the user"

docs/products/mysql/howto/connect-using-mysqlx-with-python.rst
```diff
@@ -0,0 +1,61 @@
...
+'''''''''
+
+.. list-table::
+   :widths: 20 80
+   :header-rows: 1
+
+   * - Variable
+     - Description
+   * - ``SERVICE_URI``
+     - Connection information about your Aiven for MySQLx service.
```
anelook, line 20: Maybe we can be more specific - it is "URI for the connection"

docs/products/mysql/howto/connect-using-mysqlx-with-python.rst
```diff
@@ -0,0 +1,61 @@
...
+Prerequisites
+'''''''''''''
+
+* Python 3.7 or later
+  
+* A ``mysqlx`` python library installed::
+     
+     pip install mysql-connector-python
+
+* An Aiven account with a service running. Click on **MySQLx** tab under **Connection information**, and copy ``SERVICE_URI`` and ``Port`` information.
```
anelook, line 35: if SERVICE_URI already includes port then copying port separately not needed?

code/products/mysql/connect-using-mysqlx.py
```diff
@@ -0,0 +1,26 @@
+import mysqlx
+
+MYSQLX_PASSWORD = "user-password"
```
anelook, line 3: Because we explain variables in Variable section, I think duplicating is not necessary these three parameters at the start. It doesn't work as part of the code, right?

anelook (changes requested):
Run the code snippet and all went smooth! However, I found a couple of small issues. But I think we're almost there!

docs/products/mysql/howto/connect-using-mysqlx-with-python.rst
anelook: I think better to put user  before the password, then it is more obvious to which user the password belongs

laysauchoa: ok, done

docs/products/mysql/howto/connect-using-mysqlx-with-python.rst
laysauchoa: true, removed

code/products/mysql/connect-using-mysqlx.py
laysauchoa: this would not be a valid code. However, I added a comment so the user can replace it with their variables.

code/products/mysql/connect-using-mysqlx.py
```diff
@@ -0,0 +1,28 @@
+import mysqlx
+
+# replace the variables accordingly
+SERVICE_URI = "aiven-service-URI"
```
anelook, line 4: I'm not sure why wee keep these. If used with the variables this won't work since we also put data into the brackets. I think either these three lines should be removed or brackets below.

anelook: If we decide to keep there three lines, we need to specify that aiven-service-URI is URI for mysqlx

docs/products/mysql/howto/connect-using-mysqlx-with-python.rst
```diff
@@ -0,0 +1,61 @@
+Connect to MySQL using MySQLx with Python
+=========================================
+
+Enabling MySQLx protocol support allows customers to use their MySQL instance as a document store. This example shows how to connect to your Aiven for MySQL instance using MySQLx protocol.
+
+.. note::
+
+    MySQL initially provided support for X-DevAPI (MySQLx) in v5.7.12 as an optional extension that can be installed by the user. On the MySQL v8.0+, the X-DevAPI is supported by default.
+
+Variables
```
anelook, line 10: I'm a bit concerned that we put variables before we explain that they should be taken from Mysqlx tab. Should variables section go below Prerequisites?

also in Prerequisits you mention to copy SERVICE_URI, (and we removed the port), but USER, password should be copied as well. maybe it will be more clear if you use

"An Aiven account with a service running. Click on MySQLx tab under Connection information, and copy service URI, username and password needed to access the service with the help of MySQLx. They are used below as variables"  and move variables after that section

anelook (changes requested):
Looking at the information flow I have some concerns that we mislead the user and they will copy wrong data. I added couple of comments to avoid that.

code/products/mysql/connect-using-mysqlx.py
laysauchoa: solved, i removed it :)

docs/products/mysql/howto/connect-using-mysqlx-with-python.rst
laysauchoa: I understand, how about create a separate Issue that maybe we can work on that to changes the order between variables and prerequisites in similar cases like the Java and Python code for this section? I think is better to choose a general format

laysauchoa: I can also change here in a singular file, but just want to make sure would be okay to not have a standard through all the code samples like a best option since they show up in a summary way

anelook (changes requested):
I see what you mean and that in other mysql examples  we have first Variables and then Prerequisites.Let's keep it here same, but I think it is worth to add more clear information where the properties should be taken.

Right now If I follow the instructions section by section, I'd do it wrong because I'd take the general URI, user name and password. Later I'd realise that, but it will require extra action to fix it.

I think it is important to specify that those variables are part of mysqlx connection. Maybe put extra line explaining where those must be taken. Or within their description add that those are mysqlx related.

docs/products/mysql/howto/connect-using-mysqlx-with-python.rst
```diff
@@ -0,0 +1,61 @@
...
+Prerequisites
+'''''''''''''
+
+* Python 3.7 or later
+  
+* A ``mysqlx`` python library installed::
+     
+     pip install mysql-connector-python
+
+* An Aiven account with a service running. Click on MySQLx tab under **Connection information**, and copy **SERVICE_URI**, **User** and **Password** needed to access the service with the help of MySQLx. They are used below as variables.
```
anelook, line 35:
```suggestion
* An Aiven account with a service running. Click on MySQLx tab under **Connection information**, and copy **Service URI**, **User** and **Password** needed to access the service with the help of MySQLx. They are used below as variables.
```

anelook:
Sorry for so many comments! :)

docs/products/mysql/howto/connect-using-mysqlx-with-python.rst
lornajane: This is a good point, I put the explanation of how to get the variables in the table with the variables, rather than here in the pre-requisites section (we don't normally cover the variables again here, I don't think)

docs/products/mysql/howto/connect-using-mysqlx-with-python.rst
lornajane: Looks good!

lornajane (commented):
Pushed a few minor improvements, but it was already in good shape! Thanks @laysauchoa for the code and @anelook for the reviews!

lornajane (approved):

lornajane:
mmmm, I approved but the code doesn't actually run for me. I'll double-check this

lornajane:
Seems to work for others, so I'll merge and let's keep an eye on it.

laysauchoa:
what is the error message when you run the code @lornajane? let me know if we can help

lornajane:
It was hanging on connection, then when I tried it later it was fine ....? I have no idea what I did wrong.

Issue #1123, opened 2022-07-06
Add Security policy


laysauchoa:
If someone finds some security vulnerability, how can they report? It would be nice to have short steps on a `Security policy` for this repository.

lornajane:
Check with the open source team what the usual practice is for the open source repositories; we should follow existing practice on this I think.

PR #1125, opened 2022-07-07
Add security policy

laysauchoa:
If someone finds some security vulnerability, how can they report? 
It would be nice to have short steps on a Security policy for this repository.

Fixes: https://github.com/aiven/devportal/issues/1123

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,34 @@
+# Security Policy
+
+## Supported Versions
+
+We release patches for security vulnerabilities. Which versions are eligible
+receiving such patches depend on the CVSS v3.0 Rating:
+
+| CVSS v3.0 | Supported Versions                        |
+| --------- | ----------------------------------------- |
+| 4.0-10.0  | Most recent release                       |
+
+## Reporting a Vulnerability
+
+Please report (suspected) security vulnerabilities to our **[bug bounty
+program](https://hackerone.com/aiven_ltd)**. You will receive a response from
+us within 2 working days. If the issue is confirmed, we will release a patch as
+soon as possible depending on impact and complexity.
+
+## Qualifying Vulnerabilities
+
+Any reproducible vulnerability that has a severe effect on the security or
+privacy of our users is likely to be in scope for the program.
+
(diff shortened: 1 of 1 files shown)


anelook:
This is nice I have a couple of notes: 
- from https://hackerone.com/aiven_ltd I see that it is for services, but it is not completely obvious from the text if it is for DevPortal or for Aiven services. 

- should this be  SECURITY.md  visible only to contributors of dev portal, or more noticeable article inside dev portal? 
Security audience interested in breaking Aiven services is most probably different from those who are making documentation contributions.

laysauchoa:
> - from https://hackerone.com/aiven_ltd I see that it is for services, but it is not completely obvious from the text if it is for DevPortal or for Aiven services.

I believe we can make small adjusts in the text but I was recommended to use the general policy for Aiven.

> - should this be SECURITY.md visible only to contributors of dev portal, or more noticeable article inside dev portal?

In my opinion, this is the right place to add because it will link on GitHub to our "Security Policy", and this is the standard for OSS projects. 

> - Security audience interested in breaking Aiven services is most probably different from those who are making documentation contributions.

So the security policy allows a safe way to contact the maintainer to disclose vulnerability, so it is not disclosed in public as it could affect the people who use the code. We have a lot of examples here and if for some reason an example can be harmful somehow, then we should have a way to be contacted about it. So we can take the right actions after.

anelook:
I see, thanks for explaining, now I understand better. Yes, in this case the location totally makes sense.
Should https://hackerone.com/aiven_ltd?type=team updated to clearly include developer portal resources?

laysauchoa:
Hi @jlprat do you think we can include this there? or how this can be done? alternatively, if the devportal can not be included there, how should the reader proceed if some vulnerability is on the code? thanks in advanced

jlprat:
I'd involve @st3fan because in the end this is a security policy.
@st3fan what do you think?

lornajane:
@jlprat  @st3fan do you have concerns that should block this merge? I'd like to go ahead, even if we amend later

jlprat (approved):

PR #1131, opened 2022-07-08
warn about whitespace in ip_filter

paulgrainger-aiven:

If whitespace is added between values in the ip_filter it can lead to a confusing error. Calling it out here will make it clear to omit any whitespace between values

### DIFF
--- a/docs/tools/cli/service.rst
+++ b/docs/tools/cli/service.rst
@@ -650,12 +650,14 @@ Updates the settings for an Aiven service.
     avn service update big-service        \
       --plan business-4     
 
-**Example:** Update the service named ``secure-database`` to only accept connections from the range ``10.0.1.0/24`` and the IP ``10.25.10.12``. Note that there is no whitespace between the IP addresses in the command below.
+**Example:** Update the service named ``secure-database`` to only accept connections from the range ``10.0.1.0/24`` and the IP ``10.25.10.12``.
 
 ::
 
     avn service update secure-database \
       -c ip_filter=10.0.1.0/24,10.25.10.1/32
+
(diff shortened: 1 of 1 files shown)


docs/tools/cli/service.rst
```diff
@@ -650,7 +650,7 @@ Updates the settings for an Aiven service.
     avn service update big-service        \
       --plan business-4     
 
-**Example:** Update the service named ``secure-database`` to only accept connections from the range ``10.0.1.0/24`` and the IP ``10.25.10.12``.
+**Example:** Update the service named ``secure-database`` to only accept connections from the range ``10.0.1.0/24`` and the IP ``10.25.10.12``. Note that there is no whitespace between the IP addresses in the command below.
```
anelook, line 653: To make notes and warnings more visible, we use special small sections designed for it.
You can use `.. note::` for this. For example in this case if you move the note under the command line block it can be:

```
.. note::

There is no whitespace between the IP addresses in the command above.
```

It would be great if we align notes as they are used in other places. Could you do this?

anelook (dismissed):
Thank you @paulgrainger-aiven for adding this clarification! I have one small comment related to aligning the style with other places where we use notes.

docs/tools/cli/service.rst
lornajane: I applied this change myself to keep things moving.

lornajane:
Rebased to pick up a fixed link that caused a build failure.

lornajane:
Rebased again, sorry, I think there's another failing link to pick up.

lornajane (approved):
Sometimes the smallest patches make the biggest difference! Thanks @paulgrainger-aiven

PR #1133, opened 2022-07-08
Clarify backup storage region for OpenSearch

lornajane:

After receiving some questions about backups for OpenSearch, this adds clarification that the backup will be stored in the same region as the service.


anelook (approved):
Looks good to me! Thank you for adding clarification

PR #1136, opened 2022-07-10
Fix links that got broken after the changes in OpenSearch documentation

anelook:

For several last days the link check fails for two OpenSearch links. I replaced one of the link and removed another one.


laysauchoa (approved):
thanks for fixing it!

PR #1142, opened 2022-07-13
cassandra: add help article on tombstones

allhailwesttexas:
Add a new help article on Cassandra tombstones, so that we have somewhere to direct users who receive a user alert about encountering too many tombstones during queries. Currently, this article is pretty short; it could definitely go into more detail, but the overarching advice of "change your queries or change your data model" is there, which is the most important bit for someone coming there from a user alert.

Also flesh out the Cassandra section a bit - at the moment there is only a reference document, so add a section for concepts as well. 

This PR also includes a small commit for ignoring the `.venv` folder in the `.gitignore` and when looking for docs to build.

### DIFF
--- a/docs/products/cassandra/concepts/tombstones.rst
+++ b/docs/products/cassandra/concepts/tombstones.rst
@@ -1,0 +1,66 @@
+Tombstones in Apache Cassandra®
+===============================
+
+Apache Cassandra manages deletion of data via a mechanism called *tombstones*. Because Cassandra is a distributed system,
+it cannot delete data immediately in the same way as a traditional relational database. On a high-level, when a row is
+deleted, instead of immediately deleting it, Cassandra will mark it as a tombstone row. Then, as part of regularly scheduled
+maintenance, the row will actually get deleted. This maintenance is called *compaction* and the threshold is controlled by
+the table-level setting ``gc_grace_seconds``. Any tombstone older than this setting will be removed completely during
+compaction (with some caveats - more details in the Cassandra `documentation <compaction-tombstones_>`_ on compaction). The
+default value for this setting is 864000 seconds (10 days).
+
+Problems caused by tombstones
+-----------------------------
+
+Excessive amounts of tombstone rows can lead to various problems since the data is still there on disk, it's just got a
+tombstone marker attached to it. If your workload includes a lot of data deletion, you are more likely to face problems.
+Tombstones also result in a lot of garbage collection, which can affect cluster stability. The two main things affected are
+read performance and disk usage.
+
+**Read performance**: if read queries have to scan large numbers of tombstones, the query performance can be significantly
+degraded. In particularly bad cases, the query can even time out. There are a couple of types of queries that are more likely
+to be affected by this. They all involve scanning all or a large part of a table.
+
+* Full table scans like ``SELECT * from inventory.items``
+* Any query that requires adding ``ALLOW FILTERING``
+* Range queries, i.e. queries with ``WHERE item_cost > threshold`` or similar
+
+**Disk usage**: if you are rapidly filling up your cluster with data at the same time as you are doing a lot of
+deletions, you will reach size limits sooner. This is because the tombstone data is not actually deleted and still taking up
+space on disk.
+
+
+If you suspect problems caused by tombstones for your cluster, you can check the logs. By default, if a query encounters over
+1000 tombstones (configured by `tombstone_warn_threshold <cassandra-tombstone-warn_>`_) it will generate a log entry. The
+entry will be in the format ``Read <X> live rows and <Y> tombstone cells for query <query> [...] (see tombstone_warn_threshold)``.
+
+If it encounters over 100 000 tombstones (configured by ``tombstone_failure_threshold)``, the query will be aborted with a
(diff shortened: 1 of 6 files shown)


docs/products/cassandra/concepts/tombstones.rst
```diff
@@ -0,0 +1,61 @@
+Tombstones in Apache Cassandra®
+===============================
+
+Apache Cassandra manages deletion of data via a mechanism called *tombstones*. Because Cassandra is a distributed system,
+it cannot delete data immediately in the same way as a traditional relational database. On a high-level, when a row is
+deleted, instead of immediately deleting it, Cassandra will mark it as a tombstone row. Then, as part of regularly scheduled
+maintenance, the row will actually get deleted. This maintenance is called *compaction* and the threshold is controlled by
+the table-level setting ``gc_grace_seconds``. Any tombstone older than this setting will be removed completely during
+compaction (with some caveats - more details in the Cassandra `documentation <compaction-tombstones_>`_ on compaction). By
+default, this value is 864000 seconds (10 days), which means that a single node can be unavailable for that long before you
```
rgarth, line 10: Can you expand a little on what resurrected data means. And what potential problems there are. It feels like a good reason to not lower the threshold too low.

allhailwesttexas: I've tried to make this a bit clearer, but could also just remove it.

lornajane:
Tagged as helpwanted since I think we'll need some help with the review on this

docs/products/cassandra/concepts/tombstones.rst
fmanimali: I think we are getting a little too deep in the cassandra internals for what is the purpose of our article. It is true that resurrected data is a big problem and in some cases it can go undetected for long time. It is worth a mention and a link to https://thelastpickle.com/blog/2016/07/27/about-deletes-and-tombstones.html.

fmanimali:
Other things that can help with deleting data without creating the resurrected data problem is to use TTL at table level or at update record level. TTL is a lot more efficient when it comes to purging deleted data because all nodes have the knowledge of the tombstones since the data was inserted.
One other thing is to choose the compaction strategy correctly: TimeWindowCompactionStrategy with data TTL will drop entire sstable files when they expire, using a background cassandra thread, without compaction intervention.
But again, maybe these things can just be mentioned because getting deeper into the data modeling details takes too long.

Couple of things at the end:
This line:
> Change your data model so that you don’t have to delete so much data.
change to:
> Review your data model and compaction strategy and implement table TTL or TimeWindowCompactionStrategy if possible 

this line:
> If you are planning to delete all the data in a table, you can drop the whole table to avoid creating tombstones.
change to:
> If you are planning to delete all the data in a table, you can truncate the table to avoid creating tombstones.


Your last sentence: "the tombstones will be removed automatically." should be changed into "the tombstoned data will be removed from disk"

docs/products/cassandra/concepts/tombstones.rst
```diff
@@ -0,0 +1,62 @@
...
+* Update your queries so that they don't have to scan over tombstone rows in the same manner. For range queries, this might
+  mean investigating if you can use a narrower range, or use a different approach to the query.
+* If you are planning to delete all the data in a table, you can drop the whole table to avoid creating tombstones.
+
+Removing existing tombstones
+----------------------------
+
+The best is to wait for it to happen automatically as part of regular operations. Once more time than ``gc_grace_seconds``
+has elapsed and a compaction happens, the tombstones will be removed automatically.
+
```
mparikhaiven, line 60: Shall we mention about a couple of commands (as a last resort with CAUTION) to be used in case of urgent need to remove expired tombstones?

1. `nodetool compact` : Forceful compaction of SSTables, it demands large free disk space and hence should be used on a limited set of selective tables at a time. If this is run on a table with STCS compaction strategy, it may cause a large SSTable.

2. `nodetool garbagecollect` : It runs a series of small compactions that also checks overlapping SSTables. While it uses less amount of disk space in the process, it is more CPU intensive and longer process execution than the `nodetool compact` option

docs/products/cassandra/concepts/tombstones.rst
allhailwesttexas: In the end I just removed this section since it was hard to make a statement that was both succinct and correct, and it's not central to the purpose of the article.

docs/products/cassandra/concepts/tombstones.rst
allhailwesttexas: These are good points, but there is no way for the customer to run `nodetool` commands against their Aiven service, so I deliberately any mentions out here.

allhailwesttexas:
@fmanimali Thank you for reviewing! I've added some mentions to TTL and TWCS with links to the docs + done the suggested sentence changes. Please have another look when you when you have time.

docs/products/cassandra/concepts/tombstones.rst
mparikhaiven: thnx for clarifying @allhailwesttexas

mparikhaiven (approved):

lornajane:
Thanks everyone! I really appreciate the Cassandra knowledge in the original article and in the excellent and knowledgeable reviews. You all rock!!!!

PR #1145, opened 2022-07-14
Add first set of Snowplow data collection

lornajane:

Adds anonymous tracking for page views only. As part of our wider data strategy, this should replace GA eventually.


lornajane:
Tested that the preview sends data to bigquery as expected, not sure there's much another reviewer can do here so I'm merging, totally breaking all my own rules!!

PR #1147, opened 2022-07-15
Remove pgaudit from supported extensions

kmichel-aiven:

The extension is not publicly available currently.


lornajane (approved):
Thanks for the patch!

PR #1161, opened 2022-07-20
Reorganised Kafka Connect navigation

### LINKED ISSUE #1160: Restructure Kafka Connect connectors navigation

(more information is useful here! Were you trying to do something particular?)

Source and sink connector navigation is, as of now, just a long list. It needs to be organised per source/target technology and cleaned a bit.

[X] I plan to open a pull request for this

ftisiot:

Reorganised Kafka Connect Navigation per technology

Fixes #1160

### DIFF
--- a/_toc.yml
+++ b/_toc.yml
@@ -275,32 +275,64 @@ entries:
             - file: docs/products/kafka/kafka-connect/howto/enable-automatic-restart
           - file: docs/products/kafka/kafka-connect/howto/list-source-connectors
             entries:
-            - file: docs/products/kafka/kafka-connect/howto/jdbc-source-connector-pg
-            - file: docs/products/kafka/kafka-connect/howto/jdbc-source-connector-mysql
-            - file: docs/products/kafka/kafka-connect/howto/jdbc-source-connector-sql-server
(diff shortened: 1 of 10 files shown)


lornajane:
I'm not at all sure about this re-organisation. Could we possibly spend some time with all the available options in a list and look at how to categorise them? My feeling is that I might want to source and sink on the same tech sometimes. Also that it's fine to have them all in one big bucket with a better landing page (coming after the theme upgrade and panels plugin upgrade) rather than relying on the nav for discoverability.

ftisiot:
Thanks for the comment! 
The organisation so far has been by dividing only source and sink connectors and I believe from a reader/user point of view it makes sense (either you want to pull data or push). 
This reorganisation was aimed at providing a better grouping of the connectors by technology. The alternative would be, as of now, to just keep the long lists, possibly simplifying the titles in the `_toc` file, and then manage a better navigation in the landing page. 
WDYT?

lornajane:
I'm closing this but let's keep the conversation going on the issue because I agree we need to make changes, I just don't think we've nailed the information architecture aspect of it.

PR #1170, opened 2022-07-21
Update search indexing for new templates

lornajane:

After updating to the new Furo theme, the change in page structure caused our search indexing to fail. This fixes that and the issues relating to it:
 - takes account of the new headerlink (it's a # rather than a paragraph symbol so we need to pick out the tag rather than parsing a symbol out of the text)
 - ignores the "back to top" link and the light/dark and menu item icons
 - does not index tables of contents (for example the lists on the "concept" pages)

Use the SEARCH instructions to create an opensearch and test what gets indexed.

### DIFF
--- a/scripts/index_developer_portal_pages.py
+++ b/scripts/index_developer_portal_pages.py
@@ -49,7 +49,7 @@ def parse_pages(html_build_dir):
             content = elements.text
             pages.append({
                 'title': title,
-                'content': content,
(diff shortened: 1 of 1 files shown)


kevinbowman-aiven:
LGTM, I've had it running locally and it indexes whereas current `main` does not.

kevinbowman-aiven (approved):

PR #1177, opened 2022-07-25
missing-cli-docs

ftisiot:

It's difficult to track which CLI docs we have and which are missing, so I wrote a little script to tell me that. 

Requires `os`, `aiven-client` and `docutils`

### DIFF
--- a/scripts/aiven/cli-missing-docs.py
+++ b/scripts/aiven/cli-missing-docs.py
@@ -1,5 +1,5 @@
-
 import os
+from aiven.client import argx
 from aiven.client import cli
 from docutils.core import publish_doctree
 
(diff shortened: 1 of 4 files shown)


scripts/aiven/cli-missing-docs.py
```diff
@@ -0,0 +1,72 @@
...
+                    titles = publish_doctree(text).traverse(condition=section_title)
+                    for t in titles:
+                        list_of_commands.append(t.astext())
+    # Remove from the titles the avn prefix to match it with the functions
+    list_of_commands = list(map(lambda x: x.replace('avn ', ''), list_of_commands))
+    
+    return sorted(list_of_commands)
+
+# This function parses the CLI and returns all the functions
+def find_cli_func():
```
TibsAtWork, line 34: This is pretty close, but includes some things we don't want.

The things we *do* want are annotated with `@arg` in the aiven_client source code, so it feels like we should be able to use that.

Looking at the `help` method in `aiven/client/cli.py`, it is doing:
```
321:               arg_list = getattr(prop, argx.ARG_LIST_PROP, None)
```
to get back "the stuff stored by `@arg`", and then:
```
 322:               if arg_list is not None:
```
to ignore the method if it does not have such "stuff".

(we can ignore the `patterns` code just thereafter, as it's part of allowing one to do `avn help pg`, which I didn't know one could do)

(An alternative approach would be to generate the `avn help` output, by calling  `cli.AivenCLI().run(['help'])`, but it outputs to `stdout` (that could be captured) and also includes the summary text (which would be a pain to remove), so I ruled that out.)

laysauchoa:
I've added some small changes but while debugging, I could not find the docs commands with the script.

TibsAtWork:
Yeh for the added doc strings (it's always better with docstrings 😄) and tidying up, thanks @laysauchoa

I think the diagnostic print messages can probably go away?

The script does seem to work for me?

laysauchoa:
ok, I believe as it is right now will not work if you run from this path: ``devportal/scripts/aiven`` this needs to be at ``devportal`` level to get results for the documents list (``ist_of_docs``). I've changed the path and now works both ways.

TibsAtWork:
Interesting. For me it did seem to be running in its own directory - but the change you've made is a good one anyway, to allow it to be run from anywhere.

TibsAtWork (approved):
Since this does what we want, and is for internal use only, I reckon this is now OK to merge.

laysauchoa:
I agree, we can always improve later. And it is already a great tool to have. 

Thanks, @ftisiot @TibsAtWork for your work!

PR #1180, opened 2022-07-25
Use os.walk, borrow way of finding commands

TibsAtWork:
1. Use `os.walk` as a more colloquial way of finding files
2. Borrow how to find the `avn` commands from the code in aiven_client itself


ftisiot (approved):
Looks good! 
Also seems to work!

Issue #1038, opened 2022-06-26
Calm down linkchecking


lornajane:
Rather than checking all the links on every build (there are a lot now!), it would be ideal if we could:
 * only check changed links
 * do a full link check on a regular (weekly? Like the other automated jobs?) basis so we do pick up if anything on an external site has decayed

laysauchoa:
Going over these points:

1. only check changed links
2. do a full link check on a regular (weekly? Like the other automated jobs?) basis so we do pick up if anything on an external site has decayed

Here are my thoughts:
1) Seems like the only way to run linkcheck is running in all the files. There was a tool at some point that could run on the differences but it is not supported on the latest versions. You can check the info here: https://github.com/sphinx-doc/sphinx/issues/10643#issuecomment-1177956493 

Do you have some ideas what we can try to solve this?

2) Yes, we can do this. Does this depends on point 1) to be done first? Otherwise, I can add a workflow for it.

laysauchoa:
Hi, @lornajane! Thanks for asking more details in how we can solve this here.

I figured out how to run this with GitHub actions. This will reduce the current build time for the checking the links.

-     only check changed links (https://github.com/aiven/devportal/pull/1181)
-     do a full link check on a regular (weekly? Like the other automated jobs?) basis so we do pick up if anything on an external site has decayed

Could you please clarify about the second point, should it be added to the same PR? Thank you!

lornajane:
Yes, let's add the full check as a weekly thing with the other Tuesday jobs, I think it makes sense to put it in the same PR so we keep checking everything until we put in the weekly job setup

PR #1181, opened 2022-07-26
Calm down linkchecking

laysauchoa:
This PR proposes to run linkcheck:

1) Only changed files
2) do a full link check on a regular (weekly? Like the other automated jobs?) basis so we do pick up if anything on an external site has decayed

**1) Only changed files**

Following this: https://github.com/sphinx-doc/sphinx/issues/10643#issuecomment-1187494777

This PR proposes to run linkcheck ONLY in changed files 📁. We can use a GitHub actions to check for changed files and the following command to  specify the files:
    
    ```
    sphinx-build -b linkcheck srcdir outdir path/to/target1.rst path/to/target2.rst ...
    ```

⌛ Time reduction examples for running:
~2 min to ~22 secs when you are not changing``.rst`` files 
~2min to ~50 secs (3 files changes)

**2) Do a full link check on a regular basis** 

The full linkcheck will still run every Tuesday. The schedule was chosen to be slightly earlier than other workflows, to avoid slowing down the running.

I've tested the workflow that runs on schedule here: https://github.com/laysauchoa/devportal-forked/blob/main/.github/workflows/test-file-linkcheck.yaml and it ran on schedule accordingly: https://github.com/laysauchoa/devportal-forked/actions/runs/2750692575

In any case, I can check next Tuesday if workflows runs on schedule once this PR is merged.

Fixes: https://github.com/aiven/devportal/issues/1038


Makefile
```diff
@@ -27,6 +27,10 @@ livehtml:
 livehtmlall:
 	sphinx-autobuild -a "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O) --watch _static
 
+linkcheck-changed-files:
```
TibsAtWork, line 30: I know this is nitpicking, but could we maybe call this target something more like `linkcheck-specified-files`, since if I do `make linkcheck-changed-files` at the command line, it still checks all the files?

laysauchoa: Hi @TibsAtWork this command can only check **.rst** files, but yeah, I track the changed files and run a link check.

laysauchoa: Honestly, you are much better than me with the naming. So yeah, I will change ^^

TibsAtWork (approved):
Thanks for doing this! It should speed things up a lot, and be much nicer for people when links fail in files they've not touched.

PR #1183, opened 2022-07-26
product m3db index.html update link to non-redirect link

angelinekwan:
Part of the SEO audit, it is recommended to link to a non-redirect link to minimize crawl budget and extra computing resources for web visitors.

DevPortal links that contains /index.html are now redirecting to parent folder. Updated link to support this new change.


lornajane (approved):
Thanks! I think we probably have a few of these but I should be able to check automatically.

PR #1189, opened 2022-07-28
Link Aiven videos to the docs

laysauchoa:
Link Aiven tutorial videos to the docs


lornajane:
Great idea! Could we include other DevEd videos also?

laysauchoa:
Hi @lornajane, we have only three and one is included already on CLI commands: https://developer.aiven.io/docs/tools/cli.html

lornajane (approved):
I should have known you already checked you had all the videos! Thanks for adding these

PR #1193, opened 2022-07-28
MySQL: fix parameters

laysauchoa:
Fix this database name parameter
because it could have different name.

``defaultdb`` is the default one created
by Aiven.

I'm fixing this. Thanks for the tip @anelook


ftisiot (approved):
LGTM!

PR #1200, opened 2022-08-01
Add Changelog link

laysauchoa:
This link is missing on the devportal.
Adding a mention on the platform landing page.

Fixes: https://github.com/aiven/devportal/issues/1201

### DIFF
--- a/docs/platform.rst
+++ b/docs/platform.rst
@@ -16,6 +16,10 @@ InfluxDB®,
 Grafana®,
 and M3 - all available in more than 80 regions around the world on AWS, GCP, Microsoft Azure, DigitalOcean, and UpCloud cloud platforms.
 
+What is the Aiven platform?
+----------------------------
+
+You can check latest Aiven improvements on the `Aiven Changelog <https://aiven.io/changelog>`_ page.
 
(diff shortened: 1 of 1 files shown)


docs/platform.rst
```diff
@@ -16,6 +16,10 @@ InfluxDB®,
 Grafana®,
 and M3 - all available in more than 80 regions around the world on AWS, GCP, Microsoft Azure, DigitalOcean, and UpCloud cloud platforms.
 
+What is the Aiven platform?
```
anelook, line 19: I think the title "What is the Aiven platform?" doesn't reflect  the content of the section.
Maybe we can use something like "Latest improvements" or "What's new" or something else reflecting the info on changelog. What do you think?

anelook (changes requested):
Thank you for adding it, this is a great idea! I'm just not sure about the title of the section and left a comment about it

docs/platform.rst
laysauchoa: I think I copied and did not change, my bad. Thanks for correcting it!

anelook (approved):

PR #1202, opened 2022-08-01
Add color flag to linkcheck 

laysauchoa:
In every push, we use linkcheck
to check on changed files. Currently,
there is no color on the logs.

This commit adds the color flag to enable
colors. This useful for the contributors to
easily see error messages, warnings etc in
a visual way.

Colors look like this:
Screenshot 2022-08-01 at 14 28 22


anelook (approved):
Thank you! this will be helpful! I'm not too familiar with sphinx-build linkcheck cli, but if there are any undesirable effects, this change is easy to revert.

PR #1203, opened 2022-08-02
Add corrections to ClickHouse commands

anelook:
The commands  were not wrong, however, I learned a better way to use the client, so added new changes.

### DIFF
--- a/docs/products/clickhouse/getting-started.rst
+++ b/docs/products/clickhouse/getting-started.rst
@@ -34,7 +34,7 @@ Connect to ClickHouse
 
 2. Go to the *Overview* tab and copy the **Host**, **Port**, **User**, and **Password** parameters that you need for connecting to the service.
 
-3. Run the following command to connect to your service and run SQL queries on your database, by substituing the placeholders for ``USERNAME``, ``PASSWORD``, ``HOST`` and ``PORT``:
+3. Run the following command to connect to your service and run SQL queries on your database, substitute the placeholders for ``USERNAME``, ``PASSWORD``, ``HOST`` and ``PORT``:
 
    .. code:: bash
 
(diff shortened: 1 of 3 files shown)


docs/products/clickhouse/getting-started.rst
```diff
@@ -34,19 +34,17 @@ Connect to ClickHouse
...
+3. Run the following command to connect to your service and run SQL queries on your database:
 
    .. code:: bash
 
-       docker run --interactive \
+       docker run -it \
        --rm clickhouse/clickhouse-client \
        --user USER-NAME \
        --password USER-PASSWORD \
        --host YOUR-HOST-NAME.aivencloud.com \
```
ftisiot, line 45: can we just name parameters `HOST` rather than `YOUR-HOST` I believe it's more consistent with other examples?

ftisiot (changes requested):
Hi Olena,
I would have the correct set of instructions directly here instead of forcing the user to flip between this article and the clickhouse one to find the parts to copy/paste. If the SQL is long, you can have the same approach as in the clickhouse docs to put it in one long line. Even more now that we're using a different approach compared to the main clickhouse docs (interactive vs one query at the time)

Furthermore, I got an error message when trying the first table creation, maybe you can replicate?

ftisiot:
Hi @anelook I made minor readability changes, and checked that everything seems to work. 
Can you double check everything still looks good on your end and then I can merge?

docs/products/clickhouse/getting-started.rst
anelook: Thank you for correcting these!

anelook:
@ftisiot Thanks a lot for corrections and alignment! Looks better than it was. And all looks correct to me.

Regarding of moving detail steps of how to use the dataset, I think we can keep it out of this commit and I plan to bring it to the DevRel sync, since I also have some doubts.

ftisiot (approved):
LGTM

Issue #813, opened 2022-05-13
Migrate article: SAML Authentication


lornajane:
Migrate the content from this article to fit the documentation portal format https://help.aiven.io/en/articles/3376377

(Assign this issue to yourself before you start, and remember to mention the issue number when you open your pull request)

encima:
Related Issues (subsets of this article):
#811 
#824 
#825 
#812 
#810 

It might be worth discussing the methodology for migrating these, for example:
1. Migrate all as they are and review later
2. Combine into a single article with a split for the differences
3. Have one "root" article with links to children (or text imported from the children)

ftisiot:
Close as it's solved in https://github.com/aiven/devportal/pull/1206

PR #1204, opened 2022-08-02
Migrated SAML overview article and added a reference table for common terms

encima:

Migrated SAML overview article and removed screenshots due to UI overhaul since they were taken.
Original IdP specific articles still in place and linked from this article.

Addresses https://github.com/aiven/devportal/issues/813

### DIFF
--- a/.github/vale/dicts/aiven.dic
+++ b/.github/vale/dicts/aiven.dic
@@ -14,6 +14,7 @@ boolean
 boot
 business
 Cassandra
+Centrify
 ClickHouse
 CloudWatch
 codecs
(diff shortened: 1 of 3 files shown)


ftisiot:
Hi @encima, isn't this article already contained in #1206?

encima:
it is contained yep. opened separately so that each issue has a PR attached to it. Technically, this one needs merging first. Ideally, these articles would be a series of PRs all leading from each other but it is non trivial (and likely overly complicated!)

ftisiot:
I believe we can close this and add both issues to the https://github.com/aiven/devportal/pull/1206

PR #1213, opened 2022-08-04
Add redis-cli article to list of howto connect articles on get started page

pparmin:
While looking at our get started page for Redis, I noticed that the article on how to connect to our service by using `redis-cli` was missing from the list of connect articles. 

This pull request adds the article to the list.


laysauchoa (approved):
Well spotted!!! Thank you

PR #1220, opened 2022-08-06
Fixed avn service integration-endpoint-types-list example.

dragonlancewu:
Fixed avn service integration-endpoint-types-list example.


anelook (approved):
Thank you for spotting this and fixing it

PR #1223, opened 2022-08-09
Reference blog post on finishing support of Aiven for Elasticsearch

TibsAtWork:
The `aiven-string-replacer-for-grafana` is referenced in an Aiven blog post, 
https://aiven.io/blog/finish-move-away-from-elasticsearch

More precisely, the blog post refers to this page.

It seems sensible to add a link back to the blog post.


anelook:
We don't do it for other blog posts and for this case I'm not sure how much value it provides for the reader. The blog post doesn't go into more details about replacing a string in Grafana dashboards, so this external link doesn't add a lot of extra info on the topic of the article.

TibsAtWork:
@anelook I think that's a good point, and I think I'm convinced this is not worth doing. I shall close this for now.

PR #1225, opened 2022-08-09
Clarify allowed Aiven for PostgreSQL extensions

mble:

Not all extensions listed were valid to install by normal service users
(e.g. `avnadmin`), as they require elevated privileged to install and
otherwise access functions and types available through those extensions.

Customers may consider reaching out to Aiven Support in regards to these
extensions, but a call to action is not included to prevent undue
Support burden.

This matters as customers may attempt to install extensions they are not allowed to.

### DIFF
--- a/docs/products/postgresql/reference/list-of-extensions.rst
+++ b/docs/products/postgresql/reference/list-of-extensions.rst
@@ -1,7 +1,9 @@
 Extensions on Aiven for PostgreSQL®
 ===================================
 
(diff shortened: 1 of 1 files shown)


anelook (approved):
Thank you for clarifying this!

laysauchoa:
@pparmin do we need to add a warning about this specific extension on this page? I understood that the user may not be able to use it, correct?

PR #1230, opened 2022-08-11
[concepts]: Out of Memory Conditions

brendan-stephens:
Rework: service-memory-limits to move the list of services
into an include file. This allows for multiple articles to reference the
same set with only 1 file to maintain.

Added: out-of-memory conditions to explain OOM behaviour
and how Aiven uses containers to prevent uncontrolled shutdowns.
This is also a replacement for:
https://help.aiven.io/en/articles/4509218-out-of-memory-killer

### DIFF
--- a/docs/platform/concepts/out-of-memory-conditions.rst
+++ b/docs/platform/concepts/out-of-memory-conditions.rst
@@ -13,7 +13,7 @@ However, if enough processes start using all their allocated memory simultaneous
 The Out of Memory Killer
 ------------------------
 
-The solution that the linux kernel employs is to invoke the ``Out of Memory Killer`` (or ``OOM Killer``), a Linux kernel process that reviews all running processes and kills one or more of them in order to free up system memory in order keep the system running.
+The solution that the Linux kernel employs is to invoke the ``Out of Memory Killer`` (or ``OOM Killer``), a Linux kernel process that reviews all running processes and kills one or more of them in order to free up system memory in order keep the system running.
 
 
 Which process will be killed?
--- a/docs/platform/concepts/out-of-memory-conditions.rst
+++ b/docs/platform/concepts/out-of-memory-conditions.rst
@@ -1,0 +1,61 @@
+Out of memory conditions
+========================
+
+**Many processes request more memory from the kernel then they will ever use or need.**
+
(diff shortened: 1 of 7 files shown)


docs/platform/concepts/out-of-memory-conditions.rst
```diff
@@ -0,0 +1,61 @@
...
+Which process will be killed?
+-----------------------------
+
+The ``OOM Killer`` selects process to kill based on a ``oom_score``; a calculation that balances the how much memory the process is using with how long the process has been running. 
+
+Processes that have been running for a long time are less likely to be killed. Subprocesses are summed with parent processes in terms of memory usage, so a process which forks a large number of subprocesses, but itself does not use a lot of memory, may still be killed.
+
+*In most instances, the hosted data service, or a child process, will have the highest memory footprint and be a prime candidate for termination when the OOM Killer sweeps the running processes.*
+
+Aiven's cloud data platform leverages kernel namespaces (or containers) to isolate processes from each other. Isolation has several benefits, including:
```
TibsAtWork, line 28: Is `contsainers` here another name for `kernel namespaces`? Is so, we can drop the `or` in front of the word `containers`

TibsAtWork:
This looks good to me - are the numbers in still correct?

docs/platform/concepts/out-of-memory-conditions.rst
brendan-stephens: The word “container” doesn’t mean anything super precise. Basically there are a few new Linux kernel features (“namespaces” and “cgroups”) that let you isolate processes from each other. When you use those features, you call it “containers”.

brendan-stephens:
There are several articles that reference overhead, each one providing a calculation, and if we change the overhead value (350 MB) to something else, there would end up being multiple documents to manage. 

The numbers are current, but in my next PR, I might move the calculations to an include, but I have to think of a good way to do that. It might mean moving from a math block to a code block.

docs/platform/concepts/out-of-memory-conditions.rst
TibsAtWork: Thanks - I thought I'd check. In that case the text is perfectly OK as it is.

TibsAtWork (approved):
Thanks very much for doing this.
Looks nice and clear to me.

PR #1239, opened 2022-08-12
Try specifying vale version

TibsAtWork:


TibsAtWork:
That doesn't work - ignoring

Issue #1235, opened 2022-08-12
Vale is complaining about elasticsearch name in image path


ftisiot:
# What's wrong?

Vale is complaining about an image path containing `elasticsearch` but it shouldn't check path names

<img width="1308" alt="image showing two warnings related to image path containing the word elasticsearch" src="https://user-images.githubusercontent.com/11272743/184312475-f9662cf0-928f-4583-a10c-56507044c43a.png">

# URL of affected page, and any other information

TibsAtWork:
When I run Vale 2.20.0 (the current version) on that file at the command line, I do not get any warnings.

However, if I run Vale 2.17.0, I do get the errors as shown. Foreshadowing: I also get them for 2.15.5

Our CI uses [vale-action](https://github.com/errata-ai/vale-action) to run vale. We specify
```
      uses: errata-ai/vale-action@master
```
to get the most recent version of vale-action.

The [vale-action Dockerfile](https://github.com/errata-ai/vale-action/blob/master/Dockerfile) specifies what version of vale to use:
```
FROM jdkato/vale:v2.15.5
```

It looks as if there are two things we might do to fix this:

1. Make an issue/PR for vale-action to update the version of vale.
2. Create a fork or vale-action that we can update ourselves.

(1) is a good idea, and such a request will probably be acted on fairly soon, but it doesn’t really solve the problem that we generally want the latest version of Vale, or at least a more recent one, and we don’t control vale-action’s updates.

(2) allows us to choose the version of Vale we want, but does require us to make sure that we don’t drift from the main vale-action

TibsAtWork:
Turns out if we use the `reviewdog` branch of vale-action, which is now the default, then we get the latest branch of vale, and also the option to specify a particular branch. And that's now the official branch to use, according to https://github.com/marketplace/actions/vale-linter
See https://github.com/aiven/devportal/pull/1241

TibsAtWork:
Note that this should also make work on https://github.com/aiven/devportal/issues/1130 easier.

PR #1241, opened 2022-08-12
Change our CI to use the latest version of Vale for linting

TibsAtWork:
Before, we were deliberately using the `master` branch of [vale-action](https://github.com/errata-ai/vale-action), as that *was* the best way to get a reasonably up-to-date version of Vale

However, the default branch of vale-action is now `reviewdog`, and *that* will automatically use the current latest version of vale, which is probably what we want. It also allows specifying a particular version of vale if we don't want the latest.

**Note** in fact the documentation at https://github.com/marketplace/actions/vale-linter documents using this branch of vale-action

Fixes https://github.com/aiven/devportal/issues/1235

Tested by specifying an earlier version of vale (it used it) and then allowing it to choose the latest (here).

Also, the logs provided for the `Linting / prose (pull_request)` CI check now contain useful information under `Vale`, including the version of vale that was used, and the vale output.

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -19,7 +19,7 @@ jobs:
       # 2020-08-12 The default branch of vale-action is `reviewdog`, which
       # defaults to using the latest version of `vale`
       # If we want a specific version, use `version:` in the `with` clause below
-      uses: errata-ai/vale-action
+      uses: errata-ai/vale-action@reviewdog
       with:
         files: '["index.rst", "docs"]'
       env:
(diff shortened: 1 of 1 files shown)


RyanSkraba (approved):
Very pragmatically, I checked the "Linting" action of this branch (no warning) with the Linting action on a different [PR](https://github.com/aiven/devportal/runs/7805441460?check_suite_focus=true#step:4:10) (2 warnings)

So LGTM

lornajane:
Having been burned a few times by Vale, we _might_ want to specify a version and bump it periodically, but let's figure that out if we do have a problem!

TibsAtWork:
I did think about that, but now I know what to do to specify a particular version, it seemed sensible to try the latest and change it if we need to.

PR #1247, opened 2022-08-15
Adjust description of parameter indices_recovery_max_bytes_per_sec for improved clarity

### LINKED ISSUE #1245: [Opensearch] Adjust description of parameter indices_recovery_max_bytes_per_sec for improved clarity
# What's wrong?
Currently, the explanation for the advanced parameter `indices_recovery_max_bytes_per_sec` states the following:

```
Limits total inbound and outbound recovery traffic for each node. Applies to both peer recoveries as well as snapshot recoveries (i.e., restores from a snapshot). Defaults to 40mb
```

In a [recent discussion](https://aiven-io.slack.com/archives/C01UBB57XFS/p1660556724628889?thread_ts=1660553548.605819&cid=C01UBB57XFS) related to a performance issue with a big Opensearch service, we realised that the description could be potentially misleading. In reality, the default is not 40 MB *in general* but rather 40 MB *per CPU*

I am therefore opening this issue to track this change. I will open a PR to fix this shortly.

The parameter is part of the advanced parameters for Opensearch: https://developer.aiven.io/docs/products/opensearch/reference/advanced-params.html

pparmin:
I added a clarification that the default for the advanced Opensearch parameter `indices_recovery_max_bytes_per_sec` is 40 MB *per CPU*.

Fixes #1245. 

This time I based this on the correct branch (see #1246).

### DIFF
--- a/includes/config-opensearch.rst
+++ b/includes/config-opensearch.rst
@@ -84,7 +84,7 @@
     **indices.queries.cache.size** Percentage value. Default is 10%. Maximum amount of heap used for query cache. This is an expert setting. Too low value will decrease query performance and increase performance for other operations; too high value will cause issues with other OpenSearch functionality.
 
   ``indices_recovery_max_bytes_per_sec`` => *integer*
-    **indices.recovery.max_bytes_per_sec** Limits total inbound and outbound recovery traffic for each node. Applies to both peer recoveries as well as snapshot recoveries (i.e., restores from a snapshot). Defaults to 40mb
+    **indices.recovery.max_bytes_per_sec** Limits total inbound and outbound recovery traffic for each node. Applies to both peer recoveries as well as snapshot recoveries (i.e., restores from a snapshot). Defaults to 40mb per CPU. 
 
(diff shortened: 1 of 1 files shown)


safa-topal (approved):

lornajane:
Thanks for the patch! However the documentation for params like this is autogenerated content, I think this needs to be fixed upstream in the core repo rather than edited here. It's generated by running: `make service-type-config-opensearch` if you want to try it locally - it takes its information from the service types API endpoint IIRC

pparmin:
@lornajane I had a quick look at the command and it calls the script in `./scripts/aiven/service_type_config.py`. This script then pulls all service types from `https://api.aiven.io/v1/service_types` and takes the appropriate information for the requested service type and writes it to the file, which means you remembered it correctly.

Do you happen to know how I can modify the description that comes from the endpoint?

lornajane:
It's in the source code for the API itself

lornajane:
Closing because we can't merge - it'll be overwritten next time the automation runs.

PR #1252, opened 2022-08-16
update python version to 3.8

angelinekwan:
Due to Netlify change in not supporting Xenial build image - DEPRECATION NOTICE: Builds using the Xenial build image will fail after November 15th, 2022. [more details](https://answers.netlify.com/t/please-read-end-of-support-for-xenial-build-image-everything-you-need-to-know/68239), we need to upgrade the Python version to default supported version 3.8.

Focal support default Python 3.8. [Documentation here](https://github.com/netlify/build-image/blob/focal/included_software.md)


lornajane (approved):
The build is green so I can't think of another way to test this!, I'm using 3.9 locally and seems fine; we should be moving off 3.7 anyway.

PR #1254, opened 2022-08-16
[mysql] [platform]: concepts and calculations

brendan-stephens:

To deal with several platform variables and calculations in multiple locations,
platform-variables was added as an include to several incuments.

This allows a single source for variables that may be subject to change.

```
.. |mysql_max_connections| replace:: :math:`{\times ⌊usable\_memory⌋ } + { extra\_connection }`
.. |mysql_connections_per_<4G| replace:: :math:`{75}`
```

While allowing multiple documents and locations to remain unaltered if the
values ever change.

The only downside is we loose a bit of blocking and some formatting due to
substitution. Eg. `.. math::` cannot be used. However, it will accept inline:

```
    :math:`{max\_connections} =` |mysql_connections_per_<4G| |mysql_max_connections|
```

Several documents have been reworked to accommodate this change.

The document max-number-of-connections, was moved from how-to to concepts.
Users cannot modify this value, so it's an explanation, rather than tuning.

To this we add the mysql tuning document, to help explain the difference between
max_connections and max_concurrency.


angelinekwan:
Leaving a comment here that the preview build failed in Netlify because the Python needs to be upgraded from 3.7 to 3.8 in this branch. This bump was made yesterday - more details on why [here](https://github.com/aiven/devportal/pull/1252)
Screenshot 2022-08-17 at 16 29 54

brendan-stephens:
Ty. I updated python for my devportal env. 
I generally pull and check requirements against installed libraries, but that was not visible.

TibsAtWork:
I have rebased to (a) fix the dictionary clash and (b) see if that makes the checks pass.

TibsAtWork:
I think these pages look like a useful resource (they're Aiven specific to some extent, but explain useful concepts in an understandable way), and the experiment of using "constants" is probably worthwhile. I'm not 100% sure we need to separate out the `services-memory-capped.rst` file now, but I understand why you've done it, and since it's done, we should keep it.

TibsAtWork (approved):
Thanks for all the work on this - it's both useful documentation and exploring a technique we weren't using, which should benefit us in the future.

PR #1262, opened 2022-08-17
Add a note regarding play connectivity.

dewan-ahmed:

Since the HTTPS link to **play** will not work with VPC and filtered networks, I'm adding a note/warning per @kmichel-aiven 
for services that are either restricted by IP addressed or are within a VPC without public access.


dewan-ahmed:
I'll be adding another commit to use relative path for the links.

anelook (approved):
Thank you @dewan-ahmed for adding the clarification, even though I knew it I missed it :)

Issue #1035, opened 2022-06-24
Terraform Recipe Request - Creating a PostgreSQL/MySQL read replica


dewan-ahmed:
# What should we add to our content collections?

Requesting a content per [this request](https://github.com/aiven/devportal/issues/656)

Source: https://help.aiven.io/en/articles/5721139-creating-a-postgresql-mysql-read-replica-with-aiven-terraform-provider

This will be a Terraform cookbook entry as well as a HELP migration. Please remember to do necessary steps. 

# Is this something you would work on yourself?

Yes, I'm taking on this issue.

PR #1264, opened 2022-08-17
PostgreSQL read replica recipe

dewan-ahmed:

This Terraform recipe is also a migration task for https://help.aiven.io/en/articles/5721139-creating-a-postgresql-mysql-read-replica-with-aiven-terraform-provider which resolves https://github.com/aiven/devportal/issues/1035. 

The same recipe serves the purpose for both MySQL and PostgreSQL per the original help article.

### DIFF
--- a/docs/tools/terraform/reference/cookbook/postgresql-read-replica-recipe.rst
+++ b/docs/tools/terraform/reference/cookbook/postgresql-read-replica-recipe.rst
@@ -42,9 +42,9 @@ The following sample Terraform script stands up the primary PostgreSQL service a
 
 .. code:: terraform
   
-  resource "aiven_pg" "demo-postgres-primary" {
+  resource "aiven_pg" "demo-postgresql-primary" {
     project                 = var.project_name
-    service_name            = "demo-postgres-primary"
+    service_name            = "demo-postgresql-primary"
     cloud_name              = "google-northamerica-northeast1"
     plan                    = "startup-4"
     maintenance_window_dow  = "sunday"
@@ -52,10 +52,10 @@ The following sample Terraform script stands up the primary PostgreSQL service a
     termination_protection  = false
   }
   
-  resource "aiven_pg" "demo-postgres-read-replica" {
+  resource "aiven_pg" "demo-postgresql-read-replica" {
     project                 = var.project_name
     cloud_name              = "google-northamerica-northeast1"
-    service_name            = "demo-postgres-read-replica"
+    service_name            = "demo-postgresql-read-replica"
     plan                    = "startup-4"
     maintenance_window_dow  = "sunday"
     maintenance_window_time = "10:00:00"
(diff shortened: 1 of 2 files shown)


docs/tools/terraform/reference/cookbook/postgresql-read-replica-recipe.rst
```diff
@@ -0,0 +1,105 @@
+Create a MySQL/PostgreSQL® read replica using Terraform
```
ftisiot, line 1: Isn't the title a bit misleading? This is talking about PostgreSQL, we can have a hint about being the same for Mysql but the title should reflect what is in there?

docs/tools/terraform/reference/cookbook/postgresql-read-replica-recipe.rst
```diff
@@ -0,0 +1,105 @@
...
+========================================================
+
+A MySQL/PostgreSQL® read replica can be used to offload read requests or analytics traffic from the primary instance. In this example, you'll create two Aiven for PostgreSQL® services - a primary service and the other one as its read replica.
```
ftisiot, line 4: Analytics traffic is a read request

```suggestion
A MySQL/PostgreSQL® read replica can be used to offload read requests like the analytics traffic from the primary instance. In this example, you'll create two Aiven for PostgreSQL® services - a primary service and the other one as its read replica. 
```

docs/tools/terraform/reference/cookbook/postgresql-read-replica-recipe.rst
ftisiot: Also the other titles don't have the "Create a" prefix

docs/tools/terraform/reference/cookbook/postgresql-read-replica-recipe.rst
```diff
@@ -0,0 +1,105 @@
...
+A MySQL/PostgreSQL® read replica can be used to offload read requests or analytics traffic from the primary instance. In this example, you'll create two Aiven for PostgreSQL® services - a primary service and the other one as its read replica. 
+Both services will be provisioned programmatically using the `Aiven Terraform Provider <https://registry.terraform.io/providers/aiven/aiven/latest/docs>`_. The same process can be followed for setting up MySQL read replica as well. 
+The following diagram shows the typical client interaction when a read replica is in place.
+
+.. mermaid::
+
+   flowchart LR
+      id6[Client]
+      
+      id5>WAL record]
```
ftisiot, line 13: This process with WALs is PG only, in the case of mysql the replica works differently (or at least with different technology)

ftisiot (changes requested):
Hi Dewan!
thanks for this and the recipe works! I have minor comments on the wording at the top... If we talk about PG, let's not talk about mysql! :D

docs/tools/terraform/reference/cookbook/postgresql-read-replica-recipe.rst
dewan-ahmed: Thanks Francesco. Valid point. I was using the same title from the Help article but will remove the MySQL reference and "Create a" prefix.

dewan-ahmed:
Thanks @ftisiot for the review. I've incorporated the review comments via https://github.com/aiven/devportal/pull/1264/commits/88eb95fe149229a725e51363918d23d3856ebfbf. The only line I kept is 
> The same process can be followed for setting up MySQL read replica as well.

ftisiot (approved):
LGTM

PR #1270, opened 2022-08-18
Small number of link fixes

lornajane:

I realised we had links pointing to help pages that have been migrated. This updates them to point to this site (I didn't replace with internal links, we have a separate item for that), and also fixes one internal broken reference.


ftisiot (approved):
LGTM

Issue #1269, opened 2022-08-18
Broken image on technical emails article


lornajane:
Follow up to #1115 which we seem to have merged with a missing image. @dragonlancewu do you still have that, can we add it back in please? Problem is on this page: https://docs.aiven.io/docs/platform/howto/technical-emails

dragonlancewu:
@lornajane I fixed it in this PR: https://github.com/aiven/devportal/pull/1272

PR #1272, opened 2022-08-19
Fix broken image for technical emails

dragonlancewu:
Fix broken image for technical emails how-to article per issue: https://github.com/aiven/devportal/issues/1269


ftisiot (approved):
Thanks a lot!

PR #1280, opened 2022-08-22
Instructions to connect CH with Grafana

anelook:

### DIFF
--- a/_toc.yml
+++ b/_toc.yml
@@ -498,6 +498,7 @@ entries:
         entries:
           - file: docs/products/clickhouse/howto/list-integrations
             entries:
+              - file: docs/products/clickhouse/howto/connect-to-grafana
               - file: docs/products/clickhouse/howto/integrate-kafka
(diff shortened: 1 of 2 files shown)


ftisiot (changes requested):
Hi @anelook, could you please check the 2 warnings? Also please rebase since I just merged @FloorD article about the Grafana plugins containing the fix for Altinity.
Also this is missing the entry in the _toc file, now that the Kafka integration is merged, maybe you can fix it?

ftisiot:
One note here, should this be under clickhouse, under grafana or under both?

anelook:
@ftisiot , I've rebased with latest main and added _toc entry. Fixed one warning (haven't seen the second one). The update from Floor is just into the list of all existing Grafana plugins, right? Not a separate article (I checked and don't see one)

I do believe this article belongs  to CH part and will be discovered by customers looking what integrations are available to visualise data.

ftisiot (approved):
LGTM

PR #1281, opened 2022-08-22
Adjust Snowplow actions based on cookie consent

vtainio:

Snowplow anonymization is now set based on the consent given by users
and correct cookies are in place if UX cookies are accepted.

### DIFF
--- a/_templates/base.html
+++ b/_templates/base.html
@@ -44,18 +44,35 @@ snowplow("newTracker", "at", "dc.aiven.io", {
   }
 });
 
-// enable fully anonymous tracking
-snowplow('clearUserData');
-
-snowplow("trackPageView");
-
 </script>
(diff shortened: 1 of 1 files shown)


carabaestlein (approved):
LGTM!

lornajane:
Thank you @vtainio ! I'll admit that I am confused though (I might need to test with @carabaestlein or something) - I can't see where this would track the page view when it's the box isn't still open, and it's not a first-time visit. What am I missing here?

lornajane (approved):
Looks a bit clearer this morning, approving.

vtainio:
@lornajane let me know if you have any further questions! It's a bit confusing that the same `OptanonWrapper` function gets called when accepting/denying and also on page load :/

Issue #1069, opened 2022-06-29
Migrate article: COVID-19: Coronavirus update


lornajane:
Migrate the content from this article to fit the documentation portal format https://help.aiven.io/en/articles/3792926

(Assign this issue to yourself before you start, and remember to mention the issue number when you open your pull request)

lornajane:
I think we have a few of these in the issues list, and I don't have a plan for them. I think it would be useful to add a tag and group them somehow so we can handle them all the same way when we decide.

sluhur:
For now I put the link under Reference. Let me know if this is not the right place

ftisiot:
This article shouldn't be moved since it's a one off reply valid last year but not needed anymore

PR #1295, opened 2022-08-25
Migrate article: COVID-19: Coronavirus update #1069

sluhur:
https://github.com/aiven/devportal/issues/1069


ftisiot:
Hi @sluhur, thanks for this. We had an internal chat and this article looks like a response given at a certain time that doesn't need to be ported

PR #1296, opened 2022-08-25
add path forwarding redirect from developer.aiven.io to docs.aiven.io

angelinekwan:
developer.aiven.io has been agreed to renamed to docs.aiven.io. We need to implement path forwarding redirect for developer.aiven.io URLs.

Based on the conversation at [OPS-1893](https://aiven.atlassian.net/browse/OPS-1893), since both domains are used as custom domains in Netlify, the most straightforward approach is best done in Netlify using their system of putting redirects in _redirects file. Referencing to domain-level redirects on [Redirect options](https://docs.netlify.com/routing/redirects/redirect-options/#domain-level-redirects)

Screenshot 2022-08-19 at 12 59 52

### DIFF
--- a/_redirects
+++ b/_redirects
@@ -1,3 +1,7 @@
+# Domain level redirect for path forwarding
+http://developer.aiven.io/*  https://docs.aiven.io/:splat 301!
+https://developer.aiven.io/* https://docs.aiven.io/:splat 301!
+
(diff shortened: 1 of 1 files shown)


lornajane (approved):
Thanks! Applying this now

Issue #1313, opened 2022-08-26
Terraform for  Cross-cluster replication with Apache Kafka MirrorMaker 2 wont work.


jorgeuribe-aiven:
# What's wrong?

Tried to create the Cross-cluster replication with Apache Kafka® MirrorMaker 2 using the Terraform code from the article, though it would not fail the replication does not work. 
There are wrong references when creating the 'aiven_mirrormaker_replication_flow'

# URL of affected page, and any other information

https://docs.aiven.io/docs/tools/terraform/reference/cookbook/kafka-mirrormaker-recipe.html#

PR #1318, opened 2022-08-29
fix-1313 kafka mm recipe won't work

jorgeuribe-aiven:
Fixes https://github.com/aiven/devportal/issues/1313

The Terraform code for [Cross-cluster replication with Apache Kafka® MirrorMaker 2](https://docs.aiven.io/docs/tools/terraform/reference/cookbook/kafka-mirrormaker-recipe.html#)
will create all the resources, there is no error reported from Terraform but Mirror Maker won't start the flow, issue is due to the source and target used in the aiven_mirrormaker_replication_flow, as per[ terraform docs](https://registry.terraform.io/providers/aiven/aiven/latest/docs/resources/mirrormaker_replication_flow) they should be the aliases from the service integration. 

I have tested the current code which does not complete the MM flow, and the fix in this PR works OK.


dewan-ahmed (approved):
Thanks so much @jorgeuribe-aiven for this fix.

PR #1322, opened 2022-08-30
Add redirects for renamed files

lornajane:

When we move or rename a file (which should be avoided if possible), it breaks the URL since the URL is made from the directory and file names. We do have a redirects setup to handle the special cases where we have moved something, however we have had some broken links and redirects reported when our SEO agency were doing some analysis on the docs. This pull request updates the reported URLs.

### DIFF
--- a/_redirects
+++ b/_redirects
@@ -38,7 +38,7 @@ https://developer.aiven.io/* https://docs.aiven.io/:splat 301!
 
 # Renamed/deleted files
 /docs/products/flink/howto/real-time-alerting-solution-cli.html       /docs/products/flink/howto/real-time-alerting-solution.html
-/docs/platform/howto/disable-user-2fa.html                            /docs/platform/howto/user-2fa.html
+/docs/platform/howto/disable-user-2fa.html                            /docs/platform/howto/user-2fa.html                                  301!
+/docs/tools/terraform/upgrade-provider-v1-v2.html                     /docs/tools/terraform/howto/upgrade-provider-v1-v2.html             301!
 /docs/products/mysql/howto/max-number-of-connections.html             /docs/products/mysql/concepts/max-number-of-connections.html
 /docs/products/postgresql/reference/plan-iops.html                    /docs/products/postgresql/reference/resource-capability.html
-/docs/tools/terraform/upgrade-provider-v1-v2                          /docs/tools/terraform/howto/upgrade-provider-v1-v2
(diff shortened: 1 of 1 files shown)


_redirects
```diff
@@ -39,3 +39,6 @@ https://developer.aiven.io/* https://docs.aiven.io/:splat 301!
 # Renamed/deleted files
 /docs/products/flink/howto/real-time-alerting-solution-cli.html       /docs/products/flink/howto/real-time-alerting-solution.html
 /docs/platform/howto/disable-user-2fa.html                            /docs/platform/howto/user-2fa.html
+/docs/products/mysql/howto/max-number-of-connections.html             /docs/products/mysql/concepts/max-number-of-connections.html
+/docs/products/postgresql/reference/plan-iops.html                    /docs/products/postgresql/reference/resource-capability.html
+/docs/tools/terraform/upgrade-provider-v1-v2                          /docs/tools/terraform/howto/upgrade-provider-v1-v2
```
angelinekwan, line 44: This redirect doesn't seems to work - notice in [preview](https://deploy-preview-1322--developer-aiven-io-preview.netlify.app/docs/tools/terraform/upgrade-provider-v1-v2). Will test something out in this PR

_redirects
```diff
@@ -38,4 +38,7 @@ https://developer.aiven.io/* https://docs.aiven.io/:splat 301!
 
 # Renamed/deleted files
 /docs/products/flink/howto/real-time-alerting-solution-cli.html       /docs/products/flink/howto/real-time-alerting-solution.html
-/docs/platform/howto/disable-user-2fa.html                            /docs/platform/howto/user-2fa.html
+/docs/platform/howto/disable-user-2fa.html                            /docs/platform/howto/user-2fa.html                                  301!
```
angelinekwan, line 41: Tried adding the force redirect but didn't resolve the redirect loop - can see in this [PR preview](https://deploy-preview-1322--developer-aiven-io-preview.netlify.app/docs/platform/howto/disable-user-2fa.html). Will open an Netlify support ticket. Similar case for `/docs/tools/terraform/upgrade-provider-v1-v2.html`

angelinekwan: @lornajane Netlify Support suggested to swap the order of the redirect rules above 
`# Redirect from .index.html to specific page names for landing`

And now the following 2 redirect loop are resolved and redirecting to the correct url.
https://deploy-preview-1322--developer-aiven-io-preview.netlify.app/docs/platform/howto/disable-user-2fa.html
https://deploy-preview-1322--developer-aiven-io-preview.netlify.app/docs/tools/terraform/upgrade-provider-v1-v2.html

angelinekwan (approved):

_redirects
lornajane: Oh that sort of makes sense now. Thank you for fixing this!

PR #1329, opened 2022-08-31
Migrate article about Grafana and OpenSearch integration

anelook:
migrated https://help.aiven.io/en/articles/2093033-setting-up-elasticsearch-alerts-in-grafana
The initial article is quite outdated. I also excluded the exact steps of using Grafana. Their interface is changing and also, if we need these instructions they should belong to Grafana section.

### DIFF
--- a/docs/products/opensearch/howto/integrate-with-grafana.rst
+++ b/docs/products/opensearch/howto/integrate-with-grafana.rst
@@ -34,7 +34,7 @@ Integration steps
 #. Set *URL* to ``OPENSEARCH_URI``.
 #. In *Auth* section enable **Basic auth** and **With Credentials**.
 #. In *Basic Auth Details* set your ``OPENSEARCH_USER`` and ``OPENSEARCH_PASSWORD``.
-#. Scroll down to *OpenSearch details* and set the index name(for example, ``logs-*``).
+#. Scroll down to *OpenSearch details* and set the index name or an index pattern (for example, ``logs-*``).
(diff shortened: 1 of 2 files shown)


docs/products/opensearch/howto/integrate-with-grafana.rst
```diff
@@ -0,0 +1,46 @@
...
+
+1. Follow :doc:`these instructions </docs/products/grafana/howto/log-in>` to log in into Aiven for Grafana.
+#. In *Configuration menu* select **Data sources**.
+#. Click to **Add data source**.
+#. Find **OpenSearch** in the list and select it. You'll see a panel with list of settings to fill in.
+#. Use your preferred in the *Name* field. You'll use it later for creating dashboards and alerts.
+#. Set *URL* to ``OPENSEARCH_URI``.
+#. In *Auth* section enable **Basic auth** and **With Credentials**.
+#. In *Basic Auth Details* set your ``OPENSEARCH_USER`` and ``OPENSEARCH_PASSWORD``.
+#. Scroll down to *OpenSearch details* and set the index name(for example, ``logs-*``).
```
ftisiot, line 37: should this be the index pattern? Since i believe you can use wildcards?

ftisiot (changes requested):
Really good @anelook, Minimal change possibly needed!

docs/products/opensearch/howto/integrate-with-grafana.rst
anelook: You're right. They call it index name in Grafana data source. But both work. So I changed that!

ftisiot (approved):
LGTM

PR #1331, opened 2022-08-31
Add manual trigger

laysauchoa:
Current, workflow runs periodically.
There is no option to trigger it manually.
Having the option can be useful for the
contributor if the user can not run the checks
locally but still want to see if any links are broken.
In addition, this is useful for debugging when
the build fails, and we want to reproduce the error
again.

GitHub provides a way to do this
with the ``workflow_dispatch``.

This commit is adding this option to the workflow.

### DIFF
--- a/.github/workflows/linkcheck.yaml
+++ b/.github/workflows/linkcheck.yaml
@@ -3,9 +3,10 @@ name: "LinkCheck - All files"
 on:
   schedule:
     - cron: "0 5 * * 2"
+  workflow_dispatch:
 
 jobs:
-  run-checks:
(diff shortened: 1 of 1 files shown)


lornajane:
This is neat, but I had to look up what it is. Can we add something about it alongside wherever the existing linkcheck docs are (I've forgotten if that's the README or CONTRIBUTING file) please?

laysauchoa:
I'm closing this because I do not have a good way to prevent a person running this in their branch (not main) and opening Issues with what was introduced https://github.com/aiven/devportal/pull/1340

PR #1334, opened 2022-09-01
Fixes #1317: changed Cassandra CA advice to use project ca-get

### LINKED ISSUE #1317: `avn service ca get` needs an additional parameter

We have docs for `avn service ca get` (see https://docs.aiven.io/docs/tools/cli/service.html#avn-service-ca-get) but the example command does not work as expected for a kafka service, since `ca` in this case is just a placeholder and it isn't clear what a valid value might be here. Dredging company slack history implies this could be Cassandra-only, please verify whether this is (still) the case and update the docs with a working example and any other information you think would help a user.

[ ] I plan to open a pull request for this

reedwade:
The service CA get command seems to not actually work right now.
But, CA's are shared by the project so using the project CA fetch works (and is working).

The only reference to this was in the Cassandra Python example so I've changed it there plus added the matching anchor in the `avn project` docs.

### DIFF
--- a/docs/products/cassandra/howto/connect-python.rst
+++ b/docs/products/cassandra/howto/connect-python.rst
@@ -27,7 +27,7 @@ Variable                Description
 
 .. Tip::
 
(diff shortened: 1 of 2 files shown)


ftisiot (approved):
LGTM from a docs point of view, might need more investigation on the CLI/API side

Issue #1275, opened 2022-08-22
Open an Issue when linkcheck fails


laysauchoa:
**Problem**
Recently, we have added linkcheck to run periodically, every week. However, someone has to go and look for the failed build and do the fixes.  This could go unseen or unnoticed. 

**Proposed solution**
1) Automatically open an Issue when the linkcheck fails. So we can better monitor it and fix it.

PR #1340, opened 2022-09-01
Open Issue if linkcheck fails

laysauchoa:
We are running our ``linkcheck`` on weekly basis
for all the files on devportal to check if we
spot broken links on the documentation.

We can manually check after the job run
if it failed and check why.

You can do this by going to
GitHub Action > Workflows > ``LinkCheck - All files``
and see the error messages.

This could easily go unnoticed. We need an explicit
way to show to the devportal maintainers
if something is wrong with the devportal links.

A proposal is to open a GitHub issue in case
broken links are spotted on the periodical checks.

Fixes: https://github.com/aiven/devportal/issues/1275

Tests: I've tested this on a fork project and seems to be working.

### DIFF
--- a/.github/ISSUE_TEMPLATE/linkcheck_report.md
+++ b/.github/ISSUE_TEMPLATE/linkcheck_report.md
@@ -1,0 +1,10 @@
+---
+title: 🚨 Weekly linkcheck failed
+labels: Maintanance, bug
+---
+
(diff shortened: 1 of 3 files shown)


TibsAtWork (approved):
Sounds like a good idea to me - let's see how it works in practice!

Issue #810, opened 2022-05-13
Migrate article: Setting up SAML authentication with Azure


lornajane:
Migrate the content from this article to fit the documentation portal format https://help.aiven.io/en/articles/3557077

(Assign this issue to yourself before you start, and remember to mention the issue number when you open your pull request)

dewan-ahmed:
Hi @lornajane I see that this issue has an ``SA`` tag. I also see that https://github.com/aiven/devportal/issues/813 is closed since https://github.com/aiven/devportal/pull/1206 was merged.

1. Does https://github.com/aiven/devportal/pull/1206 cover Azure SAML as well?
2. If not, Is there an SA who'll be taking this on?

Since this is one of the high-priority articles to migrate by this month, I'm happy to take it on if no one is assigned already.

encima:
> Hi @lornajane I see that this issue has an `SA` tag. I also see that #813 is closed since #1206 was merged.
> 
> 1. Does [Migrate SAML OneLogin configuration article #1206](https://github.com/aiven/devportal/pull/1206) cover Azure SAML as well?
> 2. If not, Is there an SA who'll be taking this on?
> 
> Since this is one of the high-priority articles to migrate by this month, I'm happy to take it on if no one is assigned already.

Sorry @dewan-ahmed I did not see your comment when I assigned this. Each PR/Issue covers one single IdP so #1206 covers only OneLogin.

PR #1357, opened 2022-09-07
migrate azure AD article for SAML

encima:

Migrate Azure AD article. Format to match other SAML articles (with changes to existing article as well). Remove screenshots and add context to steps to make it clearer without the need for images

Addresses #810

### DIFF
--- a/.github/vale/styles/Aiven/capitalization_headings.yml
+++ b/.github/vale/styles/Aiven/capitalization_headings.yml
@@ -6,6 +6,7 @@ scope: heading
 match: $sentence
 exceptions:
   - ACLs
+  - Active Directory
   - Amazon
   - Amazon Aurora
   - Amazon Web Services
--- a/_toc.yml
+++ b/_toc.yml
@@ -91,6 +91,8 @@ entries:
           - file: docs/platform/howto/saml/saml-authentication
           - file: docs/platform/howto/saml/setup-saml-onelogin
             title: Setting up SAML with OneLogin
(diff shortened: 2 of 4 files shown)


ftisiot (changes requested):
@encima could you please check the warnings?

Most of them can be solved by adding `Azure` in the `.github/vale/styles/capitalization_headings.yml` file :)

encima:
> @encima could you please check the warnings?
> 
> Most of them can be solved by adding `Azure` in the `.github/vale/styles/capitalization_headings.yml` file :)

Should Azure be added or should I change the reference to `Microsoft Azure`?

ftisiot:
Good question... @TibsAtWork @jlprat opinions? Should we reference to Azure as `Azure` or `Microsoft Azure`?

jlprat:
What I understand after reading their page is that we need to use `Microsoft Azure`.

https://www.microsoft.com/en-us/legal/intellectualproperty/Trademarks/

encima:
> What I understand after reading their page is that we need to use `Microsoft Azure`.
> 
> https://www.microsoft.com/en-us/legal/intellectualproperty/Trademarks/

Odd, because they call it `AAD` (Azure Active Directory) in most of their marketing... 🤔

ftisiot:
Hi @encima, fixed a bit the formatting, can you please scan it to check everything seems still ok? Then we're ready to merge!

encima:
Thank you for doing it, my eyes would have missed the numbering things!

ftisiot (approved):
LGTM!

PR #1366, opened 2022-09-09
Fixing kafka-go link within product docs.

dewan-ahmed:

https://docs.aiven.io/docs/products/kafka/howto/connect-with-go.html has a small mistake where **With library kafka-go** section refers to the Sarama library instead of kafka-go.

This PR fixes that mistake.


dewan-ahmed:
``kafka-go`` had to be taken out of the link to please Vale check which it was failing because kafka was not capitalized.

laysauchoa (approved):

PR #1370, opened 2022-09-11
Add a test for missing :alt: in image directives

TibsAtWork:
We always want alt text on our images.

This check looks for image directives that are missing the `:alt:` part.

Since the regular expression is a little complex, I've tried to explain it in a comment.

Note: if you have `.. image: <some image>` at the end of the file and don't have a newline at the end of your final line, then (a) this check won't find it, and (b) you've not been meant to do that for the last several decades at least, so please end your files with a newline at the end of the last line.


ftisiot (approved):
LGTM! Thanks @TibsAtWork for adding this functionality

PR #1372, opened 2022-09-12
pg: add documentation around deprecated TLS versions

mble:

A small subset of PG services still allow deprecated TLS versions to be used to connect to the services. This is not intentional, and will be removed very soon.

We will be creating a user alert to notify users
where insecure TLS versions are being used.


dewan-ahmed (approved):
LGTM.

Thanks for the article @mble

PR #1374, opened 2022-09-13
Set Snowplow storage strategy when disabling anonymoius tracking

vtainio:

This will set the domain user_id values correctly for users that have given consent to analytics cookies.


carabaestlein (approved):
LGTM!

PR #1375, opened 2022-09-13
Fix broken link

laysauchoa:
A broken link was spotted by our weekly
link check job. The following link returns 404 code:

https://jdbc.postgresql.org/download.html

The download file can be found here:

https://jdbc.postgresql.org/download/

So this commit replaces the broken link for a valid one.

Fixes: https://github.com/aiven/devportal/issues/1373


dewan-ahmed (approved):
LGTM

Thanks @laysauchoa

PR #1380, opened 2022-09-14
Replace master to main

laysauchoa:
We do not have a master branch,
so we should run those jobs in the
main branch instead.

If something changes in the main branch,
for example.

### DIFF
--- a/.github/workflows/linkcheck-changed-files.yaml
+++ b/.github/workflows/linkcheck-changed-files.yaml
@@ -2,7 +2,7 @@ name: "LinkCheck - Changed files"
 "on":
   push:
     branches:
-      - master
+      - main
     tags:
       - "**"
   pull_request: null
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -2,9 +2,9 @@ name: Linting
 on:
   push:
     branches:
-      - master
+      - main
     tags:
-      - '**'
+      - "**"
   pull_request:
 
 jobs:


.github/workflows/linkcheck-changed-files.yaml
```diff
@@ -2,7 +2,7 @@ name: "LinkCheck - Changed files"
 "on":
   push:
     branches:
-      - master
+      - main
     tags:
       - "**"
```
laysauchoa, line 7: I think we can leave the tags as it is cause we are using tags even is just one for now.

ftisiot:
Hi @laysauchoa, this on paper seems correct, however I dunno how to test it... any hints?

lornajane (approved):
Thanks @laysauchoa  (sorry for the slow review). We can't test this as-is but I've checked it against the docs and I think the change is correct so I'll merge it and watch closely.

laysauchoa:
@ftisiot a way to test this is by forking ``devportal`` and bringing this changes manually to there and merging to your main forked one branch. Then you will need to trigger changes to your ``main`` and see if the workflow runs when pushing to ``main``. 

@lornajane I'm not sure I understood what the change should do besides running the checks on the ``main`` brach when someone push to it. Should this stop merging broken things to the ``main``? or what is expected?

lornajane:
There's a separate configuration change that means we can no longer merge a PR with a failing build.

Your PR here means that we actually run the build on main - and anyone subscribed to the notifications will know if the build fails (there was no notification when this happened before). So now it shouldn't happen - and if it does, we'll know.

laysauchoa:
thanks, this confirms my local tests.

PR #1381, opened 2022-09-14
Added reference to PostgreSQL extension 'file_fdw' back as part of superuser-only extensions

pparmin:
I have added the PostgreSQL extension `file_fdw` back to our list of extensions as part of the superuser-only section. This extension was previously removed (see issues #705 and #706). However, since `file_fdw` is part of PostgreSQL itself: https://www.postgresql.org/docs/current/file-fdw.html we can not delete it from the output of `pg_available_extensions`. 

There were a number of other extensions that were listed in `pg_available_extensions` but could not be installed, which is why @mble created PR #1225. 

This PR adds `file_fdw` to the list introduced by Matt.

### DIFF
--- a/docs/products/postgresql/reference/list-of-extensions.rst
+++ b/docs/products/postgresql/reference/list-of-extensions.rst
@@ -240,6 +240,9 @@ The following extensions can only be installed by superusers, **and are not gene
 ``dict_xsyn`` - https://www.postgresql.org/docs/current/dict-xsyn.html
     Text search dictionary template for extended synonym processing.
 
+``file_fdw`` - https://www.postgresql.org/docs/current/file-fdw.html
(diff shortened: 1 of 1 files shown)


laysauchoa:
@pparmin do we need to add a warning about this specific extension on this page? I understood that the user may not be able to use it, correct?

mble:
@laysauchoa all of the extensions in the "superuser-only" extensions cannot be installed by the user, but are there for posterity as they can be seen in `pg_available_extensions`.

laysauchoa:
Clear to me @mble would be possible to help us to review this PR as you are familiar with this feature? Many thanks!

ftisiot (approved):
LGTM

Issue #895, opened 2022-06-01
Search should show best articles from either platform


lornajane:
# What's wrong?

The search results show ALL devportal articles, no matter how low their relevance, before ANY help center article (that might well be exactly what is needed). We should _prefer_ devportal articles since they should be more up to date and longer-lived, but if the correct content exists, it should be shown on page 1.

The Search function uses a weighting that's set at index time; I think we need to handle this differently

ak-aiven:
https://github.com/aiven/devportal/pull/1385

PR #1385, opened 2022-09-15
Remove default sort by source. Add boost for devportal articles.

ak-aiven:
Modify the search functionality so help articles are sorted better by relevancy

### DIFF
--- a/netlify/functions/search/search.js
+++ b/netlify/functions/search/search.js
@@ -27,12 +27,28 @@ const handler = async (event) => {
     const response = await client.search({
       index: "devportal",
       body: {
-        sort: [{ sort_priority: "asc" }, "_score"],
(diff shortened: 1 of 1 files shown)


ak-aiven:
https://github.com/aiven/devportal/issues/895

encima (approved):
Thanks for picking this up; much better!

laysauchoa:
Hi @encima and @ak-aiven, were you able to test it locally? I was not able to, so this is why I am asking. On the way, I found something else wrong. 


Instructions: https://github.com/aiven/devportal/blob/main/SEARCH.md instructions

encima:
I was able to test in the deployed preview. The index itself does not need to change, though, only how it is queried is changing here

PR #1397, opened 2022-09-20
Fix several warnings

laysauchoa:
Fix some warnings like unknown document and non supported highlight for syntax.


ftisiot (approved):
LGTM

PR #1297, opened 2022-08-25
Aiven challenges on devportal

maimei:
# What changed, and why it matters

PR #1398, opened 2022-09-20
Fix a spelling that was causing a broken build

lornajane:

Teeny fix to turn the build green, we merged something with a failing build from #1297


ftisiot (approved):
Lgtm

PR #1399, opened 2022-09-21
New Rolling Channel and fixes for catch-the-bus warnings

ftisiot:

Added a new Rolling Channel and fixes for catch-the-bus related warnings


dewan-ahmed:
Hi @ftisiot @Ugbot will this be an evergreen content? From a quick read, this seems like a great blog content. Please let me know and I can review either way.

ftisiot:
Hi @dewan-ahmed, 
this is a challenge we'll announce at Current, and probably we'll keep there for future reference. There's a similar one in the `catch-the-bus` file and this type of challenges and positioning has already been discussed and agreed 
previously. Since it's a challenge, part of the answer is hidden, but it would be good if you could check the steps and confirm they work as expected. 

Also please don't merge, @maimei and I need to sync on codes to use for the trial

dewan-ahmed:
Sounds good @ftisiot 

I went through the steps (validating links) sans running the actual notebook. Looks good from me (really interesting challenge) and I'll leave this PR to the safe hands of y'all.

maimei (approved):
Looks great!

Issue #1400, opened 2022-09-21
Addressable Advanced Config Options


mhoffm-aiven:
As a user it would be nice to be able to link to individual config options. This would enable us to create better alerts and in general help our users better.

laysauchoa:
Hi, thanks for the proposal and yes, I agree so we are able to refer to it with links!

PR #1401, opened 2022-09-21
Make addressable advanced config options

laysauchoa:
Fixes: https://github.com/aiven/devportal/issues/1400

### DIFF
--- a/includes/config-cassandra.rst
+++ b/includes/config-cassandra.rst
@@ -1,57 +1,83 @@
 
-``ip_filter`` => *array*
+``ip_filter``
+-------------
+*array*
   **IP filter** Allow incoming connections from CIDR address block, e.g. '10.20.0.0/16'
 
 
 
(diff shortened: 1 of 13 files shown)


ftisiot (commented):
I'm not sure about the alignment, the "array" and description in the below image are not properly aligned, maybe we need to rethink about the structure?

misalignement of field type and desc

laysauchoa:
Hi @ftisiot this was intentional because imho looks better. What would look better in your opinion? Here are some suggestions:

Screenshot 2022-09-22 at 11 31 14

Screenshot 2022-09-22 at 11 33 12

If you have a clear idea in your mind, would be possible to elaborate or show an example?

ftisiot:
I would go for your second option
image

laysauchoa:
it seems good to me @ftisiot I added now :)

ftisiot (approved):
LGTM!

Issue #263, opened 2021-09-28
Document risks and sideeffects of pg upgrades more clearly


jankatins:
# What should we add to our content collections?

Re https://github.com/aiven/devportal/blob/main/docs/products/postgresql/howto/upgrade.rst

We "recommend" that ppl test an upgrade on a fork and also "verify the impact": there are two benefits from it: a) ensuring that it succeeds and is performed quickly enough (which might not be the case) and b) to test query performance directly after upgrade under real world load (when no statistics are available and caches are cold). Basically if you cannot survive any surprises during/after a pg upgrade, then such a fork + upgrade + immediate test under real world load (before AUTOVACUUM does the analyse!) should happen. See the next two items which show possible problems/errors:

* pg_upgrade usually takes a few seconds, but can take minutes or even hours in some pathological cases (e.g. involving lot's of databases or lot's "large objects". This can potentially also lead to OOM kills if the schema is so large and the plan is too small for it to let the `pg_dump/pg_restore` phase of `pg_upgrade --link` succeed. Also some schema changes in PGs default catalog or used extensions in combination with customer objects depending on it could mean that we would not be able to finish the upgrade at all (e.g. if the customer adds a `SELECT *` view +  function on top of `pg_catalog.pg_stat_activity` which added an additional field in newer versions -> `*` returns all columns, but function return value has columns hardcoded).
* We do not transfer optimizer statistics and not running the recommended VACUUM would result in a lot of cases (my guess) in very bad query performance. So `running ANALYZE ensures that queries run efficiently` could be made stronger with an additional `Without running ANALYZE it is expected that your query performance will in some cases be drastically worse!` or something like that...

 (Sidenote: Any failed pg_upgrade will result in a rolled back user config to the old version and the pg_upgrade logs will be emailed to the customer)

In addition, any pg_upgrade has some risk of downtime and data loss if the node goes down before the system is back in a normal state: During upgrade, we will not have any standbys and starting up standbys will only happen after the upgrade has happened AND after the first basebackup is up (which can take a long time, in the double digit hours range for huge clusters!). Currently it says: `Standby nodes are removed and replacements are created for the new version` which makes it sound we prepare new standbys already beforehand. Basically, if this takes some time, it means we do not have a hot standby during that time which means downtime and data loss if a node goes down (e.g. Cloud provider has problems) as we manually (via support ticket) have to reset the service config to the old pg version, restore from the old backup and cannot replay the wal files from the time pg_upgrade finished until the node went down.

[...]

PR #1406, opened 2022-09-23
Add specifics of what issues may be encountered

ak-aiven:
Make the readers aware that usage of read replicas can reduce risk of issues Make the recommendation of forking and testing more than a cautionary call out


ak-aiven:
#263

encima (approved):
LGTM

PR #1418, opened 2022-09-27
Fixing file name in mongo source connector

jorgeuribe-aiven:
It is a minor change, I was reading the document and found that the file they are  referring to is `mongodb_source.json` file that was previously created.


staceysalamon-aiven (approved):

staceysalamon-aiven:
Good catch @jorgeuribe-aiven! Thanks for fixing this!

PR #1422, opened 2022-09-28
services-forking: update article to address customer concern

vietnguyen-aiven:

Some customers may want to fork their production clusters. One of them recently reached out to us asking if there are any additional load to their production cluster when the fork is being made.

The short answer is no, and this article is now updated to address this concern from customers by giving a short paragraph to emphasize that the fork cluster does not communicate with the old cluster at all and so will not cause any load to it.

### DIFF
--- a/docs/platform/concepts/service-forking.rst
+++ b/docs/platform/concepts/service-forking.rst
@@ -10,7 +10,7 @@ When you fork a service, the following items are copied into the new service:
 - Service users
 - Connection pools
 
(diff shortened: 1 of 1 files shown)


ftisiot (approved):
LGTM

vietnguyen-aiven:
I did not expect my PR to be reviewed and merged so quickly, thank you @ftisiot 🤩

ftisiot:
The additional text was not huge but very important, thanks for it!

PR #1443, opened 2022-10-10
Add logs to Issue message

laysauchoa:
Add partial broken logs info to it. 

Linkcheck runs weekly and reports an issue when broken links are found. One has to look into the failed action and filter the broken links from the extensive list manually. This PR automates this step. by bringing the broken link messages directly to the issue.

The logs messages should be parsed to something like this:

Screenshot 2022-10-10 at 14 04 15

### DIFF
--- a/.github/workflows/linkcheck.yaml
+++ b/.github/workflows/linkcheck.yaml
@@ -23,12 +23,22 @@ jobs:
       - name: LinkCheck
         id: run_tests
         run: |
-          make linkcheck > logs.txt || echo "BROKEN=$(grep -E "broken" logs.txt)" >> $GITHUB_ENV
+          make linkcheck > logs.txt || echo "OUTPUT=$(grep -E "broken" logs.txt | tr '\n' '@' | sed 's/@/<br>/g' | sed 's/broken/*\*`broken`\*\*/g' | sed $'s/\e\\[[0-9;:]*[a-zA-Z]//g')" >> $GITHUB_ENV
 
       - name: Create Issue
-        if: "${{ env.BROKEN != '' }}"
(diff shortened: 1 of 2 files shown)


.github/workflows/linkcheck.yaml
```diff
@@ -21,11 +21,24 @@ jobs:
           python -m pip install --upgrade pip
           pip install -r requirements.txt
       - name: LinkCheck
-        run: make linkcheck
+        id: run_tests
+        run: |
+          make linkcheck > logs.txt || echo "OUTPUT=$(grep -E "broken" logs.txt | tr '\n' '@' | sed 's/@/<br>/g' | sed 's/broken/*\*`broken`\*\*/g' | sed $'s/\e\\[[0-9;:]*[a-zA-Z]//g')" >> $GITHUB_ENV
```
TibsAtWork, line 26: A little tiny bit of me twitches at assuming there won't be an `@` in the log line already, but in reality it should be OK, and if not we can fix it if (😢) the problem occurs.

TibsAtWork (approved):
Thanks for this - it looks like a really useful change.

.github/workflows/linkcheck.yaml
laysauchoa: It is hacky, but I did not find any other better way. The main issue is the formatting from what I save to a variable to what is displayed in the GitHub body issue.

Mainly it removes colors, and try to add lines in between the commands by using html <br>.  Thanks for reviewing and yes, if it happens, we can fix it.

laysauchoa: I am actually glad you noticed it and hope if it happens, this comments can be helpful even for my future self. :)

PR #1444, opened 2022-10-10
Made search results require 2 matching query blocks

### LINKED ISSUE #1441: Search is returning 653 page(s) matching the search query "ex9jkCydLqo9rgUalygWgZhWhsPdq3" which is obviously wrong
# What's wrong?
Search is returning 653 page(s) matching the search query "ex9jkCydLqo9rgUalygWgZhWhsPdq3" which is obviously wrong

https://docs.aiven.io/search.html?q=ex9jkCydLqo9rgUalygWgZhWhsPdq3&check_keywords=yes&area=default

kevinbowman-aiven:
Prevents all pages being returned for all queries.  Fixes #1441

### DIFF
--- a/netlify/functions/search/search.js
+++ b/netlify/functions/search/search.js
@@ -32,6 +32,7 @@ const handler = async (event) => {
         from: (currentPage - 1) * pageSize,
         query: {
           bool: {
+            minimum_should_match: 2,
             should: [
               {
                 match: {


lornajane (approved):
Thanks for the fix! I verified with the local netlify dev setup that this change returns only the matching results rather than 600+ each time.

PR #1448, opened 2022-10-11
We no longer use Intercom

lornajane:

It was reported that we still refer to Intercom in our docs although this tool is no longer in use. Remove the mention of this tool.


ftisiot (approved):
LGTM

PR #1449, opened 2022-10-11
Use a nicer listing format for the TF recipes, with product icons

lornajane:

We had requests from @petralylander and team to prepare for more developers coming to this page as an entry point. I think we might need to add more words and links to help orient people who are landing here, but I'm opening the PR anyway to get some initial feedback from everyone.

### DIFF
--- a/docs/tools/terraform/reference/cookbook.rst
+++ b/docs/tools/terraform/reference/cookbook.rst
@@ -1,6 +1,110 @@
 Cookbook
 =========
 
-A collection of recipes for the Aiven Terraform provider cookbook.
+Sometimes an example is easier to understand than a long tutorial or detailed configuration documentation. We've put together some of the most common terraform setups, and listed them here. Use them as they are, or combine them to cook up something new and tasty for your own needs.
 
-.. tableofcontents::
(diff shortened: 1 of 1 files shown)


dewan-ahmed:
Thanks for the landing page @lornajane - this is great 💯 

Does the framework allow us to create search filters? For example, if we end up with a few dozen recipes, a customer can select "PostgreSQL" and the recipes with the tag PostgreSQL will be shown. 

Most text and links would be here: https://docs.aiven.io/docs/tools/terraform.html. Are you thinking of any specific text to include on this landing page? It already looks pretty good IMO. But I can certainly add some text that you suggest.

If we create developer education videos for each recipe, could the cards hold a button for a YouTube video or is there another way we can add the video links in the future?

lornajane:
It's a static site so no filtering (we can build it at some point but it doesn't come built in, no).

Might be nice to add the video indicators once we have a few of the recipes available in video format, definitely.

dewan-ahmed (approved):
Sounds good. I'll merge this one now and we can add more texts later since this already looks great 🥇 thank you.

PR #1452, opened 2022-10-12
Minor fixes for some warnings on build

lornajane:

I see some build warnings, so this fixes them


TibsAtWork (approved):
These all look good to me. Thanks for fixing them.

PR #1378, opened 2022-09-13
Put the list of kafka connectors on the kafka connect landing page

lornajane:
# What changed, and why it matters

Definitely a work in progress, to maybe-possibly group the connectors and make a bit more of a landing page feature. 
Fix #1273

PR #1482, opened 2022-10-18
Updated Source and Sink connectors to our available howtos

bcrisologo:
- Updated the links for source and sink connectors with the available Howto documentation we have.  Need to double check that each one is pointed correctly.
- Also removed the `Preview` labels for the ones we already have a Howto page


ftisiot:
Closing since this is going to be fixed via https://github.com/aiven/devportal/pull/1378

PR #1491, opened 2022-10-20
add article about high availability for OpenSearch

mkokho:
 New article, explains what happens when a node fails in an OpenSearch cluster. Similar to the existing HA articles for Redis and Postgres

works locally :) 

Screen Shot 2022-10-21 at 12 33 35 PM copy

### DIFF
--- a/_toc.yml
+++ b/_toc.yml
@@ -595,6 +595,7 @@ entries:
           - file: docs/products/opensearch/concepts/backups
           - file: docs/products/opensearch/concepts/indices
           - file: docs/products/opensearch/concepts/aggregations
+          - file: docs/products/opensearch/concepts/high-availability-for-opensearch
           - file: docs/products/opensearch/concepts/opensearch-vs-elasticsearch
           - file: docs/products/opensearch/concepts/shards-number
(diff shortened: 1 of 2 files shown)


dewan-ahmed (commented):
Hi @mkokho thanks for this contribution. 
1. Since startup and hobbyist plans don't have HA, could we rephrase "limited availability" on the table (High Availability Features column) to "none"? Leaving this to your expertise and this is just a suggestion. 
2. I see three uses of "Aiven infrastructure" and one use "Aiven platform" in this article. Since Aiven is a PaaS rather than an IaaS, do you think using  "Aiven platform" throughout will be more correct?

mkokho:
thank you @dewan-ahmed , fixed your 2nd point!
As for the 1st, I'm just staying consisten what we say in other HA article (Posrtgres, Redis). I think we say "limited availability" because we do try to replace broken nodes. So the service is available, but not highly available, which is still better than on-premise setup where a manual intervention is required to replace a broken node.

dewan-ahmed (approved):
LGTM

thanks @mkokho for your contribution 🙌🏽

PR #1501, opened 2022-10-24
Gcp-pubsub-source

ftisiot:

Missing examples for Google pubsub


encima (approved):
Unsure if we should point directly to the dedicated repo instead of the deprecated one? Approving regardless as that is the current version we run

ftisiot:
Yeah, let's keep things as are now (pointing to the old repo). Once we move to the new repo I'll update
Thanks for the speedy review @encima

PR #1506, opened 2022-10-25
Added missing Pub/Sub connect examples

ftisiot:

Added missing Pub/Sub example

### DIFF
--- a/_toc.yml
+++ b/_toc.yml
@@ -382,6 +382,8 @@ entries:
             - file: docs/products/kafka/kafka-connect/howto/redis-streamreactor-sink
             - file: docs/products/kafka/kafka-connect/howto/cassandra-streamreactor-sink
             - file: docs/products/kafka/kafka-connect/howto/couchbase-sink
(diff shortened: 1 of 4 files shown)


_toc.yml
```diff
@@ -375,6 +375,8 @@ entries:
             - file: docs/products/kafka/kafka-connect/howto/mongodb-sink-lenses
             - file: docs/products/kafka/kafka-connect/howto/redis-streamreactor-sink
             - file: docs/products/kafka/kafka-connect/howto/cassandra-streamreactor-sink
+            - file: docs/products/kafka/kafka-connect/howto/gcp-pubsub-sink
+            - file: docs/products/kafka/kafka-connect/howto/gcp-pubsub-lite-sink
```
encima, line 379: Nitpicking so feel free to ignore but the organisation is not in alphabetical order, do we want it to be?

encima (approved):

_toc.yml
ftisiot: I believe we can review the sorting at a later stage?
I created an issue for this #1525

PR #1515, opened 2022-11-01
Edited GCP Marketplace migration instructions for clarity

dbeech:

We require information from both old and new environments to complete the migration. The way this was written (all combined in one sentence) was unclear in my view, so I've separated things out to make it more obvious what is what.


dewan-ahmed (approved):
Thanks for this change @dbeech 

LGTM

PR #1534, opened 2022-11-07
mysql: bugfix for max connections [DOC-84]

wojcik-dorota:
Updated article `MySQL max_connections` to reflect our new way of defining the max number of connections for a service depending on a service plan/RAM

https://aiven.atlassian.net/browse/DOC-84

### DIFF
--- a/.github/vale/styles/Aiven/capitalization_headings.yml
+++ b/.github/vale/styles/Aiven/capitalization_headings.yml
@@ -36,6 +36,7 @@ exceptions:
   - Elasticsearch
   - European Union
   - Flink
+  - GiB
   - GitHub
   - Go
   - Google BigQuery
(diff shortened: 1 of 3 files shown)


packi (approved):
LGTM, I would squash the commits though.

lornajane:
Let me know if I can help with the squashing/rebasing/merging on this one?

.github/vale/styles/Aiven/capitalization_headings.yml
```diff
@@ -36,6 +36,7 @@ exceptions:
   - Elasticsearch
   - European Union
   - Flink
+  - GiB
```
lornajane, line 39: Could we double check the units intended here? GB would be gigabytes (which is what I expect, this is a standard measure), but GiB is for gibibytes ... is that correct?

wojcik-dorota: @packi , could you please confirm?

packi: We actually check for 4 GiB in the code

PR #1544, opened 2022-11-09
Add redirects to ClickHouse files that were renamed

anelook:


dewan-ahmed:
@anelook I assume the redirects should take effect on the preview. I don't see the link appear for the newly added redirects. For example on https://deploy-preview-1544--developer-aiven-io-preview.netlify.app/docs/products/clickhouse/howto/integrate-pg.html, the <grant-privilege> is supposed to be a link  
Screenshot 2022-11-09 at 2 05 51 PM

Could you please double-check?

lornajane:
I think there is some confusion here, there's quite a few problems to fix. In this PR, we're quickly adding redirects for the situation where content has been moved from one URL to another - this breaks existing links and so we have incoming links from external sources that will be failing. There are also broken links between these pages, but this PR doesn't tackle that.

lornajane (approved):
Thanks for this quick interim fix!

Issue #1485, opened 2022-10-19
Replace redirected links


laysauchoa:
We have more than 120 redirected links in the docs. It is better to make clear for the user what they are accessing, and even to ourselves where those redirected are taking us to visit as we have in our docs. You can find the redirects by running ``linkcheck``.

laysauchoa:
From your experience with documentation @TibsAtWork does this make sense? Any feedback is helpful, thank you

TibsAtWork:
I don't have any directly useful experience, but some thought

My worry about putting comments into the text (which I've done myself on occasion 😢) is that it doesn't get updated - and in the worst case, the target may change so the link is no longer a redirect and then the comment in the text is just wrong!

However, there are some simple things we can do to reduce the number of redirects we're seeing, which we should probably do first.

I (just now) see 129 redirected links (searching for `redirect` in the logs)

* 17 are because we are using `http` in the link when we should use `https`. We should just fix those.
* 35 are where we've still got `developer.aiven.io/` instead of `docs.aiven.io`. We should fix those as well, or even consider using `:ref:` or `:doc:` again now they no longer cause things to go wrong.
*  8 are `temporarily` redirected - these seem to match the pattern of links like `https://opensearch.org/docs/im-plugin/index/` redirecting to `https://opensearch.org/docs/latest/im-plugin/index/`. Those are OK and don't need any description, as it's just part of how the target is handling versioning.
* 85 are `permanently` redirected - for instance, `https://www.m3db.io/` to `https://m3db.io/`. We should consider updating those to the new location, _*but*_ note that the redirection of links like `https://grafana.com/docs/grafana/latest/visualizations/bar-chart/` to `https://grafana.com/docs/grafana/v9.0/visualizations/bar-chart/` *is* something we want to happen.

That's 145 where we either have something to fix, or are definitely already doing the right thing. And yes, that's more than 129, so some things are occurring more than once, which is OK, but it does suggest we might reduce the (apparent) problem rather dramatically with some relatively simple changes.

TibsAtWork:
It occurs to me that I didn't make it obvious how pleased I am that you're looking at this! It's definitely something we should be keeping track of, so thanks for doing so.

TibsAtWork:
I think I'd be tempted to sort out the easy-to-sort links by hand first - certainly the http and developer.aiven.io ones.

I'm *assuming* one always wants to ignore temporary redirects, but that is an assumption.

[...]

PR #1547, opened 2022-11-10
Reduce the number of redirected link errors

TibsAtWork:
When `make linkcheck` is run, there are a lot of redirected links (search for `redirect` in the linkcheck output)

**NOTE** I'm going to split this up into smaller PRs, for ease of review...

This PR aims to fix those that are simple to fix:

* Using `http` instead of `https` when appropriate
* Replace `developer.aiven.io` with `docs.aiven.io`

   Note: *not* for `_redirects.txt`, as that's correctly saying `developer.aiven.io` redirects to `docs.aiven.io`

   I believe it is correct to alter `robots.txt` and `conf.py`

   We *could* come back later and turn some/many of these into `:doc:` references instead - see https://github.com/aiven/devportal/issues/1548

### DIFF
--- a/SEARCH.md
+++ b/SEARCH.md
@@ -49,7 +49,7 @@ In addition to these `url` field should be provided with every document but it i
 
 ## Search function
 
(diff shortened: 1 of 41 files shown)


TibsAtWork:
This PR has been replaced by smaller PRs.

PR #1549, opened 2022-11-10
Fix link redirects: use `https` instead of `http`

TibsAtWork:

This PR fixes the links that used `http` instead of `https`

### DIFF
--- a/docs/products/postgresql/howto/pg-long-running-queries.rst
+++ b/docs/products/postgresql/howto/pg-long-running-queries.rst
@@ -5,14 +5,14 @@ Aiven does not terminate any customer queries even if they run indefinitely, but
 
 To identify and terminate such long-running queries, you can do it from either:
 
-* `Aiven Console <http://console.aiven.io>`__
+* `Aiven Console <https://console.aiven.io>`__
 * `PostgreSQL® shell <https://www.postgresql.org/docs/current/app-psql.html>`_ (``psql``)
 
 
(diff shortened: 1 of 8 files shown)


docs/products/postgresql/howto/pg-long-running-queries.rst
```diff
@@ -81,4 +81,4 @@ which you should see the role::
     
 To resolve the permission issue, you may grant the user the appropriate role as per below::
 
-    grant testrole to avnadmin;
```
laysauchoa, line 84: was this something you changed or maybe github is showing it wrong?

TibsAtWork: It's got a `+` on the next line for the same text, and I believe the only difference is adding in a newline at the end of the last line of the file.
GitHub doesn't make it very obvious, but `gitk`, for instance, mentions it explicitly.
I'm torn about a change like this - it's an incidental change, caused by my editor, but honestly all lines in a file should end with a new line.
(This matters more on Unix than on other systems - see https://medium.com/@alexey.inkin/how-to-force-newline-at-end-of-files-and-why-you-should-do-it-fdf76d1d090e for some background)

laysauchoa: I agree with you that all lines should end with a new line, so thanks for clarifying this behavior.

laysauchoa (approved):
looks good to me, thanks for fixing it!

Issue #1548, opened 2022-11-10
Use Spinnx `:doc:` roles for links to docs.aiven.io documents


TibsAtWork:
We used to have to be more cautious about using `:doc:` because Vale would spell check the document names.

Now that's not an issue, we could look for all the `https://docs.aiven.io` links and consider converting them to `:doc:` links.

PR #1550, opened 2022-11-10
Fix link redirects: Use `docs.aiven.io` instead of `developer.aiven.io`

TibsAtWork:

This PR replaces links using `developer.aiven.io` with links to `docs.aiven.io`

**Note:** not for `_redirects.txt`, as that's correctly saying `developer.aiven.io` redirects to `docs.aiven.io`

I believe it is correct to alter `robots.txt` and `conf.py`

We could come back later and turn some/many of these into `:doc:` references instead - see https://github.com/aiven/devportal/issues/1548

### DIFF
--- a/SEARCH.md
+++ b/SEARCH.md
@@ -49,7 +49,7 @@ In addition to these `url` field should be provided with every document but it i
 
 ## Search function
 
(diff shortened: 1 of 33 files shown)


laysauchoa (approved):
Great job, Tibs! I agree, let's fix those ones now and have a ticket about turning into :doc links.

PR #1551, opened 2022-11-10
Fix redirect links: Change `help.aiven.io` redirects to the `docs.aiven.io` page

TibsAtWork:

This PR fixes those `help.aiven.io` links that could instead reference `docs.aiven.io` pages.

We could come back later and turn some/many of these into :doc: references instead - see https://github.com/aiven/devportal/issues/1548

Note: there seem to be some places where a newline is being added at the end of the file.


laysauchoa (approved):
I checked all the links and looks good to me, thanks for the changes!

PR #1552, opened 2022-11-10
Fix link redirects: Change miscellaneous permanent redirects

TibsAtWork:

Change miscellaneous permanent redirects to the new URL.

**Note:** not "fixing" those redirects that go from a generic document to a specific version, as those are quite correct.

* Fix `microsoft` redirects - mostly `docs` -> `learn`


lornajane (approved):
Thank you! I would like to offer my particular compliments for the PR strategy for that one issue - this is much easier to review in batches like this!

PR #1553, opened 2022-11-10
Adding a common troubleshooting instruction.

dewan-ahmed:

It seems that we're seeing the following build error frequent enough to have the fix be part of the build instructions:

```
Error: pg_config executable not found.
      
      pg_config is required to build psycopg2 from source.  Please add the directory
      containing pg_config to the $PATH or specify the full executable path with the
      option:
      
          python setup.py build_ext --pg-config /path/to/pg_config build ...
      
      or with the pg_config option in 'setup.cfg'.
      
      If you prefer to avoid building psycopg2 from source, please install the PyPI
      'psycopg2-binary' package instead.
```

I wasn't sure if I could add a `code` directive within a `note` directive so I used literal symbol for the `brew install...` command. Another option is to add a troubleshooting section somewhere else. Please suggest any change that you prefer.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -25,9 +25,7 @@ Install the dependencies::
 
     pip install -r requirements.txt
 
-.. note::
-
-    You might need to run ``brew install postgres@<VERSION>`` (or a similar command for your package manager) to install PostgreSQL® as a ``psycopg2`` build dependency. 
+**Note:** You might need to run ``brew install postgresql@14`` to install PostgreSQL 14 (or a similar command for your package manager) to install PostgreSQL® as a ``psycopg2`` build dependency. 
 
 Start the HTML version with::
 
--- a/README.rst
+++ b/README.rst
@@ -25,7 +25,7 @@ Install the dependencies::
 
     pip install -r requirements.txt
 
-**Note:** You might need to run ``brew install postgresql@14`` to install PostgreSQL 14 (or a similar command for your package manager) to install PostgreSQL® as a ``psycopg2`` build dependency. 
+**Note:** If that fails because it can't install ``psycopg2-binary``, run ``brew install postgresql@14`` to install PostgreSQL 14 (or a similar command for your package manager) and then try again.
 
 Start the HTML version with::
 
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -25,6 +25,10 @@ Install the dependencies::
 
     pip install -r requirements.txt
 
+.. note::
+
+    You might need to run ``brew install postgres@<VERSION>`` (or a similar command for your package manager) to install PostgreSQL® as a ``psycopg2`` build dependency.
```
TibsAtWork, line 30: Thanks for this - documenting a sensible work-around while we figure out what to actually do sounds sensible,

The brew package name is `postgresql@14` at the moment - which is unusual, as one doesn't normally specify the version. I'd probably say
```
... run ``brew install postgresql@14`` to install PostgreSQL 14 (or a similar command ...
```

Also, GitHub won't render a `note` directive properly, so it's better to do "presentational markup" - that is, something more like:

```
    **Note:** You might need ...
```

(that's indented so it becomes a block quote, which makes it stand out as not part of the main text.

TibsAtWork:
To answer the question in the Discussion at the top, reStructuredText is quite happy with (for instance) a `code` directive inside a `note`, but it won't show up the way you want in GitHub because GitHub's reST rendering won't use the representation that is generated for notes (or warnings or any other admonitions). So the way you're doing it makes sense.

README.rst
```diff
@@ -25,6 +25,8 @@ Install the dependencies::
 
     pip install -r requirements.txt
 
+**Note:** You might need to run ``brew install postgresql@14`` to install PostgreSQL 14 (or a similar command for your package manager) to install PostgreSQL® as a ``psycopg2`` build dependency.
```
TibsAtWork, line 28:
```suggestion
**Note:** If that fails because it can't install ``psycopg2-binary``, run ``brew install postgresql@14`` to install PostgreSQL 14 (or a similar command for your package manager) and then try again.
```
I think it's better to put the problem first, and also I believe it's specifically a problem with the binary package.

lornajane (approved):
Thanks!

PR #1558, opened 2022-11-11
Console overview docs updates

staceysalamon-aiven:
* Updated alt text so it accurately describes the new screenshot
* Changed headings as suggested in the review of the initial publishing
* Removed how-to section in Aiven Console in Tools
* Moved new articles (Create accounts, Create and manage teams) to how-to section in Platform
* Added new section in Platform how-to section for account management

Note to reviewer: In testing, the left menu has unexpected behavior when navigating to the Create accounts article in the Account management section. The menus all collapse and the new section is not displayed. I wasn't able to figure out why this is happening or whether it is just an issue with the preview.


lornajane:
:warning: contains file renames without redirects. Let's figure out if we _need_ the rename, and if so then add the associated redirects?

staceysalamon-aiven:
> ⚠️ contains file renames without redirects. Let's figure out if we _need_ the rename, and if so then add the associated redirects?

@lornajane  Argh, sorry I totally forgot about the redirects!! 

I do think these should be moved because at the moment they have been put in a separate HowTo section under Tools, but we already have Platform instructions in that how-to section.  What do you think? If you are okay with this, I will update the redirects file.

lornajane:
Since the navigation and the file structure are (intentionally) independent, it's not essential to rename files to move them between sections. Their URLs reflect their folder locations, but we can build the navigation any way we like so I am tempted to say that we should just update the toc and not rename the files, especially since we're linking to them in both locations by the look of it. I'm open to discussion if that would bother anyone though!

lornajane:
Also I see you have a failing build job - click on the "Details" tab next to the job that failed to see what happened (let me know if we should walk through this part together, I don't know how familiar this bit is for you).

staceysalamon-aiven:
> Since the navigation and the file structure are (intentionally) independent, it's not essential to rename files to move them between sections. Their URLs reflect their folder locations, but we can build the navigation any way we like so I am tempted to say that we should just update the toc and not rename the files, especially since we're linking to them in both locations by the look of it. I'm open to discussion if that would bother anyone though!

This makes sense! I've changed the file locations/links back now.

staceysalamon-aiven:
> Also I see you have a failing build job - click on the "Details" tab next to the job that failed to see what happened (let me know if we should walk through this part together, I don't know how familiar this bit is for you).

Yeah, thanks for checking. The error were related to the broken links, so all good now. 👍

lornajane:
I'm not sure who can review the content here, would you need help finding someone?

My only other question is whether you see a lot more articles being added in the new account management section in the future? It looks a bit strange with only one item in it. Is this part of a wider plan?

staceysalamon-aiven:
> I'm not sure who can review the content here, would you need help finding someone?
I can add a reviewer.

> My only other question is whether you see a lot more articles being added in the new account management section in the future? It looks a bit strange with only one item in it. Is this part of a wider plan?
There would be more here, though the terminology is going to change so I will need to also take that into account. I did not know about all the issues with renaming when planning this.

rsalevsky (approved):

PR #1562, opened 2022-11-14
Only show top-level titles in tableofcontents output by default

lornajane:

We need to talk about and test this one I think. There are lots of places where we use a `.. tableofcontents::` entry to list all the pages in a section, particularly the howto and concepts sections. It doesn't look really great and I'm looking at ways we can restyle it. This change switches the default to be to show only the page titles and not the subheadings on a page as a first step - my idea is that we can then try to style the list items more nicely.

Does this work? Or does it make things worse?

Example of the current state: https://docs.aiven.io/docs/products/postgresql/concepts.html
And the preview of just the pages list: https://deploy-preview-1562--developer-aiven-io-preview.netlify.app/docs/products/postgresql/concepts.html

### DIFF
--- a/_toc.yml
+++ b/_toc.yml
@@ -1,5 +1,5 @@
 root: index
-options:
+defaults:
   maxdepth: 1
 
 entries:


TibsAtWork:
My thoughts went:

1. Yes, the page is neater without the subheadings - it looks less cluttered. Which matters if the ToC were much longer than this example.
2. But now it's just the same content as in the sidebar.
3. And actually sometimes the context of the subtitles is useful - it already tells me something about the topic that I didn't necessarily know.

Which means that in the end I'm sort-of ambivalent about the change.

I did wonder whether we could have "expandable" items on the page, to allow expanding out the subtitles, but I think that would only help if it was possible to do "expand all", and I'm not sure how we'd flag that as being possible.

lornajane:
The goal is to get a simpler output that can be styled as cards in sections. The sections _might_ come in through use of subtrees, but we needed to simplify the page list output.

TibsAtWork:
In that case this approach would seem sensible.

lornajane:
@staceysalamon-aiven I'd like to get your eyes on this one, it's a sort of first step towards being able to render these lists as cards

staceysalamon-aiven:
@lornajane I really like the idea of having cards on these pages. This seems like a good first step!

@TibsAtWork's point about the subheadings is interesting too. I wonder if (in a later iteration) it would be worth having a snippet of the page or some sort of description metadata that can also be displayed on the cards? That could get wordy pretty quickly too, though.

lornajane:
Right now we're using the first part of the page content as the excerpt, to keep the contribution barrier low. With a much larger staff to create and maintain these descriptions for every page, I think this would be a nice feature.

If anyone would like to approve the change, go ahead and I'll merge. Otherwise I'll close this eventually if it doesn't seem like something we want.

ftisiot:
I'm approving this since it brings simplicity on the list pages, which is very needed in some of the most cluttered platform pages. We can then iterate in the future for improvements on using cards

ftisiot (approved):
LGTM

PR #1564, opened 2022-11-14
Fix JSON syntax error that breaks highlighting

lornajane:

The JSON syntax was invalid and causing a warning about syntax highlighting. This fixes the stray `"` character (the small fixes are the best!)


ftisiot (approved):
LGTM

PR #1567, opened 2022-11-14
add upgrade tf provider from v2 to v3 page

byashimov:

Adds instructions to upgrade terraform files to use them with the latest version of provider.


dewan-ahmed:
Thanks for this contribution @byashimov 

I'm taking a 👀

dewan-ahmed:
Hi @byashimov I've pushed minor style fixes via https://github.com/aiven/devportal/pull/1567/commits/ca0cb73b5f19f069c1a325cc7327df07d6ad5537

On Step3, is there an extra space after `-backup=./` ? I was going to run the steps myself but in case you already ran all the commands for this scenario, I'll skip that part. 

https://developer.hashicorp.com/terraform/cli/commands/state/rm says that `-backup` is a legacy option. Is that because there's another alternative or they don't recommend storing backup state files locally and `-backup` is still the only option to do this?

byashimov:
@dewan-ahmed I have accidentally force-pushed the branch, just a bad habit, so sorry :( 
Your commit is cherry-picked. And thanks for testing that, I have made few amends.  
I moved backup tip to the "tip", since terraform makes a backup on it's own, but one can make a mistake, better have a copy of that.

Issue #1569, opened 2022-11-15
Introduce EOL for M3 version


hoang-vo:
# What should we add to our content collections?

M3 1.1 and 1.2 will be EOL by next year 1/5/2023. This issue is opened to add a section about EOL policy for M3 on devportal.

# Is this something you would work on yourself?

[x ] I plan to open a pull request for this

PR #1570, opened 2022-11-15
Tweak m3 and opensearch eol policy

hoang-vo:

Introduce M3 EOL policy and tweak previous OpenSearch EOL policy that we had.

solve #1569


harshini-rangaswamy (approved):
Looks good!

Issue #1512, opened 2022-10-28
Add more explicit tags for ClickHouse and Flink


laysauchoa:
Those services are in beta version maybe we can add in more pages like "Get started page"about the release cycle that those versions are at the moment. 

> Explicitly mentioning it is beta everywhere publicly can avoid some bad expectations on cases

PR #1593, opened 2022-11-23
Add explicit beta tags for Flink and ClickHouse

laysauchoa:
This is a request from support to make it
more visible that those products are on beta
stage. Helps to manage expectations between
our clients and our support team about those
two products.

Fixes: https://github.com/aiven/devportal/issues/1512

### DIFF
--- a/docs/products/clickhouse.rst
+++ b/docs/products/clickhouse.rst
@@ -1,5 +1,5 @@
 ClickHouse® |beta|
-===================================================================
+==================
 
 .. note::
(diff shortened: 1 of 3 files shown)


ftisiot:
Hi @laysauchoa, I saw that you added the beta flag in a couple more pages, is the extent enough to cover the requirements of the support team?

laysauchoa:
They said it would be nice to have in the Getting started pages. So it should be fine. I can add also to more pages if you think it makes sense, let me know :)

ftisiot (approved):
LGTM

PR #1599, opened 2022-11-24
InfluxDB: Add Advanced Parameters workflow

laysauchoa:
This will automatically fetch changes for
InfluxDB advanced parameters and open a pull
request with those changes.

Runs periodically.


TibsAtWork (approved):
Matches the equivalent for the other services, so looks good.

PR #1601, opened 2022-11-25
postgresql: add subsection and note for recreation of replication slots

wojcik-dorota:
https://aiven.atlassian.net/browse/DOC-8

Adding information on the feature enabling the recreation of replication slots to the existing articles on the related subject

### DIFF
--- a/docs/products/postgresql/concepts/upgrade-failover.rst
+++ b/docs/products/postgresql/concepts/upgrade-failover.rst
@@ -61,8 +61,5 @@ During maintenance updates, cloud migrations, or plan changes, the below procedu
 Recreation of replication slots
 -------------------------------
 
-In case of failover or controlled switchover of an Aiven for PostgreSQL service, the replication slots from the old primary server are not recreated in the new primary server by default.
+In case of failover or controlled switchover of an Aiven for PostgreSQL service, the replication slots from the old primary server are recreated in the new primary server by default.
 
-.. important::
-
-    Enable the recreation of replication slots in the new primary server before opening connections and redirecting the traffic.
--- a/docs/products/postgresql/concepts/upgrade-failover.rst
+++ b/docs/products/postgresql/concepts/upgrade-failover.rst
@@ -81,4 +81,4 @@ At the time when a failover to a standby occurs, the standby already has active 
 .. warning::
     
     * In case of uncontrolled failover, slots created up to 30 seconds before the failover might be lost.
-    * Arrival of recovered replication slots positions from the old primary to the new primary might be several seconds delayed. If case of delay, you might receive the same entities that you've already received when you connected to the slot after the failover without specifying a position.
+    * Arrival of recovered replication slots positions from the old primary to the new primary might be several seconds delayed. If delayed, you might receive the same entities that you've already received when you connected to the slot after the failover without specifying a position.
--- a/docs/products/postgresql/concepts/upgrade-failover.rst
+++ b/docs/products/postgresql/concepts/upgrade-failover.rst
@@ -63,6 +63,10 @@ Recreation of replication slots
 
 In case of failover or controlled switchover of an Aiven for PostgreSQL service, the replication slots from the old primary server are automatically recreated in the new primary server.
 
+.. important::
+
+    The recreation of replication slots feature is enabled automatically and doesn't require restarting the nodes.
+
 One-node cluster
 """"""""""""""""
 
--- a/docs/products/postgresql/concepts/upgrade-failover.rst
+++ b/docs/products/postgresql/concepts/upgrade-failover.rst
@@ -61,5 +61,25 @@ During maintenance updates, cloud migrations, or plan changes, the below procedu
 Recreation of replication slots
 -------------------------------
 
-In case of failover or controlled switchover of an Aiven for PostgreSQL service, the replication slots from the old primary server are recreated in the new primary server by default.
+In case of failover or controlled switchover of an Aiven for PostgreSQL service, the replication slots from the old primary server are automatically recreated in the new primary server.
+
+One-node cluster
+""""""""""""""""
+
+Before replacing a node in the one-node cluster, the new node acquires information on replication slots on the original service, re-creates them, and only then the failover is performed.
+
+Multi-node cluster
+""""""""""""""""""
+
+For multi-node setups, replication slots from the primary are synchronized to the standbys periodically. At regular time intervals
+
+* Dependencies for newly-created slots are installed in the corresponding databases (currently, every 30 seconds)
+* Positions (``confirmed_flush_lsn``) of the slots are synchronized on the primary and the standbys.
+
+At the time when a failover to a standby occurs, the standby already has active replication slots with fairly up-to-date positions from the primary (with a possible 5-second delay).
+
+.. note::
+
+    * In case of uncontrolled failover, slots created up to 30 seconds before the failover might be lost.
+    * Since, positions of recovered replication slots on the new primary might be several seconds delayed from the old primary, you might receive entities that you've already received before when connecting to the slot after the failover without specifying a position.
 
--- a/docs/products/postgresql/concepts/upgrade-failover.rst
+++ b/docs/products/postgresql/concepts/upgrade-failover.rst
@@ -65,7 +65,7 @@ In case of failover or controlled switchover of an Aiven for PostgreSQL service,
 
 .. important::
 
-    The recreation of replication slots feature is enabled automatically and doesn't require restarting the nodes.
+    The recreation of replication slots feature is enabled automatically.
 
 One-node cluster
 """"""""""""""""
@@ -81,9 +81,13 @@ For multi-node setups, replication slots from the primary are synchronized to th
   When the new slot is created on a database and we want to re-create this slot on a standby, we use a functionality from the ``aiven_extras`` extension, which needs to be installed in the database. Therefore, every 30 seconds there is a job checking that this extension is installed on the databases with logical replication slots.
 * Positions (``confirmed_flush_lsn``) of the slots are synchronized between the primary and the standbys.
 
-When a failover to a standby occurs, the standby node already has active replication slots with an up-to-date (maximum 5-second delay) positions from the primary.
+When a failover to a standby occurs, the standby node already has replication slots with an up-to-date (maximum 5-second delay) positions from the primary.
 
 .. warning::
     
     * In case of uncontrolled failover, slots created up to 30 seconds before the failover might be lost.
+
+.. note::
+    
     * Position of recovered replication slots might be up to several seconds older than on the original primary. Therefore, when re-connecting to PostgreSQL and reading from replication slots, it's recommended to use start positions known to the client until which the data was already received. Otherwise, the client might receive duplicate entries.
+    * In case of failover with a huge lag between the primary node and the standby node (for example, when a master disappears), the position of the replication slot restored on a new master is not newer than the position on the standby node, even though the position of that slot on the old master was newer. 
--- a/docs/products/postgresql/concepts/upgrade-failover.rst
+++ b/docs/products/postgresql/concepts/upgrade-failover.rst
@@ -77,14 +77,13 @@ Multi-node cluster
 
 For multi-node setups, replication slots from the primary are synchronized to the standbys periodically. At regular time intervals
 
-* Dependencies for newly-created slots are installed in the corresponding databases (currently, every 30 seconds)
-* Positions (``confirmed_flush_lsn``) of the slots are synchronized on the primary and the standbys.
+* Dependencies for newly-created slots are installed in the corresponding databases (currently, every 30 seconds).
+  When the new slot is created on a database and we want to re-create this slot on a standby, we use a functionality from the ``aiven_extras`` extension, which needs to be installed in the database. Therefore, every 30 seconds there is a job checking that this extension is installed on the databases with logical replication slots.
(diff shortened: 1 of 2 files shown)


docs/products/postgresql/concepts/upgrade-failover.rst
```diff
@@ -57,3 +57,12 @@ During maintenance updates, cloud migrations, or plan changes, the below procedu
 
 .. Note::
     The old primary server is kept alive for a short period of time (minimum 60 seconds) with a TCP forwarding setup pointing to the new primary server allowing clients to connect before learning the new IP address.
+
+Recreation of replication slots
+-------------------------------
+
+In case of failover or controlled switchover of an Aiven for PostgreSQL service, the replication slots from the old primary server are not recreated in the new primary server by default.
```
alexole, line 64: Not the case anymore, after this https://github.com/aiven/aiven-core/pull/26663 they will be created automatically.

docs/products/postgresql/concepts/upgrade-failover.rst
```diff
@@ -57,3 +57,9 @@ During maintenance updates, cloud migrations, or plan changes, the below procedu
 
 .. Note::
     The old primary server is kept alive for a short period of time (minimum 60 seconds) with a TCP forwarding setup pointing to the new primary server allowing clients to connect before learning the new IP address.
+
+Recreation of replication slots
+-------------------------------
+
+In case of failover or controlled switchover of an Aiven for PostgreSQL service, the replication slots from the old primary server are recreated in the new primary server by default.
```
ftisiot, line 64: Can we get a bit more info on what's happening here? I believe this deviates from the standard PostgreSQL feature so we need to tell people a bit more what we're doing under the hood

alexole: Replication slots are recovered after failover on the new primary automatically. This is a new behaviour after https://github.com/aiven/aiven-core/pull/26663 
I would avoid word `default` as there is no other non-default behaviour now.

How it works (not sure if everything should be put to the public doc):
1. If we have only one node cluster, the new node before replacing the old one will get info about replication slots on the original service and will re-create them and will then fail over.
2. For multi-node setups we synchronize the replication slots from the primary to standbys periodically.
    * every 30 (we might change it, so probably should not put a specific number here) seconds we ensure that for newly created slots the dependencies are installed in the corresponding databases
    * every several seconds we synchronize the positions (confirmed_flush_lsn) of the slots on the primary and standbys
    * if failover happens to the standby, it will have replication slots with almost up-to-date positions from the primary (position of the replication slots on the new primary might be around 5 seconds behind the positions of old primary)

Therefore in case of uncontrolled failover slots created up to 30 seconds before failover might be lost.
Due to the specifics of implementation, the position of the recovered replication slot might be several seconds older then on the primary, so in case of connecting to the slot after the failover without specifying position, the client might receive old entities which has already received before.

docs/products/postgresql/concepts/upgrade-failover.rst
```diff
@@ -57,3 +57,29 @@ During maintenance updates, cloud migrations, or plan changes, the below procedu
 
 .. Note::
     The old primary server is kept alive for a short period of time (minimum 60 seconds) with a TCP forwarding setup pointing to the new primary server allowing clients to connect before learning the new IP address.
+
+Recreation of replication slots
```
ftisiot, line 61: Should we have a dedicated article for this? It's a pretty important (and not usual) setup on our side

docs/products/postgresql/concepts/upgrade-failover.rst
```diff
@@ -57,3 +57,29 @@ During maintenance updates, cloud migrations, or plan changes, the below procedu
...
+
+Before replacing a node in the one-node cluster, the new node acquires information on replication slots on the original service, re-creates them, and only then the failover is performed.
+
+Multi-node cluster
+""""""""""""""""""
+
+For multi-node setups, replication slots from the primary are synchronized to the standbys periodically. At regular time intervals
+
+* Dependencies for newly-created slots are installed in the corresponding databases (currently, every 30 seconds)
+* Positions (``confirmed_flush_lsn``) of the slots are synchronized on the primary and the standbys.
```
ftisiot, line 77:
```suggestion
* Positions (``confirmed_flush_lsn``) of the slots are synchronized between the primary and the standbys.
```

docs/products/postgresql/concepts/upgrade-failover.rst
```diff
@@ -57,3 +57,29 @@ During maintenance updates, cloud migrations, or plan changes, the below procedu
...
+
+At the time when a failover to a standby occurs, the standby already has active replication slots with fairly up-to-date positions from the primary (with a possible 5-second delay).
```
ftisiot, line 79:
```suggestion
When a failover to a standby occurs, the standby node already has active replication slots with an up-to-date (maximum 5-second delay) positions from the primary.
```

docs/products/postgresql/concepts/upgrade-failover.rst
```diff
@@ -57,3 +57,29 @@ During maintenance updates, cloud migrations, or plan changes, the below procedu
...
+
+.. note::
+
+    * In case of uncontrolled failover, slots created up to 30 seconds before the failover might be lost.
```
ftisiot, line 83: This should be a warning

docs/products/postgresql/concepts/upgrade-failover.rst
```diff
@@ -57,3 +57,29 @@ During maintenance updates, cloud migrations, or plan changes, the below procedu
...
+    * Since, positions of recovered replication slots on the new primary might be several seconds delayed from the old primary, you might receive entities that you've already received before when connecting to the slot after the failover without specifying a position.
```
ftisiot, line 84: I believe this sentence needs to be ironed out, as of now is difficult to understand

ftisiot (dismissed):
Hi @wojcik-dorota, good progress, left some comments

docs/products/postgresql/concepts/upgrade-failover.rst
wojcik-dorota: @divdby0 @alexole do you think we need a separate article for Recreation of replication slots? Or is the section in this article ok?

docs/products/postgresql/concepts/upgrade-failover.rst
```diff
@@ -57,3 +57,28 @@ During maintenance updates, cloud migrations, or plan changes, the below procedu
...
+
+* Dependencies for newly-created slots are installed in the corresponding databases (currently, every 30 seconds)
+* Positions (``confirmed_flush_lsn``) of the slots are synchronized on the primary and the standbys.
+
+At the time when a failover to a standby occurs, the standby already has active replication slots with fairly up-to-date positions from the primary (with a possible 5-second delay).
+
+.. warning::
+    
+    * In case of uncontrolled failover, slots created up to 30 seconds before the failover might be lost.
+    * Arrival of recovered replication slots positions from the old primary to the new primary might be several seconds delayed. If delayed, you might receive the same entities that you've already received when you connected to the slot after the failover without specifying a position.
```
wojcik-dorota, line 84: @alexole , could you please help me improve the second sentence here `you might receive the same entities that you've already received when you connected to the slot after the failover without specifying a position.`? It's unclear what's happening here.

docs/products/postgresql/concepts/upgrade-failover.rst
divdby0: I understand @ftisiot 's point that this is a valuable feature and requires a bit more highlighting. However, I also believe that this is categorically the correct place for this feature. Because most of the importance of this feature comes from its recent built-up customer expectations. I believe, over time it will be normalized, and this feature will be a standart, expected, intuitive behavior, similarly of our existing handling of the physical replication slots.

Note that we have already been in touch lately with some of the customers via the CSMs, and we'll continue to use the CSM channel to promote this feature to the customers that demanded have this feature in the past.

docs/products/postgresql/concepts/upgrade-failover.rst
alexole: When client connects to replication slots and starts reading from it, it sends back information to the server about how much it read, the server will move to position of the slot forward (`confirmed_flush_lsn`) to the position client confirmed it received data until.

When the client connects to the slot it can specify the position from which it wants to continue reading, if position is not specified it will start receiving data from `confirmed_flush_lsn` on the server.

Say, client has read until position P1, it will be stored on the server as `confirmed_flush_lsn` and if the client disconnects and connects again, the client has two options, either connect without starting position and it will receive all the data from P1 or it can specify explicitly start position P1, in normal scenario it's the same.

Now let's imaging the client has read until P1 and then fail-over happens. In this case the `confirmed_flush_lsn` on the new master might fallback to P0 (0 < P1 - P0 <~5 sec). Fail-over will trigger client re-connection and now if the client connects without a start position, it will start reading from P0 (not P1), so it will receive whatever was sent between P0 and P1 again. In order to mitigate this, it should explicitly start reading from P1.

We should also specify one important moment I forgot to mention in the beginning, in case of fail-over when there is a huge lag between the primary and standby (say master disappears), the position of replication slots will also be old, so the positions could not be bigger than the position of the standby

docs/products/postgresql/concepts/upgrade-failover.rst
```diff
@@ -57,3 +57,32 @@ During maintenance updates, cloud migrations, or plan changes, the below procedu
...
+""""""""""""""""
+
+Before replacing a node in the one-node cluster, the new node acquires information on replication slots on the original service, re-creates them, and only then the failover is performed.
+
+Multi-node cluster
+""""""""""""""""""
+
+For multi-node setups, replication slots from the primary are synchronized to the standbys periodically. At regular time intervals
+
+* Dependencies for newly-created slots are installed in the corresponding databases (currently, every 30 seconds)
```
alexole, line 80: This sentence without a context is a bit confusing, I would add a bit more.
When the new slot is created on a database, in order to re-create this slot on a standby, we use functionality from `aiven_extras` extension, which needs to be installed in that database. Therefore every 30 seconds there is a job checking that this extension is installed on the databases with logical replication slots.

docs/products/postgresql/concepts/upgrade-failover.rst
alexole: Even though the complexity of the feature might deserve a separate article, from customer POV it should have worked already :) So I would just leave it here.

alexole:
I could not find information about this
```
We should also specify one important moment I forgot to mention in the beginning, in case of fail-over when there is a huge lag between the primary and standby (say master disappears), the position of replication slots will also be old, so the positions could not be bigger than the position of the standby
```

wojcik-dorota:
> I could not find information about this
> 
> > We should also specify one important moment I forgot to mention in the beginning, in case of fail-over when there is a huge lag between the primary and standby (say master disappears), the position of replication slots will also be old, so the positions could not be bigger than the position of the standby

@alexole , regarding " the positions could not be bigger" - could you please specify what positions you mean here? 
"...so the positions of ???? could not be bigger ..." . Sorry about that, I guess we need that for clarify here.

alexole:
I mean the position of the replication slot restored on a new master will not be newer than the position of this new master, even though the position of that slot on the old master was newer.

docs/products/postgresql/concepts/upgrade-failover.rst
```diff
@@ -57,3 +57,33 @@ During maintenance updates, cloud migrations, or plan changes, the below procedu
...
     The old primary server is kept alive for a short period of time (minimum 60 seconds) with a TCP forwarding setup pointing to the new primary server allowing clients to connect before learning the new IP address.
+
+Recreation of replication slots
+-------------------------------
+
+In case of failover or controlled switchover of an Aiven for PostgreSQL service, the replication slots from the old primary server are automatically recreated in the new primary server.
+
+.. important::
+
+    The recreation of replication slots feature is enabled automatically and doesn't require restarting the nodes.
```
alexole, line 68: Customer can't restart the node anyway, so I would remove this second part of the sentence.

docs/products/postgresql/concepts/upgrade-failover.rst
```diff
@@ -57,3 +57,33 @@ During maintenance updates, cloud migrations, or plan changes, the below procedu
...
+Multi-node cluster
+""""""""""""""""""
+
+For multi-node setups, replication slots from the primary are synchronized to the standbys periodically. At regular time intervals
+
+* Dependencies for newly-created slots are installed in the corresponding databases (currently, every 30 seconds).
+  When the new slot is created on a database and we want to re-create this slot on a standby, we use a functionality from the ``aiven_extras`` extension, which needs to be installed in the database. Therefore, every 30 seconds there is a job checking that this extension is installed on the databases with logical replication slots.
+* Positions (``confirmed_flush_lsn``) of the slots are synchronized between the primary and the standbys.
+
+When a failover to a standby occurs, the standby node already has active replication slots with an up-to-date (maximum 5-second delay) positions from the primary.
```
alexole, line 84: Please remove word `active` as it can be confusing, replication slot is active in PG when the client is already connected to it.

ftisiot (approved):
LGTM!

alexole:
Looks OK to me.
I just thought that maybe we should add into uncontrolled failover section another sentence, saying something like
"If due to cloud provider failure the node from one node cluster disappears, it's not possible to restore replication slots on a new - replacement node, as replication slots information is not stored anywhere else but running primary node"

Issue #1507, opened 2022-10-26
Local devportal `pip install` can fail on Mac M1 because of `psycopg2-binary`


TibsAtWork:
The `requirements.txt` file has a requirement for `psycopg2-binary`.

On Mac M1 machines, installation of `psycopg2-binary` can sometimes be a problem. This has recently been an issue for @wojcik-dorota.

The script `scripts/create_feedback_table.py`, which is used by `make create-feedback-table`, does `import pscopg2`, but I assume this script is rarely  run (I hope table creation is infrequent).

Since that script is not needed in local development work with devportal, the work-around was to temporarily delete the requirement from the requirement file, do the `pip install -r requirements.txt`, and then undo the change. This works, but isn't very friendly.

https://github.com/aiven/devportal/issues/1243 may also be related - this was an undiagnosed problem that sounds similar but was fixed by installing postgresql itself.

Do we actually need `psycopg2-binary` in the requirements file?

If creating the feedback table is the only reason it is needed, I'd vote for either putting a comment in the script and/or Makefile to say that the package needs installing before running the script, or (less nice) putting an explicit `pip install` in the relevant target in the Makefile.

If we do still need `psycopg2-binary` in the requirements file, we should document why we need PostgreSQL support (preferably in the README) as it's not obvious, and we should also have a note for M1 Macs with what the workaround or solution is if this problem occurs again.

laysauchoa:
Would it help if we move it to [requirements-dev.txt](https://github.com/aiven/devportal/blob/main/requirements-dev.txt)?

TibsAtWork:
Sometimes doing `brew install postgres@14` (or `brew install postgres` I assume!) will fix it - but I've seen reports that this doesn't *always* help, and it does assume the person is using `homebrew`.

dewan-ahmed:
This is probably a workaround and not a permanent fix: https://github.com/aiven/devportal/pull/1553

@TibsAtWork looking at you to suggest a better fix or if we can close this PR without merging.

TibsAtWork:
@dewan-ahmed is updating the README with a workaround at https://github.com/aiven/devportal/pull/1553

We can readdress that if/when we fix this by some other means.

TibsAtWork:
I think the workaround is a good idea (https://github.com/aiven/devportal/pull/1553) as we're clearly not going to fix this issue "properly" before that PR gets accepted.

PR #1605, opened 2022-11-28
Remove requirement for psycopg2-binary

TibsAtWork:
The `psycopg` module is only needed by one script, which is not expected to be used in normal use cases. Unfortunately, on current M1 macs, installing psycopg2-binary can be a problem. So move addressing the problem into the script.

Fixes https://github.com/aiven/devportal/issues/1507

**Note** also adds a newline to the end of the script - this should be there anyway.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -25,8 +25,6 @@ Install the dependencies::
 
     pip install -r requirements.txt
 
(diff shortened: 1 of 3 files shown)


laysauchoa (approved):
I found it clever, Tibs! Thank you

PR #1609, opened 2022-11-29
Improve automated workflows

laysauchoa:
- Add manual trigger in case an update needs to be done before the scheduled time.
This will give more flexibility to create the
automated docs at any time when needed it.

- Signal the committer from bot. Helps to account the bots contributions over time.

This has been tested and looks like this:

Screenshot 2022-11-29 at 12 22 03
Screenshot 2022-11-29 at 12 22 08


TibsAtWork (approved):
1. I looked up the `workflow_dispatch` feature, and that does look like a very useful thing to add.
2. Specifying the committer/author seems like a good idea, as you show in the screenshots.

PR #1610, opened 2022-11-29
Clean up linkcheck workflow

laysauchoa:
This is step is not needed and
it was added by mistake. Removig because
it is not needed and takes longer to run with
it.


TibsAtWork (approved):
Seems sensible, thanks

PR #1614, opened 2022-11-30
Update list of Grafana plugins

FloorD:
Cross-checked with the plugins available for Aiven for Grafana by spinning up an instance. Adding a few, modifying others for which a new and an old version are available.


anelook (commented):
For reviewers, do you need that a reviewer also runs the instance and double checks?

The change to plugin.rst looks good to me (but I didn't spin up the service). 
One questions are changes to package-lock.json intentional? They don't seem to fit into the PR

FloorD:
Ha! That IS weird. I spun up a codespace to only make changes to plugins.rst in a branch, maybe something got changed on the main branch?

PR #1615, opened 2022-12-01
Update Grafana plugins list

FloorD:
Cross-checked with the list available in Aiven for Grafana today.


anelook (approved):
Looks good to me

PR #1619, opened 2022-12-06
Fixing build warnings and errors.

dewan-ahmed:
# What changed, and why it matters

This PR addresses some final naming and referencing issues that were causing the `make linkcheck` to show warnings.

Hi @harshini-rangaswamy the other changes on the PR is not related to you but I'm tagging you for your input on the following file:

```
/Users/dewan.ahmed/code/github.com/devportal/docs/products/clickhouse/howto/integrate-kafka.rst:186: WARNING: duplicate label reference, other instance in /Users/dewan.ahmed/code/github.com/devportal/docs/tools/terraform/reference.rst
/Users/dewan.ahmed/code/github.com/devportal/docs/products/kafka/howto/kafka-klaw.rst:2: WARNING: Duplicate explicit target name: "klaw documentation".
```

I think this results when an anchor tag is used within reST/Sphinx and Sphinx gets confused on the reference i.e. it's thinking **Klaw documentation** on lines 54 and 139 as duplicates. What's your thought?

PR #1577, opened 2022-11-17
[Schedule merge 19.12] Feature/use aws

hieu-aiven:
# What changed

- Migrate search & feedback functions away from Netlify and into AWS Lambda
- Remove runtime.txt that define the Python version to use during runtime and set it as environment variable in the deployment platform instead to ensure the site works in both Netlify and Cloudflare Pages during the transition period for DNS update. This is because Netlify requires [Python 3.8](https://github.com/netlify/build-image/blob/focal/included_software.md) while Cloudflare Pages requires [Python 3.7](https://github.com/aiven/devportal/issues/1528).

# Why

- Eliminate deployment platform dependency for search & feedback function
- Better development workflow as 2 functions can now be tested locally with AWS SAM.
- Reduce platform cost as Netlify pricing based on user and Git contributors doesn't scale with Open source project like DevPortal.

# Process for the switch Netlify to Cloudflare Pages
- Function deployment to production with Aiven AWS account. Pending [access](https://aiven.atlassian.net/browse/AM-15898)
- Approve PR feature/use-aws and merge to main
- Update Cloudflare Pages production build from branch feature/use-aws to main
- Update DNS CNAME pointing to CP so the [docs.aiven.io](https://www.google.com/url?q=http://docs.aiven.io&sa=D&source=calendar&ust=1669804544883034&usg=AOvVaw2n4JWTHERxBrZv4TID3wG_) will serve from CP → [SRE-5211](https://aiven.atlassian.net/browse/SRE-5211)

Switch completed

PR #1620, opened 2022-12-06
All builds fail if warnings are generated

lornajane:

This pull request causes all Sphinx warnings to be treated as errors, and will cause the build to fail. This is because we regularly commit with fairly serious errors (broken internal links, invalid markup) that isn't picked up at review stage. When this happens, we have to go back and create followup change to fix the broken things, and the original contributor/reviewer don't necessarily get the feedback.

This change will cause all warnings to be treated as errors.


lornajane:
This branch needs to be rebased when the main branch is clean (#1619 will help, thank you @dewan-ahmed !).

@angelinekwan I am seeing builds on Netlify, should I be seeing this on Cloudflare? Also did we work out if we could get the logs back to the context of the PR? I think we're going to need them in order to be able to enable this setting.

angelinekwan:
>I am seeing builds on Netlify, should I be seeing this on Cloudflare?
DevPortal is still serving from Netlify till [feature/use-aws](https://github.com/aiven/devportal/pull/1577) are approved and merged.
>Also did we work out if we could get the logs back to the context of the PR
Error log is only accessible for authenticated user in the deployment platform apply for both Netlify or Cloudflare Pages

lornajane:
Thanks for the info, I'm trying to think of other ways we can give users the feedback they need to understand what happened. Maybe we can also build the HTML in the GitHub action, since the same errors should show up in most cases.

lornajane:
Adding this build HTML job means that if the build doesn't succeed, this job will fail (as will cloudflare, but we can't see the output from that). However it does mean that we're building twice on every commit, which isn't ideal.

lornajane:
@laysauchoa would you be able to review this one? Would like to get your input.

laysauchoa (approved):
Thanks for the fix, I tested and works well. From the tests, here is what I noticed:
- Check time will increase to about 6 min
- Check stops as soon as it finds an error which means the user don't have a full report when looking in the GitHub Actions
- Running ``make livehtml`` locally won't stop it but it shows an error each time. It does not render new changes until is fixed. There is an error message about it, so it should be fine.

I believe is ready to merge, so I'm approving it. :)

lornajane:
Thanks @laysauchoa ! I assume you didn't merge since it's the last Friday before a holiday? I think I agree, let's merge this when we are back in the New Year :fireworks:

dewan-ahmed:
Very good call from both of you 💯

PR #1623, opened 2022-12-06
Add some signposting to the config listings from the API section

lornajane:

We have all these config sections, but you really need this information when you're using the tools that accept the configuration. So this pull request just adds a small explanation and links to each of the service pages. This could also be moved to a snippet and included for all the tools that use config, so we only have to maintain the links in one place, but I wanted to get something in as a starting point. Feedback welcome!

### DIFF
--- a/docs/products/clickhouse/reference/advanced-params.rst
+++ b/docs/products/clickhouse/reference/advanced-params.rst
@@ -1,5 +1,5 @@
-List of advanced parameters
-============================
+Advanced parameters for Aiven for ClickHouse®
+=============================================
 
 Below you can find a summary of every configuration option available for Aiven for ClickHouse® service:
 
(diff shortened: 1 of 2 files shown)


anelook (commented):
I find it useful if we show links to advanced parameters for every service, however, I am not sure if API is the best place for it. Those parameters can be set through the console as well. 

What if we add a separate file under Aiven tools "Advance parameters" or "Advance configuration" that will have "Per-service configuration" section?

lornajane:
Right. As I said in the PR description, this set of links should probably be a snippet that's used in a few different tools places, I am suggesting this as the starting place.

lornajane:
Sorry, that was a super short response earlier! Do you think we can publish this helper section for the API docs? Or we should wait until we add them in many other places as well and make it one PR?

lornajane:
not a priority

PR #1625, opened 2022-12-07
clickhouse: reorganize info architecture for clickhouse section

wojcik-dorota:
https://aiven.atlassian.net/browse/DOC-69

This PR is targeting the Aiven for ClickHouse GA release - 13 Dec.

This is addressing the issue discussed on Slack/#devportal at https://aiven-io.slack.com/archives/C01T6379FBQ/p1670255426253589?thread_ts=1666607973.174269&cid=C01T6379FBQ

Please find more details on at https://aiven-io.slack.com/archives/C01NVM89Z5Y/p1670433211307929

Essentially, this PR is to restructure the ClickHouse section as specified at https://miro.com/app/board/uXjVPK1O1is=/

Note: The order of the sections in the screenshot below is random - this is just an artificially generated sitemap to give an overview of section subcategorization.

What is the essence of the change made via this PR?

1. New section `Overview` (There is still one more article upcoming for this section - Feature overview).
2. Subcategories added for the `HowTo` section (all HowTo articles used to be on the same level, now they are grouped into five categories).

new-clickhouse-sitemap


docs/products/clickhouse/howto/get-started/connect-to-service.rst
```diff
@@ -0,0 +1 @@
+.. include:: /docs/products/clickhouse/howto/connect.rst
```
harshini-rangaswamy, line 1: Looks like this is not loading the content from connect.rst file as expected. I see a blank screen.

harshini-rangaswamy (changes requested):
At first glance, it looks great. Just found one page where content is not loading correctly.

lornajane:
This feels like it could be a high-risk pull request, in particular there are a lot of renames that will break existing URLs and we are actively trying to avoid carrying a large burden of redirects so we should consider what our options are to achieve the same things. I would also like us to take advice before following the content-duplication pattern introduced by the includes, I'm not sure about the impacts of this either from a content performance or a user experience point of view. Once this is ready to come out of draft, I'd like to be one of the reviewers before we merge, please.

anelook:
I agree with @lornajane. 
One more comment - the .toc tree can be changed according to our desires without renaming and moving files. So, in case we do want to experiment with tree structure, we can still keep files and their locations intact, but just use .toc to put them into desirable flow. This will avoid the need for redirects and also allows us to change the tree structure painlessly multiple times.

dewan-ahmed:
I strongly agree with @lornajane on the includes pattern. I know I've reviewed and merged a few cookbook recipes for ClickHouse that had the `includes` but I was pretty uncomfortable doing so. The main reason is that this pattern creates a lot of hidden dependencies and changing one file creates a cascading effect. If possible, let's stick to :doc: and :ref: please.

wojcik-dorota:
> 

@dewan-ahmed @lornajane , thank you for your feedback. Please confirm I understand it correctly: If I need to reuse an article somewhere else than its original location, instead of creating a new file in the target location and using `include` directive inside it, I actually need to create a new file in the target location and build it as an actual document of its own with just one sentence including the :doc: link to the original article. This would generate a number of articles including one link each just to redirect the user. Or how else would you recommend to handle that?

wojcik-dorota:
@sbijjal @jkowall , @divdby0 please have a look at the comments from DevRels colleagues. With this issues open we might have some challenges to be ready with restructuring the ClickHouse docs by GA.

lornajane:
My main concern with this approach to the includes is that we specifically discussed that we would signpost content, rather than duplicating it. This pull request takes exactly the opposite approach and puts the same content under multiple URLs. This could be confusing to the user, and may also be a problem for SEO - which is why we have a guideline to avoid this.

lornajane:
@wojcik-dorota Please also mention if the reviews you are requesting are particularly timely, I see the comment about something going GA, but most of the reviewers here won't have that context. It is also fine to use internal channels for that discussion rather than the public repository thread if appropriate.

wojcik-dorota:
> 

@lornajane here's the discussion is opened for this PR with the timeline specified  https://aiven-io.slack.com/archives/C01T6379FBQ/p1670499770936199

anelook:
@wojcik-dorota, what I meant regarding **_toc.yml** file, is when you look at https://docs.aiven.io/docs/products/kafka/howto/connect-with-python.html, it is located in howTo folder, however in the menu tree on the left it is under **Code samples**. 

If you look at **_toc.yml** you'll see something like this:

```
entries:
          - file: docs/products/mysql/howto/list-code-samples
            entries:
              - file: docs/products/mysql/howto/connect-from-cli
              - file: docs/products/mysql/howto/connect-with-python
              - file: docs/products/mysql/howto/connect-using-mysqlx-with-python
              - file: docs/products/mysql/howto/connect-with-java
              - file: docs/products/mysql/howto/connect-with-php
```

wojcik-dorota:
This one is followed up on and fixed via https://github.com/aiven/devportal/pull/1629

PR #1626, opened 2022-12-08
clickhouse: add features overview article

wojcik-dorota:
https://aiven.atlassian.net/browse/DOC-59

### DIFF
--- a/docs/products/clickhouse/concepts/features-overview.rst
+++ b/docs/products/clickhouse/concepts/features-overview.rst
@@ -1,0 +1,66 @@
+Aiven for ClickHouse® features overview
+=======================================
+
+Discover key features that Aiven for ClickHouse® offers.
+
+About Aiven for ClickHouse
+--------------------------
+
+Aiven for ClickHouse® is a fully managed distributed columnar database which is built on the open source ClickHouse datastore. It is a fast highly-scalable fault-tolerant solution designed for online analytical processing and data warehousing. Aiven for ClickHouse enables you to execute complex SQL queries on large datasets quickly and effectively to process large amounts of data in real time.
+
+ClickHouse's purpose and application
+------------------------------------
+
+ClickHouse is designed for generation of detailed real-time analytical data reports using advanced SQL queries on large data-sets.
+Its design and high degree of scalability make itallow it to be fast and efficient even when handling large amounts of data.
+ClickHouse might be a good choice for your organization if your data is high volume, usually inserted in batches, and rarely updated or deleted.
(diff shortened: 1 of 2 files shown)


docs/products/clickhouse/concepts/features-overview.rst
```diff
@@ -0,0 +1,106 @@
...
+
+* Reliable operations
+
+  * Self-healing platform with 99.99% SLA
+  * Zero downtime during scaling, upgrading and other management operations
+
+* Superior observability
+
+  * Use pre-integrated Aiven observability services for enhanced monitoring and logging with Grafana, M3, OpenSearch®
+  * Integrate with your favorite external observability tooling — from Datadog to Prometheus, Jolokia, and more
```
jkowall, line 83: Jolokia is only for Kafka (JMX metrics). I would say Datadog and Prometheus.

docs/products/clickhouse/concepts/features-overview.rst
```diff
@@ -0,0 +1,106 @@
...
+
+* Security & compliance
+
+  * Nodes run on dedicated VMs
+  * End-to-end encryption
+  * ISO 27001:2013, SOC2, GDPR, HIPAA and PCI/DSS compliant
+  * Secure networking with VPC peering, PrivateLink or Transit GW
```
jkowall, line 90: Should we use the full name of the service on AWS "Transit Gateway"?

jkowall (changes requested):

docs/products/clickhouse/concepts/features-overview.rst
```diff
@@ -0,0 +1,106 @@
+Aiven for ClickHouse® features overview
+=======================================
+
+What is Aiven for ClickHouse® and what are its key features? Why and when to use Aiven for ClickHouse? What are its most common use cases? This article helps you discover it all and more.
+
+About Aiven for ClickHouse
+--------------------------
+
+Aiven for ClickHouse® is a fully managed distributed columnar database which is built on the open source ClickHouse datastore. It is a fast highly-scalable fault-tolerant solution designed for online analytical processing and data warehousing. Aiven for ClickHouse enables you to execute complex SQL queries on large datasets quickly and effectively to process large amounts of data in real time.
```
divdby0, line 9: I think we should use a more powerful word than datastore. Datastore is kind of limiting the meaning towards the datawarehouse use-case but ClickHouse is also very powerful on the online analytical processing (OLAP) use-case which we should not diminish. A generic term solution could work here IMHO.

Issue #1226, opened 2022-08-09
Add a GitHub Actions job to format embedded Python code within RST


laysauchoa:
This seems like a good idea. I'd recommend black formatter for python code as it is the standard and the default for python code.

Reference: https://pypi.org/project/black/

lornajane:
(triaging the issue list which made me read this one again). We've also talked recently about having testable code, I suggest we move all code to the `code/` folder so we can work on it outside the RST context, and then include the code in the articles as needed. Just mentioning this as a pointer for when we do prioritise this work.

laysauchoa:
Great point, I will change the proposal here to cover only ``.py`` files which can be done independent from ``.rst`` embedded code.

PR #1632, opened 2022-12-10
Add Python autoformatter

laysauchoa:
Use psf/black to format Python code. Run once per week seems fine, but we can also change some of the customizations parameters from the workflow. This can be triggered manually and on the schedule.

Fixes: https://github.com/aiven/devportal/issues/1226

### DIFF
--- a/.github/workflows/python-run-autoformatter.yaml
+++ b/.github/workflows/python-run-autoformatter.yaml
@@ -1,7 +1,7 @@
 name: Python formatting - Run psf/black
 on:
   schedule:
-    - cron: "0 6 * * 2"
+    - cron: "0 7 * * 2"
   workflow_dispatch:
 jobs:
   linter_name:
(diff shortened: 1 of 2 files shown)


laysauchoa:
Rebasing with main branch.

jkowall:
No need to implement anything new on this IMO.

laysauchoa:
Hi @jkowall, I am not so sure I understand your reasons to not have a formatter check for the Python code 🐍. You can check that the sample code on ``code/products/cassandra/connect.py`` does not work for the user, and having such formatter check would prevent that.

jkowall:
We are moving away from this codebase in the coming months, so we'll not be using the code here. The current thoughts are migrating the docs to a standard [docusaurus](https://docusaurus.io/) based system. Much of the developer content will be moving to our new developer center once that is launched.

PR #1634, opened 2022-12-12
Migrate service metrics articles to new documentation portal

pparmin:
The following two lists containing Prometheus metrics were used by some of our customers and have not been migrated to https://docs.aiven.io/. 

They were captured in the docs archive: 
- [5144867](https://github.com/aiven/aiven-intercom-docs-archive/tree/main/archive/5144867)
--> Old site: [Web archive](https://web.archive.org/web/20220627052050/https://help.aiven.io/en/articles/5144867-aiven-service-metrics)
- [5144953](https://github.com/aiven/aiven-intercom-docs-archive/blob/main/archive/5144953/5144953-additional-service-metrics.rst)
--> Old site: [Web archive](https://web.archive.org/web/20220519045153/https://help.aiven.io/en/articles/5144953-additional-service-metrics)

The below commits are a one-to-one translation from the old help portal. Only the links between both pages have been updated to reflect the new structure.

### DIFF
--- a/_toc.yml
+++ b/_toc.yml
@@ -71,6 +71,8 @@ entries:
         title: Monitoring management
         entries:
           - file: docs/platform/howto/monitoring-services
+          - file: docs/platform/howto/aiven-service-metrics
+          - file: docs/platform/howto/additional-service-metrics
           - file: docs/platform/howto/integrations/prometheus-metrics
           - file: docs/platform/howto/integrations/datadog-increase-metrics-limit
           - file: docs/platform/howto/integrations/access-jmx-metrics-jolokia
--- a/docs/platform/howto/additional-service-metrics.rst
+++ b/docs/platform/howto/additional-service-metrics.rst
@@ -1,0 +1,3907 @@
+Additional Service Metrics
+==========================
+
+Aiven platform provides service metrics integrations to monitor the
+states of your Aiven services. It helps you to quickly get to the bottom
+of performance issues, as well as anticipate and prevent future
+problems.
+
+To help you get started easily, our service integrations enable a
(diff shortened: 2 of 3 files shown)


ftisiot:
Hi @pparmin, can you please have a look at the errors in the build? https://github.com/aiven/devportal/pull/1634/files

lornajane:
Some of the errors are due to using a very old main branch as the origin for this work, we should rebase it.

Any word that is not a dictionary English word should be marked up in code/literal markup, that's what most of the build errors are explaining.

I also have some other higher level questions:
* where was this generated from? We might not want to put the details into the public repository discussion but perhaps we could create some internal docs for the wiki so that we could recreate if necessary to maintain the page.
* Could we put the "all services" docs first?
* Do we still need the Aiven for ElasticSearch metrics listed, since the product isn't available any more?

TibsAtWork:
I think that:
1. There's a lot of Elasticsearch content here, which is not of relevance anymore. Since there's no mention of OpenSearch, I think this means the original text is very old.
2. There are a *lot* of metrics here, and I don't see how we can reasonably check that all of them are useful and/or should be mentioned.
3. If we do want to expose this information, it should probably be autogenerated, so we can actually keep it up-to-date, or else we should work it out again from scratch by consultation with the appropriate backend teams, so that we know the information has some currency.

@staceysalamon-aiven @wojcik-dorota I propose that we close this PR and maybe open an issue to consider the correct approach, if that seems worthwhile.

wojcik-dorota:
> 

@TibsAtWork , thanks for looking this this one. What you're proposing makes sense to me.

staceysalamon-aiven:
Closing per discussion.

PR #1640, opened 2022-12-13
Update title for ClickHouse JDBC

kmichel-aiven:
## What changed, and why it matters

I think we should change the title, the current one look like it's about connecting from ClickHouse to something else, over JDBC.

The ClickHouse JDBC driver is the opposite, it's to connect from a Java app to ClickHouse (and the Java app is not really a source).

Though, since we have the "Connect with Java" article maybe this article is now redundant ?

### DIFF
--- a/docs/products/clickhouse/howto/connect-with-jdbc.rst
+++ b/docs/products/clickhouse/howto/connect-with-jdbc.rst
@@ -1,7 +1,7 @@
-Connect Aiven for ClickHouse® to external databases via JDBC
+Connect to Aiven for ClickHouse® using JDBC
 ===============================================================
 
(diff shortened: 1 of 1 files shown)


wojcik-dorota (changes requested):
The whole idea behind this section is to cover different integration scenarios of Aiven4CH with other services, external DBs, etc. Since you're saying that this article does not cover any integration scenario but only the connection to Aiven4CH service using JDBC, it does not belong to the integration section but to the connection section. And since the connection section already includes the article on connecting with java/jdbc, indeed this one is a duplicate and needs to be removed. Please add the redirection from the removed one to the existing java connection article that we're gonna keep.

jkowall:
Appreciate the idea on this one, but we are going to stick to the current naming.

PR #1643, opened 2022-12-14
Improve wording input-output

laysauchoa:
It is easier for the reader when we use input and output when referring to where the data goes. So this is to fix that.
@dewan-ahmed tagging you on this one

### DIFF
--- a/docs/products/opensearch/howto/import-opensearch-data-elasticsearch-dump-to-aiven.rst
+++ b/docs/products/opensearch/howto/import-opensearch-data-elasticsearch-dump-to-aiven.rst
@@ -25,21 +25,21 @@ Here are some information you need to collect to copy your data:
 
 OpenSearch cluster:
 
-* ``INPUT_SERVICE_URI``: OpenSearch cluster URI, for e.g. ``http://opensearch-url:9243/``
+* ``INPUT_SERVICE_URI``: OpenSearch cluster URI, in the format ``https://user:password@host:port``
 * ``INPUT_INDEX_NAME``: the index that you aim to copy from your input source.
 
 Aiven for OpenSearch:
 
-* ``OUTPUT_INDEX_NAME``: the index that you want to have in your output with the copied data.
 * ``OUTPUT_SERVICE_URI``: your output OpenSearch service URI. You can find it in Aiven's dashboard.
+* ``OUTPUT_INDEX_NAME``: the index that you want to have in your output with the copied data.
 
 .. note::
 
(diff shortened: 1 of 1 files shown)


dewan-ahmed (commented):
Thanks @laysauchoa 

Do you think we can add the `$` on all of `INPUT_INDEX_NAME`, `OUTPUT_INDEX_NAME`, `INPUT_SERVICE_URI`, and `OUTPUT_SERVICE_URI`? 

We can add a sentence to have all these variables exported as environment variables and then the `$` will make sense.

This way, you will be able to directly copy the command during the video or demo.

laysauchoa:
Sure, you can take a look again, thanks! @dewan-ahmed

dewan-ahmed (commented):
Thanks for the changes @laysauchoa 

When I look at an example OpenSearch service, I see the URI in this form: `https://USERNAME:PASSWORD@opensearch-yellow-dev-advocates.aivencloud.com:13039` so `https://USERNAME:PASSWORD@HOST:PORT`

The example you used `http://opensearch-url:9243` is a URL which is in the form of `https://HOST:PORT`. 

Could you please confirm if `elasticdump` is expecting a URI or a URL?

laysauchoa:
> Could you please confirm if elasticdump is expecting a URI or a URL?

**Both.** Because all URLs are URIs, but not all URIs are URLs. An URL is a subtype of URI.

What Aiven uses as ``SERVICE_URI`` is an URL. What ``elasticdump`` refers to is a URL that is still a URI. 

I'll give you an example. All tigers are cats, but not all cats are tigers. So if you adopt a cat using the general concept, you can end up with a tiger at home. Same as URI and URL. If we know this is a URL, we should be referring to it as a URL rather than a URI. I can only think this was chosen as URI to be in the safe side ;)

This is a long and old debate about URI and URL naming which goes beyond this PR, and what I initially proposed to change ``input`` and ``output`` naming. But thanks to your questions, I got to learn a lot about URI's and URL's! :)

Here are a great resource that I hope can be useful for you as well: https://danielmiessler.com/study/difference-between-uri-url/ 

And here are my changes to have a more similar example type of URL in the description: https://github.com/aiven/devportal/pull/1643/commits/809f467f4f06cf1db35c59901c0fde96d84dd10a

dewan-ahmed (approved):
Coincidentally and thanks to SEO, https://danielmiessler.com/study/difference-between-uri-url/ is also the same site I checked when I was learning the differences :)

LGTM and thanks for this update @laysauchoa

PR #1644, opened 2022-12-14
Add missing button

laysauchoa:
Ops, seems like we are missing a button for influxdb docs. Adding it here :)


lornajane (approved):
Good catch!

PR #1646, opened 2022-12-16
clickhouse: add limitations and limits for a service

wojcik-dorota:
Adding info on restrictions on using Aiven for ClickHouse so that the users are aware of the service capabilities and can make educated choices when operating it.

https://aiven.atlassian.net/browse/DOC-61

### DIFF
--- a/docs/products/clickhouse/reference/limitations.rst
+++ b/docs/products/clickhouse/reference/limitations.rst
@@ -6,7 +6,7 @@ This article covers quotas for the Aiven for ClickHouse® service and restrictio
 Limitations
 -----------
 
-From the information about restrictions on using Aiven for ClickHouse, you can easily draw conclusions on how get your service to operate closer to its full potential. Use *Recommneded approach* as guidelines on how to work around specific restrictions.
+From the information about restrictions on using Aiven for ClickHouse, you can easily draw conclusions on how get your service to operate closer to its full potential. Use *Recommended approach* as guidelines on how to work around specific restrictions.
 
 .. list-table::
    :widths: 25 50 25
@@ -19,7 +19,7 @@ From the information about restrictions on using Aiven for ClickHouse, you can e
      - Aiven for ClickHouse service takes a single snapshot a day. Any data inserted before the next snapshot is be lost when nodes recycle.
      - None
    * - Service integrations
-     - You can integrate your Aiven for ClickHouse service with PostgreSQL and Kafka only.
+     - You can integrate your Aiven for ClickHouse service with PostgreSQL® and Kafka® only.
      - None
    * - Table engines availability
      - * Log engine is not supported in Aiven for ClickHouse.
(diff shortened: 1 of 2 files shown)


aris-aiven (changes requested):
Thanks for the the article @wojcik-dorota ! I left two comments.

aris-aiven:
> Looks good to me, the only thing is on the backups. It says data is lost on the recycling of nodes, does that include patches/migrations/scaling?

None of the example you mention @ROConnor-Aiven  are data loss scenarios, they are handled safely and automatically.

aris-aiven:
If anyone has any insights into the Cloudflare Pages build failure feel free to share, not able to get visibility into it.

docs/products/clickhouse/reference/limitations.rst
```diff
@@ -0,0 +1,92 @@
...
+
+.. list-table::
+   :widths: 25 50 25
+   :header-rows: 1
+
+   * - Name
+     - Description
+     - Recommended approach
+   * - Backups
+     - Aiven for ClickHouse service takes a single snapshot a day. Any data inserted before the next snapshot is be lost when nodes recycle.
```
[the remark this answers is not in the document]
kmichel-aiven: I don't think we should focus too much on what happens when you lose all the nodes of a shards, the point of having multiple copies of the data is to not worry about that.

However there are other consequences of only-daily backups that I think are more relevant to users :

- When powering off the service, all data after the last backup is lost.
- We do not provide point in time recovery, we can only restore the database to the state of one of the daily backups.
- When creating a database fork, you can only create a fork that match the state of one of the backups. 

The first one is the most surprising and the most likely cause sadness if missed. The "normal" meaning of power off does not clearly convey the idea that it can cause data loss. (The console warns about it before validating the power off, but it's worth repeating it in the doc).

In any case, we do not lose data when recycling a node :)

I don't think we talk a lot of "recycling nodes" in the console and docs. We mostly say "apply a maintenance update" or "change the service plan". We go into the details of the nodes only if/when we describe how each operation is implemented.

docs/products/clickhouse/reference/limitations.rst
```diff
@@ -0,0 +1,92 @@
...
+       * Some engines are remapped to their `Replicated` alternatives, for example, `MergeTree` -> `ReplicatedMergeTree`.
+     - * For storing data, use the `Buffer engine <https://clickhouse.com/docs/en/engines/table-engines/special/buffer/>`_ instead of the Log engine.
+
+       * Use the available table engines listed in :doc:`Supported table engines in Aiven for ClickHouse </docs/products/clickhouse/reference/supported-table-engines>`.
+   * - Cloud availability
+     - Available on AWS, GCP, and Azure only
+     - Use the available cloud providers.
+   * - Kafka Schema Registry
+     - Aiven for ClickHouse doesn't support Kafka Schema Registry, which allows to build stream processing pipelines with schemas.
+     - Coming soon...
```
kmichel-aiven, line 38: It's too early to promise a Coming soon on that :)

docs/products/clickhouse/reference/limitations.rst
```diff
@@ -0,0 +1,92 @@
...
+   * - Querying all shards at once
+     - If you have a sharded plan, you must use a Distributed view on top of your MergeTree table to query all the shards at the same time, and you should use it for inserts too.
+     - Use the `Distributed` view with sharded plans.
+   * - ON CLUSTER queries
+     - Aiven for ClickHouse doesn't support ON CLUSTER queries because it actually runs each query on all the servers of the cluster without using `ON CLUSTER`.
```
kmichel-aiven, line 43:
```suggestion
     - Aiven for ClickHouse doesn't support ON CLUSTER queries because it actually runs each data definition query on all the servers of the cluster without using `ON CLUSTER`.
```
We run all DDL queries on the cluster (CREATE, ALTER TABLE, etc...) but we don't run SELECT or INSERT on the entire cluster, that would conflict with the previous point of the table.

wojcik-dorota:
> 

@aris-aiven fixed that :-)

aris-aiven (approved):
lgtm!

PR #1656, opened 2022-12-20
Uploaded additional content about high disk usage

bcrisologo:
Uploaded additional content about inactive replication slots causing increase in WAL size, resulting in high disk usage.  Added content on how to resolve this issue and pointing to our setup logical replication slot docs as well.


lornajane (approved):
Thanks for adding this, looks like a useful addition. Extra marks for linking to to the other article as well!

PR #1660, opened 2022-12-27
clickhouse: add a concept on working with string data type

wojcik-dorota:
https://aiven.atlassian.net/browse/DOC-123

Adding a short concept to provide basic info that our users would need to use strings in ClickHouse correctly and handle them efficiently.

### DIFF
--- a/docs/products/clickhouse/concepts/strings.rst
+++ b/docs/products/clickhouse/concepts/strings.rst
@@ -1,0 +1,47 @@
+String data type in Aiven for Clickhouse®
+=========================================
+
+ClickHouse databases can store diverse types of data, such as decimals, booleans, or arrays. This article focuses on strings and its usage in Aiven for ClickHouse. Discover key functions and data type conversions required for working efficiently with strings in Aiven for ClickHouse.
+
+About strings in ClickHouse
+---------------------------
+
+Aiven for ClickHouse allows strings of any length. Strings can contain an arbitrary amount of bytes, which are stored and output as-is. The String type replaces the types VARCHAR, BLOB, CLOB, and others from other DBMSs. When creating tables, numeric parameters for string fields can be set (e.g. TEXT(140)), but are ignored.
+Aiven for ClickHouse supports the following aliases for strings: LONGTEXT, MEDIUMTEXT, TINYTEXT, TEXT, LONGBLOB, MEDIUMBLOB, TINYBLOB, BLOB, VARCHAR, CHAR.
+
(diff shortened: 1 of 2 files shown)


docs/products/clickhouse/concepts/strings.rst
```diff
@@ -0,0 +1,62 @@
...
+More supported data types
+-------------------------
+
+Some other useful data types supported in Aiven for ClickHouse are 
+
+* `Parquet <https://parquet.apache.org/>`_
+    Column-oriented data format, efficient when accessing and reading specific fields and fast when searching for a specific data item
+* `Avro <https://avro.apache.org/>`_
+    Row-based data format (stores all the fields for each record together), used for storing, serializing, and transmitting data over messaging systems (for example, Apache Kafka®) 
+* JSONAsObject
```
kmichel-aiven, line 62: This should not be in this article:
- strings are a type of data for a column in ClickHouse.
- Parquet/Avro/JSONAsObject are data formats for input/output when selecting or inserting data, they are not data types.

wojcik-dorota: Thanks @kmichel-aiven! Two questions:

- Where in the article am I mentioning that "strings are a type of data for a column in ClickHouse" - I'd like to remove that but cannot find it
- I'd remove "Parquet/Avro/JSONAsObject" from this article and perhaps we could add them to the ClickHouse formats article at https://docs.aiven.io/docs/products/clickhouse/reference/supported-input-output-formats - I'd rename this article into more general "ClickHouse-supported formats for data exchange" and include two subchapters: (1) with the existing Kafka info (2) with the Parquet/Avro/JSONAsObject info. We could have this developed in the future. What do you think?

kmichel-aiven: By "this", I meant the entire section about  "More supported data types",  the list of items below was just the details of why this shouldn't be in this article.

I wouldn't mix the data exchange formats for ClickHouse-Kafka integration and the other unrelated formats in the same article. One list is specific to Aiven for ClickHouse and the Kafka integration, and the answer to a frequently asked question about that integration, the other is general info about ClickHouse.

We could make a separate article with all the supported data formats outside of the integration, there are approximately 90 data formats though.

wojcik-dorota: Understood :-) Thanks - removing the subsection!

kmichel-aiven (approved):

PR #1667, opened 2023-01-03
bugfix: Correct grammar/style  in VPC headings

celestehorgan:

Small nonbreaking changes: normalizing the grammar and verbs in the titles of the Platform > HowTo > VPC related topics.

### DIFF
--- a/docs/platform/howto/use-aws-privatelinks.rst
+++ b/docs/platform/howto/use-aws-privatelinks.rst
@@ -1,4 +1,4 @@
-Using AWS PrivateLink with Aiven services
+Use AWS PrivateLink with Aiven services
 =========================================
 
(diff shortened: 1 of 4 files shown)


docs/platform/howto/use-aws-privatelinks.rst
```diff
@@ -1,4 +1,4 @@
-Using AWS PrivateLink with Aiven services
+Use AWS PrivateLink with Aiven services
 =========================================
```
laysauchoa, line 2: to avoid warnings or errors, I'd suggest to have the underline exactly where the sentence ends. So in this case, you need to remove two characters of this `=` similar as you removed in the title.

TibsAtWork (approved):
Looks good to me (following style guide for titles).
There's a separate argument for when/if we expand acronyms in titles, as is done in the final example. but I think that should be separate 😄

PR #1670, opened 2023-01-04
Revert "Bump json5 from 2.2.1 to 2.2.3 in /aws"

lornajane:
Reverts aiven/devportal#1668 since it causes the build to fail.


TibsAtWork (approved):
Seems sensible in context

PR #1675, opened 2023-01-05
Uploaded rst for hourly billing model and updated toc

bcrisologo:
An overview of the hourly billing model for all services doesn't appear to have been migrated over from this [original help article](http://help.aiven.io/en/articles/495327-hourly-billing-model-for-all-services). 

Transferring over that content from intercom (with some edits) and added under the section:

**Platform > Concepts > Billing > Hourly billing model for all services**

Includes updating the `toc` file along with the new `rst` file.

### DIFF
--- a/_toc.yml
+++ b/_toc.yml
@@ -18,6 +18,7 @@ entries:
           - file: docs/platform/concepts/tax-information
           - file: docs/platform/concepts/billing-groups
           - file: docs/platform/concepts/corporate-billing
+          - file: docs/platform/concepts/hourly-billing-model
       - file: docs/platform/concepts/beta_services
       - file: docs/platform/concepts/cloud-security
       - file: docs/platform/concepts/logs-metrics-alerts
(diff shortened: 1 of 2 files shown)


dewan-ahmed:
Hi @lornajane I've reviewed this PR and it's good to merge. Before I merge, two questions:

1. Do I follow the same process of adding this doc to the migration spreadsheet or has something changed since intercom was retired?
2. Because this talks about billing/pricing, do we confirm with @koenigwilliam if this information is still up-to-date and no addition is necessary?

lornajane:
Great questions. Yes, add to the spreadsheet but probably ping the #devportal channel to say that there's an additional redirect needed - I'm not running that process regularly now so we could easily miss the request. It's less time-critical in the sense that we aren't duplicating content - but it would be good to get any old bookmarks forwarding correctly so we should still create the redirect IMO.

The billing information looks accurate to me, we bill hourly.

aprepo:
Looks ok, although we do have some exceptions to the 'all-inclusive plan pricing': for example AWS Privatelink is charged based on network traffic and customer can add DDS (additional storage) to service that is charged per GB on top of the plan price.

dewan-ahmed:
Thanks all for the inputs.

@aprepo do you suggest we add the exception as a note in this article?

bcrisologo:
I can put a note that says something along the lines of: 

> _Note: Use of PrivateLink and/or adding additional storage will incur additional costs on top of the hourly service usage rate._

If there's a better language for it, do let me know and I'll add it per recommendation.

PR #1678, opened 2023-01-06
Add instructions on creating orgs and org units

staceysalamon-aiven:
With the release of organizations and organizational units, accounts will be replaced with organizations in Aiven. This PR updates the article on creating accounts with instructions on creating organizations and organizational units.

These features are planned for release in Q1.

### DIFF
--- a/docs/tools/aiven-console/howto/create-accounts.rst
+++ b/docs/tools/aiven-console/howto/create-accounts.rst
@@ -7,7 +7,7 @@ Create organizations and org units
 Create an organization
 --------------------------
 
-#. At the top right side of the page, click the organization name and select **Create organization**. 
+#. At the top right side, click the organization name and select **Create organization**. 
 
    If you don't have an organization yet, click **Create organization**. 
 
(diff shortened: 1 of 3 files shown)


docs/tools/aiven-console/howto/create-accounts.rst
```diff
@@ -1,23 +1,49 @@
...
+==============================================
 
-**Accounts** are groups of projects with common settings like authentication and teams (user groups).
+**Organizations** and **organizational units** (or **unit**) can be used to group projects and apply common settings like authentication and teams (user groups). For details and recommendations on creating hierarchical organizations in Aiven, see :doc:`Organizations, projects, and managing access permissions </docs/platform/concepts/projects_accounts_access>`.
 
-Create a new account
-------------------------------
 
-#. At the top right side of the page, click the account name and select **Create account**. 
+Create an organization
```
rsalevsky, line 7: We should add here that we don't recommend more than one organization per customer but rather the usage of Units. We should also mention here what all is managed per organization.

- Billing Groups
- Authentication
- Team

Another org would mean all this needs to be resetup.

rsalevsky (approved):

PR #1682, opened 2023-01-10
Update eol for OS 1.x

hoang-vo:

EOL date is introduced in upstream OpenSearch. https://opensearch.org/releases.html

We follow the same eol for Aiven for opensearch


harshini-rangaswamy (approved):
Looks good.

PR #1687, opened 2023-01-10
mentioning kafka rebalancing

zhangmmaiven:

In Kafka, we manage/balance partitions for the customers. This is automated and they don't have to worry about it whatsoever. I don't think this is documented publicly on the dev portal.

Mentioned kafka topic rebalancing in the Apache Kafka / Concepts / Partition segments, added a note.


ftisiot (approved):
LGTM

PR #1694, opened 2023-01-13
Patched()Fix Prototype Pollution in JSON5 via Parse Method

imhunterand:
The parse method of the JSON5 library before and including version 2.2.1 does not restrict parsing of keys named `__proto__`, allowing specially crafted strings to pollute the prototype of the resulting object.

This vulnerability pollutes the prototype of the object returned by JSON5.parse and not the global Object prototype, which is the commonly understood definition of Prototype Pollution. However, polluting the prototype of a single object can have significant security impact for an application if the object is later used in trusted operations.

### DIFF
--- a/aws/package-lock.json
+++ b/aws/package-lock.json
@@ -4847,9 +4847,9 @@
       "dev": true
     },
     "node_modules/json5": {
-      "version": "2.2.1",
-      "resolved": "https://registry.npmjs.org/json5/-/json5-2.2.1.tgz",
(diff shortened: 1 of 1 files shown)


TibsAtWork:
See https://github.com/json5/json5/security/advisories/GHSA-9c47-m6qq-7p4h

TibsAtWork:
@hieu-aiven and @angelinekwan are you able to comment on this?
It looks as if the packages using json5 may also have been updated (e.g., opensearch 1.3 uses a fixed json5), so would it be preferable to update the `package.json` file?

Issue #1691, opened 2023-01-11
Canonical tags ending with .html creates redirect-canonical loop and makes the pages non-indexable


angelinekwan:
# What's wrong?
In all pages, the canonical tags of [docs.aiven.io](http://docs.aiven.io/) are pointing to redirected URL versions ending with .html. For example, https://docs.aiven.io/docs/tools/api has a canonical to https://docs.aiven.io/docs/tools/api.html. This creates a redirect-canonical loop and makes the pages non-indexable. Can you, please, remove those canonical tags or replace them by the version without .html? So the canonical in the example should be https://docs.aiven.io/docs/tools/api

# URL of affected page, and any other information
Issue in all pages. This issue appeared around the holidays.

PR #1696, opened 2023-01-16
Custom canonical ul without html extension

angelinekwan:
In all pages, the canonical tags of [docs.aiven.io](http://docs.aiven.io/) are pointing to redirected URL versions ending with .html. For example, https://docs.aiven.io/docs/tools/api has a canonical to https://docs.aiven.io/docs/tools/api.html. This creates a redirect-canonical loop and makes the pages non-indexable. The canonical tags need to be the version without .html. So the canonical in the example should be https://docs.aiven.io/docs/tools/api

Resolving this [issue](https://github.com/aiven/devportal/issues/1691)

# Solution
Instead of using default [html_baseurl](https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-html_baseurl), set custom canonical tag in base template to have control of the `pagename`.

# Demo
https://b0372b27.devportal.pages.dev/docs/tools/api
Screenshot 2023-01-16 at 12 11 07

### DIFF
--- a/_redirects
+++ b/_redirects
@@ -11,42 +11,42 @@
 /grafana            /docs/products/grafana
 
 /api                /docs/tools/api
-/cli                /docs/tools/cli.html
+/cli                /docs/tools/cli
 /terraform          /docs/tools/terraform
 
 # Renamed/deleted files
-/docs/products/flink/howto/real-time-alerting-solution-cli.html       /docs/products/flink/howto/real-time-alerting-solution.html
+/docs/products/flink/howto/real-time-alerting-solution-cli.html       /docs/products/flink/howto/real-time-alerting-solution
 /docs/products/flink/howto/real-time-alerting-solution-cli            /docs/products/flink/howto/real-time-alerting-solution
-/docs/platform/howto/disable-user-2fa.html                            /docs/platform/howto/user-2fa.html
+/docs/platform/howto/disable-user-2fa.html                            /docs/platform/howto/user-2fa
 /docs/platform/howto/disable-user-2fa                                 /docs/platform/howto/user-2fa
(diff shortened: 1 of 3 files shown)


lornajane:
I'm confused about why this PR doesn't have a preview URL for us to check?

angelinekwan:
> I'm confused about why this PR doesn't have a preview URL for us to check?
Not sure neither. I got the link from the Checks earlier but is now missing as well. I added preview url in the PR description under Demo. https://b0372b27.devportal.pages.dev/

lornajane:
I can see where we are going with this but the html_baseurl configuration is used by the sitemap, which now doesn't show URLs. I think we should try not to break the sitemap when fixing the canonical URLs.

angelinekwan:
> I can see where we are going with this but the html_baseurl configuration is used by the sitemap, which now doesn't show URLs. I think we should try not to break the sitemap when fixing the canonical URLs.

You are right - checking from the [sitemap](https://b0372b27.devportal.pages.dev/sitemap.xml) the links still have `.html` in the page url and missing domain.

I am not aware where else in the sphinx build uses [html_baseurl](https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-html_baseurl), as the documentation only mentioned ->  It is used to indicate the location of document using [The Canonical Link Relation](https://datatracker.ietf.org/doc/html/rfc6596).

Do you have other suggestion to tackle this issue?
I tried to remove .html extension on sphinx-build. [Reference](https://www.sphinx-doc.org/en/master/man/sphinx-build.html)
`-b dirhtml` -> Build HTML pages, but with a single directory per document. Makes for prettier URLs (no .html) if served from a webserver but it didn't work.

PR #1698, opened 2023-01-16
updated the Flink integration topic

harshini-rangaswamy:

Updated the Flink Integration topic: 
 - Updated the title to "Create Apache Flink® data service integrations" to avoid confusion with other Aiven and external integrations. 
 - Updated screenshots (most like, the screenshots might need to updated again as Jobs and data tab will be removed)


ftisiot (approved):
LGTM

Issue #1697, opened 2023-01-16
Fix various `make spell` issues


TibsAtWork:
Running `make spell` locally, I get 59 errors. A casual glance at the last few shows they are indeed things that need fixing. This issue is for fixing them. (there's a variety of reStructuredText and/or dictionary problems, for instance).

I've assigned this to me, and intend to work on it today/tomorrow, but if someone else wants to take it over, let me know.

**Note** it also looks as if we need to start being more sophisticated with our dictionary definitions (to allow possessives/plurals).

*I am working on this*

PR #1700, opened 2023-01-16
Fix various `make spell` issues, and make the dictionary a little cleverer

TibsAtWork:
Fix a cluster of errors revealed on my local machine by `make spell`.

Fix: https://github.com/aiven/devportal/issues/1697

In particular:
* Start using `/S` (plural) and `/M` (possessive) markers in the dictionary. This also means putting some rules into the `.aff` dictionary file. I haven't tried to find all the places where we might use `/S` in the dictionary - there are probably others.
* Fix some odd cases where Vale was complaining about not enough ® marks, even though they already appeared to be sufficient when just looking at the file. In particular, add extra ® marks to `docs/tools/cli/service/flink.rst` beyond what Vale required, so that the text is self-consistent in how it uses them.
* Fix some cases of incorrect reStructuredText

I was not quite sure whether to keep this as one PR, or whether to split out the dictionary-related changes. Please ask if you'd prefer it split up into two or more PRs.

### DIFF
--- a/docs/products/postgresql/reference/terminology.rst
+++ b/docs/products/postgresql/reference/terminology.rst
@@ -18,8 +18,8 @@ Write-Ahead Log (WAL)
 
 .. _Terminology PGLookout:
 
-PGLookout
-    `PGLookout <https://github.com/aiven/pglookout>`_ is a PostgreSQL replication monitoring and failover daemon.
+pglookout
+    `pglookout <https://github.com/aiven/pglookout>`_ is a PostgreSQL replication monitoring and failover daemon.
 
 .. _Terminology PGHoard:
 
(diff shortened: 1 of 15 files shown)


docs/products/postgresql/reference/terminology.rst
```diff
@@ -18,8 +18,8 @@ Write-Ahead Log (WAL)
 
 .. _Terminology PGLookout:
 
-PGLookout
-    `PGLookout <https://github.com/aiven/pglookout>`_ is a PostgreSQL replication monitoring and failover daemon.
+pglookout
+    `pglookout <https://github.com/aiven/pglookout>`_ is a PostgreSQL replication monitoring and failover daemon.
```
TibsAtWork, line 22: Note that the documentation on the GitHub page for pglookout does not capitalise the name at all

celestehorgan (approved):
The only (non-blocking) issue I see is that from a stylistic POV it's generally acceptable to use ®, tm, etc. in the first instance in a given document and drop it in subsequent instances.

That's a maybe-later task for all of us, on the whole this LGTM. I'll approve!

PR #1703, opened 2023-01-17
Remove reference to unnecessary help article on old service deletion

TibsAtWork:
The help article on deleting old services didn't actually add any new information, so the simplest thing is just to stop that text from being a link.


ftisiot (approved):
LGTM

PR #1710, opened 2023-01-20
Flink create kafka based flink table

harshini-rangaswamy:

Edited the topic "Create an Apache Kafka®-based Apache Flink® table"  to include updated steps and changes to flow based on the Flink application.


ftisiot (changes requested):
I believe all the section of table parameters is in the TABLE SQL now, we should provide an example of the SQL for insert and for upsert so people could copy/paste

ftisiot (approved):
LGTM

PR #1712, opened 2023-01-20
Edit content in OpenSearch®-based Apache Flink® table topic

harshini-rangaswamy:

Edit the content for the topic - Create an OpenSearch®-based Apache Flink® table based on the new Flink Application and table.


ftisiot (changes requested):
Also here, we should showcase a SQL example of the definition with all the possible parameters

ftisiot (approved):
LGTM

PR #1713, opened 2023-01-22
Added slack connector topic

harshini-rangaswamy:

Added a topic for information about using Slack connectors for sink tables.

### DIFF
--- a/docs/products/flink/howto/slack-connector.rst
+++ b/docs/products/flink/howto/slack-connector.rst
@@ -8,7 +8,7 @@ You can access the open-source Slack connect for Apache Flink on Aiven's GitHub 
 Prerequisites
 -------------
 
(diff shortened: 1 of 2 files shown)


docs/products/flink/howto/slack-connector.rst
```diff
@@ -0,0 +1,45 @@
...
+5. In the **Table SQL** section, set the connector to **slack** and enter the necessary token as shown in the example below:
+
+::
+
+    CREATE TABLE channel_name (
+    channel_id STRING,
+    message STRING
+    ) WITH (
+        'connector' = 'slack',
+        'token' = 'your_token_here'
```
ftisiot, line 40: We should write that the `channel_id` is the channel ID parameter and the token is the token value

ftisiot (changes requested):
Almost good to go, I would add a note to substitute the parameters

docs/products/flink/howto/slack-connector.rst
harshini-rangaswamy: Fixed.

ftisiot (approved):
LGTM

PR #1720, opened 2023-01-24
contributing: add reference to google docs style guide

wojcik-dorota:
Adding a reference to the Google developer documentation style guide for the contributors to devportal - style guidelines to be followed if our Aiven-specific rules dont cover a specific stylistic problem.

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -67,11 +67,11 @@ The instructions for setting up local development are in the ``README``.
 Style guide
 -----------
 
-These are the guidelines that we use for Aiven Docs; following them helps us keep all the content consistent and easy to follow for our audience. We review your pull request against these guidelines, so for fewer rounds of revisions, follow this advice!
+To keep all the content consistent and easy to follow for our audience, use `Google developer documentation style guide <https://developers.google.com/style>`_ for Aiven docs. On top of that, this section provides you with a few Aiven-specific guidelines you're requested to follow.
 
-.. note::
+.. tip::
 
-    For any style-related challenges not covered in this guide, `Google developer documentation style guide <https://developers.google.com/style>`_ applies.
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.rst
```diff
@@ -69,6 +69,10 @@ Style guide
 
 These are the guidelines that we use for Aiven Docs; following them helps us keep all the content consistent and easy to follow for our audience. We review your pull request against these guidelines, so for fewer rounds of revisions, follow this advice!
 
+.. note::
```
staceysalamon-aiven, line 72: Would it be better to instead put this in the introduction? My thought is that the introduction would be revised to explain that the Google developer documentation style guide is the default and that exceptions/gaps are detailed in the rest of the style guide.

staceysalamon-aiven (commented):
Thank you for doing this! I left a comment with a suggestion that we incorporate this into the introduction.

CONTRIBUTING.rst
wojcik-dorota: Thanks @staceysalamon-aiven :-) It's an option definitely. Still, if you mean the into to the CONTRIBUTING file, this doc covers a wide range of areas, style guide being only one of them. And an introduction seems to stay a general part for providing info regarding the whole thing, not addressing a specific section. From my pov, style guide info belongs to the style guide chapter, which is actually there, so why not use it. @harshini-rangaswamy , any thought on that?

staceysalamon-aiven: Yeah, so I meant the introduction to the style guide, not the whole contributing file. 

image

harshini-rangaswamy (approved):
Looks good to me with the info about the style guide as a part of the introduction.

staceysalamon-aiven (approved):
LGTM

PR #1732, opened 2023-01-27
cassandra: enable cross-cluster replication

wojcik-dorota:
https://aiven.atlassian.net/browse/DOC-163


docs/products/cassandra/howto/enable-cross-cluster-replication.rst
```diff
@@ -0,0 +1,174 @@
...
+1. Log in to the `Aiven console <https://console.aiven.io/>`_.
+2. From the **Current services** view, select **+Create service**.
+3. In the **Create service** view
+
+   1. Select Apache Cassandra® and, next, a cloud provider and a cloud region for your service.
+   2. To activate CCR, select the **Enable Cross Cluster Replication** toggle, which unfolds an additional section for setting up another Apache Cassandra service.
+   3. Select a cloud provider and a cloud region for the second service.
+   4. Select one service plan for your two new services.
+   5. Optionally, add an additional disk storage for your services.
+   6. Enter names for your services.
```
dmitry-potepalov, line 46: maybe mention the datacenter names somewhere? (at least that you see them in the UI at this moment, so you could use those when executing actual statements?)

docs/products/cassandra/howto/enable-cross-cluster-replication.rst
```diff
@@ -0,0 +1,174 @@
...
+            }
+         }'
+
+Related reading
+---------------
+
+* :doc:`OpenSearch® cross-cluster replication</docs/products/opensearch/concepts/cross-cluster-replication-opensearch>`
+* :doc:`Set up cross-cluster replication for OpenSearch</docs/products/opensearch/howto/setup-cross-cluster-replication-opensearch>`
+* :doc:`Enabling cross-cluster replication for Apache Kafka® via Terraform</docs/tools/terraform/reference/cookbook/kafka-mirrormaker-recipe>`
+* `Cassandra® documentation <https://cassandra.apache.org/doc/latest/>`_
```
dmitry-potepalov, line 174: IMHO too generic, maybe this would be better:
https://cassandra.apache.org/doc/latest/cassandra/architecture/dynamo.html#multi-master-replication-versioned-data-and-tunable-consistency

docs/products/cassandra/howto/enable-cross-cluster-replication.rst
```diff
@@ -0,0 +1,174 @@
...
+
+.. code-block:: bash
+
+   curl --request PUT                                                                     \
+      --url https://api.aiven.io/v1/project/YOUR_PROJECT_NAME/service/YOUR_SERVICE_NAME   \
+      --header 'Authorization: Bearer YOUR_BEARER_TOKEN'                                  \
+      --header 'content-type: application/json'                                           \
+      --data
+         '{
+            "user_config": {
```
dmitry-potepalov, line 163: datacenter is necessary here as well, otherwise you risk getting the 400 because it's `aiven` by default

docs/products/cassandra/howto/enable-cross-cluster-replication.rst
```diff
@@ -0,0 +1,174 @@
...
+   3. Select a cloud provider and a cloud region for the second service.
+   4. Select one service plan for your two new services.
+   5. Optionally, add an additional disk storage for your services.
+   6. Enter names for your services.
+
+4. Verify your choices in the **Services summary** card on the right and, if your configuration turns out as expected, select **+Create services** if the setting looks.
+
+.. topic:: Result
+   
+   You've created two new services that are connected for CCR purposes. You can preview CCR details for your services in the **Overview** tab under **Cross Cluster replication**.
```
dmitry-potepalov, line 52: From the Aiven side of things that's it, the cluster is CCR-ready; but for the replication to actually work, the user also needs to create a keyspace with correct replication settings; probably a brief mention of this fact + a link to the concept article? otherwise it sounds like "yay, you're done, now everything replicates automagically"

docs/products/cassandra/howto/enable-cross-cluster-replication.rst
```diff
@@ -0,0 +1,174 @@
...
+On an existing service
+''''''''''''''''''''''
+
+1. Log in to the `Aiven console <https://console.aiven.io/>`_.
+2. From the **Current services** view, select an Aiven for Apache Cassandra service on which you'd like to enable CCR.
+3. In the **Overview** tab of your service's page, navigate to **Cross Cluster Replication** and select **Migrate to Cross Cluster**.
+
+   .. note::
+      
+      When you enable CCR on a particular service, you connect this service to a new service that comes with the same plan, number of nodes, and disk storage.
```
dmitry-potepalov, line 63: > you connect this service to a new service

maybe just "create a new service/create another service with the same ..."? probably that's more important here that you're _creating_ a new service "behind the scenes" (well, not really _behind_, it's right there in the UI, but still), which might not be initially obvious and costs $$

docs/products/cassandra/howto/enable-cross-cluster-replication.rst
```diff
@@ -0,0 +1,174 @@
...
+You can enable CCR for your new or existing Aiven for Apache Cassandra service using the Aiven CLI.
+
+.. tip::
+
+   Check out how to get started with the Aiven CLI in :doc:`Aiven CLI </docs/tools/cli>`.
+
+On a new service
+''''''''''''''''
+
+Use the :ref:`avn service create <avn-cli-service-create>` command to create a new service and add the ``service_to_join_with`` parameter. The parameter indicates a service to which you'd like to connect to enable CCR and create a CCR service pair.
```
dmitry-potepalov, line 86: > The parameter indicates a service

indicates _how_? probably you meant that the user should supply a name of an existing service in the same project?

docs/products/cassandra/howto/enable-cross-cluster-replication.rst
```diff
@@ -0,0 +1,174 @@
...
+
+.. code-block:: bash
+
+   avn service create                                   \
+      --service-type cassandra                          \
+      --cloud cloud_region_name                         \
+      --plan service_plan_name                          \
+      -c cassandra.datacenter=datacenter_name           \
```
dmitry-potepalov, line 94: also the text above misses the fact that the datacenter needs to be different. Maybe add a common section that explains the contents of service_to_join_with and cassandra.datacenter and link to it from all the recipes? in order not to repeat it again and again

dmitry-potepalov (changes requested):
overall looks nice, left some suggestions in comments. Also you should probably squash the commits before submitting? commit history like "draft/fix/fix/fix" does not add much value for posterity, IMHO :)

conf.py
```diff
@@ -298,6 +298,8 @@
 
 .. |beta| replace:: :bdg-secondary:`beta`
 
+.. |private beta| replace:: :bdg-secondary:`private beta`
```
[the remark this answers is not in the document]
wojcik-dorota: This is what I found from Serhat a while ago. @divdby0 could you confirm we need we need ``private beta`` or ``beta``?

docs/products/cassandra/howto/enable-cross-cluster-replication.rst
```diff
@@ -0,0 +1,174 @@
...
+.. important::
+
+   This private beta feature is available on request only. Contact sales@Aiven.io or support@Aiven.io to opt in. 
+
+This article provides you with instructions on how to set up the cross-cluster replication (CCR) in your Aiven for Apache Cassandra® service using the Aiven console.
+
+About enabling CCR
+------------------
+
+You can enable CCR either when creating a new Aiven for Apache Cassandra service or on an existing service. For that purpose, you can use the following tools:
```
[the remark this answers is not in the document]
wojcik-dorota: yes, that makes sense. How about adding the note below for the time being, and later on, when the concept is available, adding a link inside this note leading to the relevant info in the concept?
`Note: Enabling CCR on an existing service can affect your productions workload since CCR imposes restrictions on replication settings of the user's keyspace. Make sure you understand the impact before you enable CCR on your service`

conf.py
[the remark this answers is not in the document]
wojcik-dorota: @dmitry-potepalov , I checked with @divdby0 and his conclusion is `beta` after all :-) so updating accordingly, thanks!

docs/products/cassandra/howto/enable-cross-cluster-replication.rst
```diff
@@ -0,0 +1,199 @@
...
+.. For the other part (the configuration of the replication factor on the Apache Cassandra side), which ultimately makes the replication work, see the instruction in **Set up the replication factor**.
+
+About enabling CCR
+------------------
+
+You can enable CCR either when creating a new Aiven for Apache Cassandra service or on an existing service.
+
+.. note::
+
+   Enabling CCR on an existing service is only possible if the service has ``NetworkTopologyStrategy`` set as a replication strategy for the keyspace.
```
dmitry-potepalov, line 15: > for the keyspace

`for all distributed keyspaces` would be better, since the user can have more than one keyspace

docs/products/cassandra/howto/enable-cross-cluster-replication.rst
```diff
@@ -0,0 +1,199 @@
...
+You can enable CCR for your new or existing Aiven for Apache Cassandra service using CLI.
+
+.. note::
+   
+   In this instruction, the :doc:`Aiven CLI client </docs/tools/cli>` is used to interact with Aiven APIs.
+
+.. topic:: Understand parameters to be supplied
+
+   * ``service_to_join_with`` parameter value needs to be set to a name of an existing service in the same project. The supplied service name indicates the service you connect to for enabling CCR. The two connected services create a CCR service pair.
+   * ``cassandra.datacenter`` parameter value needs to be set to a name of a datacenter for your service. Make sure each of the two service constituting a CCR pair belongs to a different datacenter.
```
dmitry-potepalov, line 99: > belong to a different datacenter

this section kind of hints that the datacenter name is supposed to be obtained from the infra or the API, while it's really the user who's supposed to make a choice here... maybe write something like:

`cassandra.datacenter` is the datacenter name used to identify nodes from a particular service in the cluster's topology. In Aiven for Cassandra CCR, nodes of either of the two services belong to a single datacenter, therefore this value needs to be unique for each service. We suggest to set it equal to the service name. You will need this value later on when setting up replication for your keyspaces.

docs/products/cassandra/howto/enable-cross-cluster-replication.rst
```diff
@@ -0,0 +1,199 @@
...
+.. important::
+
+    * Your existing service has ``NetworkTopologyStrategy`` set as a replication strategy for its keyspace.
+    * Your existing service and the service you connect to (``service_to_join_with``) are hosted on different datacenters.
+
+Use the :ref:`avn service update <avn-cli-service-update>` command to modify your service configuration by adding the ``service_to_join_with`` parameter and set its value as needed.
+
+.. code-block:: bash
+
+   avn service update service_name                     \
```
dmitry-potepalov, line 128: I think there's a misunderstanding about what it means to enable CCR on an existing service. It means "I have a service X, I'm creating a new empty service Y and setting it up to join with X (replicate from X)". That's what happens if you do this in the UI. 

On the other hand, the command you've added here means "I have a service X, with some data, and a service Y, with probably some other data; now I command Y to join X (start replicating from X)". This setup is not supported and is supposed to give you an error; if this command has not errored out, it's a bug.

dmitry-potepalov: maybe it would be better if we go "Create a new CCR service pair/Add a CCR peer in another region to an existing service" instead of "Enable CCR for a new service/existing service", this would stress the fact that there's two services involved

dmitry-potepalov (changes requested):
I see one major issue: the "enable-on-existing-service" with CLI/API chapters seem to cover a different scenario from what happens when doing this in the Console

docs/products/cassandra/howto/enable-cross-cluster-replication.rst
wojcik-dorota: Thanks, Dmitry :-) Does it mean that for the CLI method:
1. For the "Create a new CCR service pair" scenario: We need run two commands - (1) create a regular new service (2) create another new service with `service_to_join_with=service_1_name` specified?
2. For the "Add a CCR peer in another region to an existing service" scenario: We need only one command - create a new service with `service_to_join_with=existing_service_name` specified?

wojcik-dorota: Also, am I right I need to same approach for the API method instruction? @dmitry-potepalov

wojcik-dorota:
> I see one major issue: the "enable-on-existing-service" with CLI/API chapters seem to cover a different scenario from what happens when doing this in the Console

comments addressed :-) @dmitry-potepalov

docs/products/cassandra/howto/enable-cross-cluster-replication.rst
```diff
@@ -0,0 +1,278 @@
...
+You can enable CCR either when creating a new Aiven for Apache Cassandra service or for an existing service.
+
+.. _limitations:
+
+Limitations
+'''''''''''
+
+* Enabling CCR on an existing service is only possible if this service has ``NetworkTopologyStrategy`` set as a replication strategy for all distributed keyspaces.
+* Each of the services in the CCR pair needs to be hosted on a different datacenter.
+* Two CCR peer services need to share one service plan and the same amount of dynamic disk space.
```
alanfranz, line 20: nitpicking: what does "sharing" a service plan mean? I'd say something like "use an identical plan"

alanfranz (approved):
Mostly good to me. You choose what to do with the "sharing" terminology.

dmitry-potepalov (approved):
LGTM

divdby0 (approved):

PR #1742, opened 2023-02-01
PostgreSQL and Redis tutorial

TibsAtWork:
Pre-draft - just setting up the files.

This is (will be) a tutorial on adding caching to a PostgreSQL® webapp using Redis®*, with some explanation of why this is a useful thing to do.

### DIFF
--- a/docs/tutorials/add-caching-to-pg-with-redis.rst
+++ b/docs/tutorials/add-caching-to-pg-with-redis.rst
@@ -43,17 +43,29 @@ least Python 3.7
           support 3.6, but ``fastapi`` requires at least 3.7, and honestly
           that's already quite an old version of Python!
 
-* CLI tooling: The [Redis CLI](https://redis.io/docs/ui/cli/) for Redis and [Psql](https://www.geeksforgeeks.org/postgresql-psql-commands/) for PostgreSQL are useful to know as they're transferrable anywhere you go. We built the [avn CLI](https://docs.aiven.io/docs/tools/cli) to take advantage of all the features Aiven offers for its products, and this works too! We'll provide examples with both in this tutorial. 
+* CLI tooling: The [Redis CLI](https://redis.io/docs/ui/cli/) for Redis and
+  [Psql](https://www.geeksforgeeks.org/postgresql-psql-commands/) for
+  PostgreSQL are useful to know as they're transferrable anywhere you go. We
+  built the [avn CLI](https://docs.aiven.io/docs/tools/cli) to take advantage
+  of all the features Aiven offers for its products, and this works too! We'll
+  provide examples with both in this tutorial.
 
-If you're following along without using Aiven, we still recommend deploying to a cloud provider like AWS or Google Cloud. This tutorial assumes the databases will be configured and deployed for you like Aiven does, and starts at the point where we connect to a running service.
+If you're following along without using Aiven, we still recommend deploying to
+a cloud provider like AWS or Google Cloud. This tutorial assumes the databases
+will be configured and deployed for you like Aiven does, and starts at the
+point where we connect to a running service.
 
 
 Set up a Python virtual environment
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-We'll do our development at the command line, in a Python virtual environment. This will prevent any of the work we're doing in this tutorial from affecting anything else you might be working on.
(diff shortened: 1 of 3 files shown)


docs/tutorials/add-caching-to-pg-with-redis.rst
```diff
@@ -0,0 +1,159 @@
+Tutorial: Add caching to your PostgreSQL® app using Redis®*
+===========================================================
+
+
+Learning objectives
+-------------------
+
+* Creating a simple Python web application, that accesses data in a PostgreSQL database
```
celestehorgan, line 8: Are we specifically doing a cache to improve read performance, or are we addressing write performance at any point in this tutorial?

TibsAtWork: Since read caching is what I have worked with, and it's easy to explain, that's all I'd expected to address in the tutorial.

Write caching (and not losing the data if the cache goes away?) isn't something I really know about.

(But I'm more than willing to defer to more knowledge!)

docs/tutorials/add-caching-to-pg-with-redis.rst
```diff
@@ -0,0 +1,174 @@
...
+
+As a result of this, improving read performance gives us a far greater increase in overall performance for our work. The tried and tested way to do this is through using a cache. This becomes especially relevant when we start developing applications in the cloud. When we start making our applications highly available – that is, duplicating various services, like our databases in multiple regions globally to manage speed, and within regions themselves, to manage traffic spikes – we suddenly have data that needs to be read and written to internally, by the application itself, to maintain data integrity, in addition to externally, to be served to our users so they can do things.
```
TibsAtWork, line 10: Love all of this - it gives good motivation for why we're bothering with this tutorial

docs/tutorials/add-caching-to-pg-with-redis.rst
```diff
@@ -0,0 +1,174 @@
...
+
+Caching is the act of writing to a block of memory specifically designed for quick retrievals (reads) of common requests. In traditional hardware terms, the cache is typically a memory chip with particularly fast read and write access, like RAM. The computer uses a cache it kind of like a whiteboard: when it needs to, it writes a small amount of information to the cache quickly for a specific set of task, like retrieving a customer profiles. This lets the application access the customer's profile quickly for the many times it's needed during checkout, while only making the expensive and potentially slow call to the database once. The differences are mere milliseconds, but on the scale of global computing, those add up quickly!
```
TibsAtWork, line 12: I like the whiteboard analogy, and the emphasis on "small" bits of data.

I'm not sure whether we should mention hardware caching, though - it's introducing another idea that's a diversion from our main topic (I *like* diversions, mind you) - I don't expect most readers to care about hardware these days.

On the other hand, I like that last sentence about how small differences add up quickly.

docs/tutorials/add-caching-to-pg-with-redis.rst
```diff
@@ -0,0 +1,174 @@
...
+
+However, we're developing applications for the cloud, not directly onto our computer. This adds a few layers of complexity, but opens up some interesting opportunities to optimize our performance further. Instead of using a block of memory (and the unbounded, chaotic nature that entails), we can use two databases instead! We can use one database as a data store, and one as a cache. This lets us optimize our data store for things like concurrency control and our cache for speedy reads and writes, while still taking advantage of everything the cloud offers us in terms of scalability.
```
TibsAtWork, line 14: Yes, this is good.

Perhaps we could draw out more that some databases (like PostgreSQL) are aimed at scalability, being able to handle large amounts of data - so they're a good choice for the actual datastore - whereas others (like Redis) are aimed at speed, but to do that they sacrifice the ability to scale as large.

TibsAtWork (commented):
I do like your adding in a "why this is worth doing" section at the start - motivation is important.

docs/tutorials/add-caching-to-pg-with-redis.rst
TibsAtWork, line 12: I like the whiteboard analogy, and the emphasis on "small" bits of data.

I'm not sure whether we should mention hardware caching, though - it's introducing another idea that's a diversion from our main topic (I like diversions, mind you) - I don't expect most readers to care about hardware these days.

On the other hand, I like that last sentence about how small differences add up quickly.

PR #1747, opened 2023-02-02
Added a blurb about max users for servics. (SUP-436)

josh-martin-aiven:
Added a blurb to platform/howto/create_new_service_user to inform customers of the default max user limit, and provide guidance if they want to increase their limit.

### DIFF
--- a/docs/platform/howto/create_new_service_user.rst
+++ b/docs/platform/howto/create_new_service_user.rst
@@ -4,13 +4,11 @@
 Service users are users that only exist in the scope of the corresponding Aiven service. They are unique to this service and not shared with any other services, and can be granted restricted permissions compared to the default ``avnadmin`` user. You can add service users for all the Aiven services, with the exception of Aiven for Apache Flink® and Aiven for Grafana®.
 
 .. warning::
-   By default, the maximum amount of users allowed on a service is 50. If a service user is created that exceeds the max users, the following error will be produced.
+   By default, the maximum amount of users allowed on a service is 50. 
    
-   .. code-block:: python 
       
-      "Service user 'xxx' cannot be created, maximum number of users for this plan exceeded"
 
-   If you would like to increase the maximum amount of users on a service, please reach out to the Aiven support team to request an increase.
+   If you would like to increase the maximum number of users allowed for a service, :doc:`create a support ticket </docs/platform/howto/project-support-center>` to request an increase.
 
 1. Log in to the `Aiven console <https://console.aiven.io/>`_.
 
--- a/docs/platform/howto/create_new_service_user.rst
+++ b/docs/platform/howto/create_new_service_user.rst
@@ -3,6 +3,15 @@
 
 Service users are users that only exist in the scope of the corresponding Aiven service. They are unique to this service and not shared with any other services, and can be granted restricted permissions compared to the default ``avnadmin`` user. You can add service users for all the Aiven services, with the exception of Aiven for Apache Flink® and Aiven for Grafana®.
 
+.. warning::
+   By default, the maximum amount of users allowed on a service is 50. If a service user is created that exceeds the max users, the following error will be produced.
+   
+   .. code-block:: python 
+      
+      "Service user 'xxx' cannot be created, maximum number of users for this plan exceeded"
+
+   If you would like to increase the maximum amount of users on a service, please reach out to the Aiven support team to request an increase.
+
 1. Log in to the `Aiven console <https://console.aiven.io/>`_.
 
 2. On the **Services** page, click on the service name.


docs/platform/howto/create_new_service_user.rst
```diff
@@ -3,6 +3,15 @@
 
 Service users are users that only exist in the scope of the corresponding Aiven service. They are unique to this service and not shared with any other services, and can be granted restricted permissions compared to the default ``avnadmin`` user. You can add service users for all the Aiven services, with the exception of Aiven for Apache Flink® and Aiven for Grafana®.
 
+.. warning::
+   By default, the maximum amount of users allowed on a service is 50. If a service user is created that exceeds the max users, the following error will be produced.
```
staceysalamon-aiven, line 7:
```suggestion
   By default, the maximum amount of users allowed on a service is 50. 
```

docs/platform/howto/create_new_service_user.rst
```diff
@@ -3,6 +3,15 @@
...
+   
+   .. code-block:: python
```
staceysalamon-aiven, line 9:
```suggestion

```

docs/platform/howto/create_new_service_user.rst
```diff
@@ -3,6 +3,15 @@
 
 Service users are users that only exist in the scope of the corresponding Aiven service. They are unique to this service and not shared with any other services, and can be granted restricted permissions compared to the default ``avnadmin`` user. You can add service users for all the Aiven services, with the exception of Aiven for Apache Flink® and Aiven for Grafana®.
 
+.. warning::
+   By default, the maximum amount of users allowed on a service is 50. If a service user is created that exceeds the max users, the following error will be produced.
+   
+   .. code-block:: python 
+      
+      "Service user 'xxx' cannot be created, maximum number of users for this plan exceeded"
```
staceysalamon-aiven, line 11:
```suggestion
```

staceysalamon-aiven: Just removing the error code. As it is shown in console, we don't need to document it here.

docs/platform/howto/create_new_service_user.rst
```diff
@@ -3,6 +3,15 @@
...
+
+   If you would like to increase the maximum amount of users on a service, please reach out to the Aiven support team to request an increase.
```
staceysalamon-aiven, line 13:
```suggestion
   If you would like to increase the maximum number of users allowed for a service, :doc:`create a support ticket </docs/platform/howto/project-support-center>` to request an increase.
```

staceysalamon-aiven (changes requested):
Just a few minor edits and also added a link to the article explaining how to contact support.

staceysalamon-aiven:
@josh-martin-aiven Thank you for adding this! I've made just a few minor suggestions in the review. Let me know if you have any questions. :)

docs/platform/howto/create_new_service_user.rst
josh-martin-aiven: That's great! I was actually trying to figure out how to put a link in there and couldn't figure it out, so it's super helpful for the next time I add to the docs. :)

docs/platform/howto/create_new_service_user.rst
josh-martin-aiven: Works for me. :)

docs/platform/howto/create_new_service_user.rst
josh-martin-aiven: For this and the code block, I would like to keep these if possible. The main reason for this is that customers will file support cases with us saying they got the error. If we include an example and description of the error here, customers would be able to find it through google, and would be better informed on what they need to ask for in the case. Let me know your thoughts on that.

docs/platform/howto/create_new_service_user.rst
josh-martin-aiven: For this one, my main goal was to provide and example error and explanation for customers getting this error, the hope being that they would be better informed on what to ask for when they open a case with us. For instance, as opposed to getting cases like "I got this error, what do I do do" we would get cases like  "Request to increase max users for service". Let me know your thoughts on that.

staceysalamon-aiven: Thanks for explaining. I see where you are coming from with this. In general, we try to avoid documenting things in the UI like this just because it can also easily go out of date with changes. Maybe we could just reword to say:
 "If you try to create too many service users, you will receive an error message stating, "maximum number of users for this plan exceeded".

This way we could also remove the code block and keep that formatting for code, but the key words will still be available for a search. Let me know if that works!

staceysalamon-aiven (commented):
Thank you for getting to this so quickly and for explaining the reasoning for the language used here. I just added a suggestion in the comments on how we could keep the search terms, but remove the code formatting. 

Happy to the final review/approval when this is ready. Thanks again!

staceysalamon-aiven (approved):
Thanks for your work on this, Josh! Looks good.

PR #1748, opened 2023-02-03
Fix link how to enable schema registry authorization on karapace

tvainika:

The context here is to enable schema registry authorization for older services, not enabling karapace schema registry itself.


harshini-rangaswamy (approved):

harshini-rangaswamy:
LGTM

Issue #1731, opened 2023-01-27
The examples in the Flink table from OpenSearch need to follow the new semantic


ftisiot:
# What's wrong?

(be specific, screenshots or a copy/paste of the affected part of the page is useful)
We switched from toggles to the full Flink SQL semantic for the table definition, the OpenSearch page needs to be updated

https://docs.aiven.io/docs/products/flink/howto/connect-opensearch

# URL of affected page, and any other information

PR #1751, opened 2023-02-03
updated example table for OS

harshini-rangaswamy:

Update the example in the Define an Apache Flink® table to OpenSearch® section. 

Changed to sink table instead of source in the procedure above. 

Fixes: #1731


ftisiot (approved):
LGTM

PR #1752, opened 2023-02-03
Update ogimage

angelinekwan:
The current preview image that are visible when share a page is not of the best visual and has the old Aiven developer name. Update to new visual.

ogp_image needs to be absolute path as [documented here](https://github.com/wpilibsuite/sphinxext-opengraph#simple-config). Currently it has a relative path that makes the image not crawl able - see [generator preview](https://socialsharepreview.com/?url=https://docs.aiven.io).


staceysalamon-aiven (approved):
Looks much better, thanks!

angelinekwan:
> Looks much better, thanks!

Agree it looks much polished :) https://socialsharepreview.com/?url=https://docs.aiven.io
og

PR #1764, opened 2023-02-08
platform: add the `do-syd1` region to the supported cloud regions list

wojcik-dorota:
https://aiven.atlassian.net/browse/DOC-172


wojcik-dorota:
No need for numbers for any DO zones

PR #1769, opened 2023-02-08
initial commit for custom AES serde example

paulgrainger-aiven:
Adding a how-to article on a way to encrypt data client-side before it enters the Kafka service. 

Some customers have additional compliance needs in which data has to be encrypted prior to entering the Aiven clusters, this is one way of achieving that.


jhepp00 (approved):

jhepp00 (approved):
Fixed Vale issues

PR #1772, opened 2023-02-10
clickhouse: add limitation for data load from S3

wojcik-dorota:
https://aiven.atlassian.net/browse/DOC-184

### DIFF
--- a/docs/products/clickhouse/reference/limitations.rst
+++ b/docs/products/clickhouse/reference/limitations.rst
@@ -52,6 +52,9 @@ From the information about restrictions on using Aiven for ClickHouse, you can e
    * - Creating a database using SQL
      - You cannot create a database directly using SQL, for example, if you'd like to add a non-default database.
      - Use the Aiven's public API.
+   * - Loading data from S3 (coming soon!)
+     - You cannot load you data from S3 into Aiven for ClickHouse directly using SQL.
+     - Either use vector or a local ClickHouse server and run the SELECT/INSERT query using both s3() and remote() functions to pull from S3 and push to Aiven.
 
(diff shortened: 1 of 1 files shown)


docs/products/clickhouse/reference/limitations.rst
```diff
@@ -52,6 +52,12 @@ From the information about restrictions on using Aiven for ClickHouse, you can e
    * - Creating a database using SQL
      - You cannot create a database directly using SQL, for example, if you'd like to add a non-default database.
      - Use the Aiven's public API.
+   * - Loading data from S3 (coming soon)
+     - You cannot load your data from S3 into Aiven for ClickHouse directly using SQL.
+     - There are two ways to work around this issue:
+
+       * Use `Vector <https://vector.dev/>`_.
+       * Use a local ClickHouse server and run the SELECT/INSERT query using both ``s3()`` and ``remote()`` functions to pull from S3 and push to Aiven.
```
kmichel-aiven, line 60:
```suggestion
     - There are many ways to work around this limitation, for instance:

       * Use `Vector <https://vector.dev/>`_.
       * Use a local ClickHouse server and run the SELECT/INSERT query using both ``s3()`` and ``remote()`` functions to pull from S3 and push to Aiven.
```

I'm not firmly attached to it, but we're just suggesting two options, there are many many more acceptable options.

Listing these two is enough to get the general idea: "you need something in the middle to pull/push",
and suggests both a read-made tool and a more technical solution. So I don't think we need to list many more, but we probably shouldn't make it look there are only two options.

kmichel-aiven (approved):

PR #1774, opened 2023-02-13
postgresql: how to migrate pg database using the console

wojcik-dorota:
https://aiven.atlassian.net/browse/DOC-175
This is the guide on how to use the console migration tool to migrate PG databases.

Based on https://www.figma.com/file/CLVBPTf0A9DPeXrigqAThO/Database-migration?node-id=369%3A62896&t=nqpc1mFG7YBVJtUa-0


divdby0 (approved):

docs/products/postgresql/howto/migrate-db-to-aiven-via-console.rst
```diff
@@ -0,0 +1,278 @@
...
+    For the other migration method (using ``aiven-db-migrate``), see :doc:`Migrate to Aiven for PostgreSQL® with aiven-db-migrate </docs/products/postgresql/howto/migrate-aiven-db-migrate>`.
+
+About migrating via console
+---------------------------
+
+* With the console migration tool, you can migrate existing on-premise or cloud-hosted PostgreSQL databases to clusters in your Aiven account.
+
+  .. note:: 
+    
+     The console migration tool doesn't support migrating managed database clusters on Aiven to other managed database clusters on Aiven.
```
alexole, line 17: I need to check this still, it should be possible actually with `aiven_extras` installed.

docs/products/postgresql/howto/migrate-db-to-aiven-via-console.rst
```diff
@@ -0,0 +1,278 @@
...
+
+* Console migration tool uses `logical replication <https://www.postgresql.org/docs/current/logical-replication.html>`_ by default.
+
+* With the console migration tool, you can migrate PostgreSQL® databases either using the continuous migration method, which is the one recommended and detailed in this guide, or using the dump method (``pg_dump``).
+
+  * Continuous migration keeps the source database operational during the migration but requires superuser permissions.
```
alexole, line 23: We use the same `aiven-db-migrate` under the hood for migration from the console, so the requirements should be the same, which means super user is needed or `aiven_extras` installed with probably replication permissions, I will check it and comment later.

docs/products/postgresql/howto/migrate-db-to-aiven-via-console.rst
alexole: I've checked between the Aiven services and it worked, using avnadmin user and after installing `aiven_extras` on the source.

docs/products/postgresql/howto/migrate-db-to-aiven-via-console.rst
alexole: It does not require super user, just checked with avnadmin and installed aiven_extras, replication did work.

docs/products/postgresql/howto/migrate-db-to-aiven-via-console.rst
wojcik-dorota: @alexole , thank you - removing the note in that case :-)

docs/products/postgresql/howto/migrate-db-to-aiven-via-console.rst
wojcik-dorota: @alexole , thank you - in that case I'm adding the use of installed aiven_extras as an alternative to superuser perms.

docs/products/postgresql/howto/migrate-db-to-aiven-via-console.rst
```diff
@@ -0,0 +1,308 @@
...
+    .. code-block:: bash
+
+       \du
+
+    .. code-block:: bash
+       :caption: Expected output
+
+       Role name |                      Attributes                            |                 Member of
+       ----------+------------------------------------------------------------+-----------------------------------------
+       _dodb     | Superuser, Replication                                     | {}
```
packi, line 45: We should probably use a different `Role name` here, not sure DO appreciate us using their platform as a source of migration, suggesting users should migrate to aiven.

docs/products/postgresql/howto/migrate-db-to-aiven-via-console.rst
```diff
@@ -0,0 +1,308 @@
...
+
+  .. code-block:: bash
+     :caption: Expected output if enabled
+
+     wal_level
+     -----------
+     logical
+     (1 row)
+
+  If the command prompt returns something different, enable the logical replication in your database by setting ``wal_level`` to ``logical``:
```
packi, line 149: I think we require _at least_ `logical` (I remember reading that somewhere in the code) so it may be worth trying that out or double checking that with the documentation.

docs/products/postgresql/howto/migrate-db-to-aiven-via-console.rst
```diff
@@ -0,0 +1,308 @@
...
+      :width: 700px
+      :alt: Connection details
+
+2. Select the **SSL encryption (recommended)** checkbox.
+
+3. Select **Run checks** to have the connection validated.
+
+.. topic:: Unable to use logical replication?
+
+   If your connection check returns the **Unable to use logical replication** warning, either resolve the issues or give up using the logical replication and opt for the dump method by selecting **Start the migration using a one-time snapshot (dump method)** > **Run check** > **Start migration**.
```
packi, line 230: wording: _give up_ seems a bit strong, is _revert to the dump method_ better?

packi (commented):
Couple of comment but looks good. Caveat: as with MySQL I haven't tried it out.

I remember an issue with the migration still being active and if it wasn't removed and the service was cloned, the clone started the migration again by truncating the data. Not sure if that is still the case, but something to check and document to really make sure the migration is not active anymore.

divdby0 (approved):

divdby0:
All good from my side. I am still not sure about PG 10 version. I couldn't finish testing it myself. We might consider changing it to PG 11.

docs/products/postgresql/howto/migrate-db-to-aiven-via-console.rst
Matthieu68857: There's nothing more after `logical` [https://www.postgresql.org/docs/current/runtime-config-wal.html](https://www.postgresql.org/docs/current/runtime-config-wal.html)
It's good for me!

wojcik-dorota:
> 

@packi , thank you! I asked around about this scenario and people are either not familiar with it or consider it as definitely an edge case. Therefore, it's been decided to leave it our from the docs. Hope that makes sense to you?

wojcik-dorota:
> All good from my side. I am still not sure about PG 10 version. I couldn't finish testing it myself. We might consider changing it to PG 11.

@divdby0 , upon checking, PG 10 should be fine. The guide does not mention any specific PG version as a prerequisite. Do you think we need to add it? IIRC we didn't add MySQL version as a prerequisite to the other migration guide at least.

docs/products/postgresql/howto/migrate-db-to-aiven-via-console.rst
```diff
@@ -0,0 +1,308 @@
...
+
+* With the console migration tool, you can migrate PostgreSQL® databases either using the continuous migration method, which is the one recommended and detailed in this guide, or using the dump method (``pg_dump``).
+
+  * Continuous migration keeps the source database operational during the migration. It requires either superuser permissions or the ``aiven_extras`` extension.
+
+    * Superuser permissions can be verified using the guidelines in :ref:`Check out how to verify that you have superuser permissions <check-superuser>`. If you do not have superuser permissions, you can request them from your system administrator.
+    * ``aiven_extras`` extension (`GitHub repository <https://github.com/aiven/aiven-extras>`_) can be installed on your source database using the guidelines in :doc:`Install or update an extension </docs/products/postgresql/howto/manage-extensions>`.
+
+  * Dump is a point-in-time snapshot and requires the source database to be offline or in the maintenance mode during the migration process since any data it gets after initiating the dump are not migrated to the target database.
```
alexole, line 28: `to be offline or in the maintenance mode` sound a bit misleading, actually we need it to be online for the migration, I understand what it means, but not sure will be clear for everybody. I would just say that dump method is a snapshot and the data written during the dump won't be copied to the target. So when this method is used client should not be writing into the source DB in order to get consistent data on the target or something like that.

docs/products/postgresql/howto/migrate-db-to-aiven-via-console.rst
```diff
@@ -0,0 +1,308 @@
...
+
+       Role name |                      Attributes                            |                 Member of
+       ----------+------------------------------------------------------------+-----------------------------------------
+       _source_pg_db     | Superuser, Replication                                     | {}
+       example   | Create role, Create DB, Replication, Bypass RLS            | {pg_read_all_stats,pg_stat_scan_tables,pg_signal_backend}
+       postgres  | Superuser, Create role, Create DB, Replication, Bypass RLS | {}
+
+    Identify your role name in the ``Role name`` column and check if it has the ``Superuser`` attribute assigned in the ``Attributes`` column.
+
+.. topic:: No superuser permissions? Migrate using the dump method.
```
alexole, line 51: We mentioned it's possible to use replication without superuser, but with `aiven_extras` extension installed on the source databases, so we should probable describe how. And only if that is also not possible for any reason - fallback to dump.

divdby0 (approved):

PR #1787, opened 2023-02-20
Orgs diagram 

staceysalamon-aiven:
Added a diagram for the organizational hierarchy with units and projects for clarity.


harshini-rangaswamy (approved):
LGTM

PR #1789, opened 2023-02-24
fixed list of flink CLI commands

ftisiot:

Fixed in the service page the list of CLI flink commands


harshini-rangaswamy (approved):
LGTM

Issue #1276, opened 2022-08-22
SEO: Have more descriptive text  in linktexts


laysauchoa:
Checking [devportal](https://developer.aiven.io/) website with [Lighthouse tool](https://chrome.google.com/webstore/detail/lighthouse/blipmdconlkpinefehnmjammfjpmpbjk?hl=en), we have performed average (83/100) on SEO. To be considered good, we need to be above 90 points. 

These checks ensure that devportal page is following basic search engine optimization advice.

Problem found:
It found 10 links which the link text has not descriptive text. They are all located in the landing page https://developer.aiven.io/

For example:
[…products/postgresql](https://developer.aiven.io/docs/products/postgresql)
(developer.aiven.io)
	Read more

As explained, descriptive link text helps search engines understand your content. You can read more: https://web.dev/link-text/?utm_source=lighthouse&utm_medium=devtools

Proposed solution:
The solution for this is to include some more specific descriptions in the link text. I am adding some proposals here which would make it more specific, but feel free to add your thoughts on how would you solve that.

1) Use service name with logo on the button instead of Read more
<img src="https://user-images.githubusercontent.com/36624597/185863019-ab3dda73-8a73-450e-b344-5a840628ea23.png" width="600" height="350">

2) Use service name without logo on the button instead of Read more. Have logo on the beginning of the box. 
<img src="https://user-images.githubusercontent.com/36624597/185863025-c5d81073-5c75-4d9b-93dc-4c4784f8fb28.png" width="600" height="350">

lornajane:
I would prefer that we do not make this change. If the buttons are confusing our users, let's get some advice from our UX colleagues on what would make this landing page more understandable and inclusive for all.

laysauchoa:
@lornajane would be possible to tag the colleagues that can help us with that? thanks

staceysalamon-aiven:
@lornajane @laysauchoa I'm happy to work on this from a UX copy perspective and discuss with the design team as needed. :)

laysauchoa:
This would be amazing @staceysalamon-aiven thank you so much! I am available for any questions or help from our side.

staceysalamon-aiven:
@laysauchoa @lornajane I think the best option for the product cards might be to change the button text to: "Learn about [PRODUCT]". It's more accessible and better for SEO, but I think also short enough to look okay and be easily scannable.

[...]

PR #1794, opened 2023-02-27
Make links descriptive

staceysalamon-aiven:
Made links on the docs homepage descriptive for better accessibility and SEO. Some other minor changes to tool descriptions and other microcopy. 

Resolves: #1276

### DIFF
--- a/index.rst
+++ b/index.rst
@@ -222,7 +222,7 @@ Interfaces
             :color: primary
             :outline:
 
-            Aiven Console Overview
+            Aiven Console overview
     
     .. grid-item-card::
         :shadow: md
@@ -237,7 +237,7 @@ Interfaces
             :color: primary
             :outline:
 
-            Aiven CLI Quickstart
+            Aiven CLI quickstart
 
 
 Automation
@@ -257,7 +257,7 @@ Automation
             :color: primary
             :outline:
 
-            API Quickstart
+            API quickstart
 
     .. grid-item-card::
         :shadow: md
--- a/index.rst
+++ b/index.rst
@@ -1,16 +1,8 @@
 Aiven documentation
 ===================
 
-Aiven is a database-as-a-service for open source data solutions including
-PostgreSQL®,
-M3,
-Apache Kafka®,
-Apache Flink®,
-OpenSearch®,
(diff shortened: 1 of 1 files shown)


harshini-rangaswamy (changes requested):
A few minor things in this PR, 

- could you right-align the links to "Aiven Console Overview and Aiven CLI Quickstart (should Overview and Quickstart be lowercase? )

- Automation section - make the boxes similar in size to the above section for uniformity. 

image

I- Not sure about how feasible this is, but it would be nice to align the links, for example, "Learn about M3" and "Learn about MySQL" 

image

staceysalamon-aiven:
> A few minor things in this PR,
> 
> * could you right-align the links to "Aiven Console Overview and Aiven CLI Quickstart (should Overview and Quickstart be lowercase? )
> * Automation section - make the boxes similar in size to the above section for uniformity.
> 
> image
> 
> * Not sure about how feasible this is, but it would be nice to align the links, for example, "Learn about M3" and "Learn about MySQL"
> 
> image

I changed the alignment to keep everything from looking the same; same with the automation tools. I wanted to have a little different look/weight because otherwise I think everything on the page just looks the same.

I'm not aware of a way to manually align the links, but I do think the copy for services should be updated, which could help with these being a little more aligned (though, that will depend on the device).

harshini-rangaswamy:
> I changed the alignment to keep everything from looking the same; same with the automation tools. I wanted to have a little different look/weight because otherwise I think everything on the page just looks the same.
> 
> I'm not aware of a way to manually align the links, but I do think the copy for services should be updated, which could help with these being a little more aligned (though, that will depend on the device).

I think having some degree of symmetry in the design is visually appealing. While scrolling through the page, there is a sense of disconnection in its overall look and feel. For instance, varying alignment of links, such as the center alignment of "Aiven Console overview" and "Aiven CLI quickstart", compared to the rest that is right-aligned. (this may be just my perception, and others may view it differently). 
The points may seem trivial since it's uncertain whether users browse the page or use the search function to find information.
Maybe, we could push the changes and see if we get any feedback?

harshini-rangaswamy (approved):

PR #1798, opened 2023-03-06
platform: backup to another region (BTAR) docs

wojcik-dorota:
PR transferred to https://github.com/aiven/aiven-docs/pull/6 due to docs migration to a new platform.

### DIFF
--- a/docs/platform/concepts/backup-to-another-region.rst
+++ b/docs/platform/concepts/backup-to-another-region.rst
@@ -1,0 +1,79 @@
+About backup to another region for Aiven services
+=================================================
+
+.. important::
+
+    Backup to another region (BTAR) is available on ``:doc:Pro Platform </docs/platform/concepts/pro-platform>``.
+
+Discover the backup to another region (BTAR) feature, check why use it, and learn how it works.
+
+About BTAR
+----------
+
+The backup to another region (BTAR) feature allows existing backup files to be copied from the service's primary backup region to an additional (secondary) region. As a disaster recovery feature, BTAR is particularly useful when the service-hosting region is down since it allows forking the service from an additional copy of the backup residing outside the service-hosting region.
+
+BTAR is currently supported for the following services:
+
+* Aiven for PostgreSQL®
+* Aiven for MySQL®
+
+Why use BTAR
+------------
+
+BTAR provides an additional level of data resilience, helping you make your data safe and protected. BTAR is a cost-efficient low-complexity disaster recovery feature that allows you to replicate your data in a secondary backup region and recover it there in case of outage, downtime, data loss, or data corruption in the primary backup region.
+
+How BTAR works
+--------------
+
+By enabling the backup to another region (BTAR) feature, you create a service backup that is additional to the default backup. This secondary backup is always located in a region of your choice different from the primary backup region.
+
+Secondary backups are generated from primary backups, not from the service itself. For this reason, your secondary backup becomes available only after the primary backup is taken. There might be a data replication lag between the primary region and the secondary region.
+
+Restoring from a secondary backup, for example in case of an outage of the primary region, can be done by creating a fork of the service in the region where the secondary backup is located.
+
+.. mermaid::
+
+    flowchart LR
+        subgraph Primary_region
+            direction TB
+            subgraph Service_X
+                end
+            subgraph Primary_backups
+                direction LR
+                PB1
+                PB2
+                PB3
+                PBn
+                end
+            end
+        subgraph Secondary_region
+            direction TB
+            subgraph Forked_service_X
+                end
+            subgraph Secondary_backups
+                direction LR
+                SB1
+                SB2
+                SB3
+                SBn
+                end
+            end
+        Service_X -- Default \n backups --> Primary_backups
+        Primary_backups -- Cross-region backups \n if BTAR  enabled --> Secondary_backups
+        Secondary_backups -- Secondary backups \n to restore service X --> Forked_service_X
+        Service_X -- Forking service X \n if primary region down --> Forked_service_X
+
+Limitations
+-----------
+
+* When selecting a cloud region for your additional backup, you need to use the same cloud provider that your service uses.
+* When you want to :ref:`restore your service from an additional backup <fork-and-restore>` and you use a point in time to specify the scope of data to be restored, you need to set up the time to no later than the time of the latest backup.
+* Secondary backup can only be restored in the region where it was stored.
+* Secondary backup is generated only after a primary backup is complete, and there might be a data replication lag between the primary region and the secondary region.
+
+Related pages
+-------------
+
+* :doc:`Enable BTAR </docs/platform/howto/enable-backup-to-another-region>`
+* :doc:`Manage BTAR </docs/platform/howto/manage-backup-to-another-region>`
+* :doc:`Disable BTAR </docs/platform/howto/disable-backup-to-another-region>`
--- a/docs/platform/howto/enable-backup-to-another-region.rst
+++ b/docs/platform/howto/enable-backup-to-another-region.rst
@@ -1,0 +1,162 @@
+Add a backup to another region
+==============================
+
+.. important::
+
+    Backup to another region (BTAR) is available on ``:doc:Pro Platform </docs/platform/concepts/pro-platform>``.
+
+Enable :doc:`the backup to another region (BTAR) feature </docs/platform/concepts/backup-to-another-region>` and create an additional cross-region service backup on top of a regular backup stored in the region where your service is hosted.
+
+About enabling BTAR
+-------------------
+
+Tools
+'''''
+
+To add an additional service backup for your service, you can use the following tools:
+
+* `Aiven console <https://console.aiven.io/>`_
+* CLI
+* API
+
+Limitations
+'''''''''''
+
+* Currently, BTAR is supported for Aiven for MySQL® and Aiven for PostgreSQL® only.
+
+Prerequisites
+-------------
+
+* Aiven organization, project, and service
+* ``:doc:Pro Platform </docs/platform/concepts/pro-platform>`` enabled
+* Depending on the method you choose to use for enabling CCR
+
+  * Access to the `Aiven Console <https://console.aiven.io/>`_
+  * `cURL` CLI tool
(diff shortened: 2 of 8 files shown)


docs/platform/concepts/backup-to-another-region.rst
```diff
@@ -0,0 +1,70 @@
...
+            end
+        Service_X -- Default \n backups --> Primary_backups
+        Primary_backups -- Cross-region backups \n if BTAR  enabled --> Secondary_backups
+        Secondary_backups -- Secondary backups \n to restore service X --> Forked_service_X
+        Service_X -- Forking service X \n if primary region down --> Forked_service_X
+
+Limitations
+-----------
+
+* When selecting a cloud region for your additional backup, you need to use the same cloud provider that your service uses.
```
sjamgade, line 60: a few words about:  "this is a temporary limitation, about to be lifted away soon" would be good

docs/platform/concepts/backup-to-another-region.rst
```diff
@@ -0,0 +1,70 @@
...
+* When you want to :ref:`restore your service from an additional backup <fork-and-restore>` and you use a point in time to specify the scope of data to be backed up, you need to set up the time to no later than the time of taking the latest backup.
```
sjamgade, line 61:
```suggestion
* When you want to :ref:`restore your service from an additional backup <fork-and-restore>` and you use a point in time to specify the scope of data to be restored, you need to set up the time to no later than the time of latest backup.
```

this is actually a limitation, that only upto the latest backup can be restored (from secondary location) and nothing after that. This limitation will also be lifted in future

docs/platform/howto/enable-backup-to-another-region.rst
```diff
@@ -0,0 +1,167 @@
...
+  * `Aiven CLI tool <https://github.com/aiven/aiven-client>`_
+
+Back up to another region via console
+-------------------------------------
+
+1. Log in to the `Aiven console <https://console.aiven.io/>`_.
+2. From the **Current services** view, select an Aiven service on which you'd like to enable BTAR.
+3. In the **Overview** tab of your service's page, navigate to **Advanced configuration** and select **Change**.
+4. In the **Edit advanced configuration** window, select **Add configuration option** > **additional backup regions**.
+5. Enter a region name into the **additional backup regions** field and select **Save advanced configuration**.
```
sjamgade, line 40: The `region name` here would not be obvious. 
the user has to actually copy this from some list of regions, (not sure which could be the correct source, but found this https://docs.aiven.io/docs/platform/reference/list_of_clouds)

sjamgade: may be a comment indicating where to get that list would be good.

UX-Irene:
Hi @caydinavn, I have a question about forking to the secondary location. If we have a disaster scenario and the user forks the service to the secondary location, which is still up and running. Is it true that the new forked service will have the same backup locations as the source. So the primary location will be the one that currently is down?

docs/platform/concepts/backup-to-another-region.rst
```diff
@@ -0,0 +1,79 @@
...
+.. important::
+
+    Backup to another region (BTAR) is available on ``:doc:Pro Platform </docs/platform/concepts/pro-platform>``.
+
+Discover the backup to another region (BTAR) feature, check why use it, and learn how it works.
+
+About BTAR
+----------
+
+The backup to another region (BTAR) feature allows existing backup files to be copied from the service's primary backup region to one or more regions different from the primary one. As a disaster recovery feature, BTAR is particularly useful when the service-hosting region is down since it allows forking the service from an additional copy of the backup residing outside the service-hosting region.
```
caydinavn, line 13:
```suggestion
The backup to another region (BTAR) feature allows existing backup files to be copied from the service's primary backup region to an additional region. As a disaster recovery feature, BTAR is particularly useful when the service-hosting region is down since it allows forking the service from an additional copy of the backup residing outside the service-hosting region.
```

We may or may not allow multiple regions in the future but we don't at the moment.

docs/platform/concepts/backup-to-another-region.rst
```diff
@@ -0,0 +1,79 @@
...
+
+BTAR is currently supported for the following services:
+
+* Aiven for PostgreSQL®
+* Aiven for MySQL®
+
+Why use BTAR
+------------
+
+BTAR provides an additional level of security for your data. It is a cost-efficient low-complexity disaster recovery feature that allows you to spread your data across different regions and recovery it in case of data loss or corruption in the primary region. During an outage or a downtime on the primary region, your data is protected and safe in a secondary backup location.
```
caydinavn, line 23:
```suggestion
BTAR provides an additional level of resilience for your data. It is a cost-efficient, low-complexity disaster recovery feature that allows you to replicate your data in a different region and recover there in case of outage, downtime, data loss or corruption in the primary region.
```

docs/platform/concepts/backup-to-another-region.rst
```diff
@@ -0,0 +1,79 @@
...
+
+How BTAR works
+--------------
+
+By enabling the backup to another region (BTAR) feature, you create an additional backup on top of the default one located in a primary region where your service is hosted. Contrary to the primary backup, the secondary backup is located outside the primary region.
```
caydinavn, line 28:
```suggestion
By enabling the backup to another region (BTAR) feature, you create an additional backup in addition to the default one. This secondary backup is always located in a secondary region of your choice.
```

docs/platform/concepts/backup-to-another-region.rst
```diff
@@ -0,0 +1,79 @@
...
+
+Secondary backups are taken from primary backups, not from the service itself. For this reason, your secondary backup is available only after a primary backup is there. There might be a data replication lag between the primary region and the secondary region.
```
caydinavn, line 30:
```suggestion
Secondary backups are generated from primary backups, not from the service itself. For this reason, your secondary backup becomes available only after the primary backup is taken. There might be a data replication lag between the primary region and the secondary region.
```

docs/platform/concepts/backup-to-another-region.rst
```diff
@@ -0,0 +1,79 @@
...
+
+If you want to have your service restored from a secondary backup, for example, in case of an outage of the primary region, you create a fork of the service using a region where your secondary backup is located.
```
caydinavn, line 32:
```suggestion
Restoring from a secondary backup, for example in case of an outage of the primary region, can be done by creating a fork of the service in the region where the secondary backup is located.
```

docs/platform/concepts/backup-to-another-region.rst
```diff
@@ -0,0 +1,79 @@
...
+            end
+        Service_X -- Default \n backups --> Primary_backups
+        Primary_backups -- Cross-region backups \n if BTAR  enabled --> Secondary_backups
+        Secondary_backups -- Secondary backups \n to restore service X --> Forked_service_X
+        Service_X -- Forking service X \n if primary region down --> Forked_service_X
+
+Limitations
+-----------
+
+* When selecting a cloud region for your additional backup, you need to use the same cloud provider that your service uses (support for creating cross-cloud-provider backups coming soon!).
```
caydinavn, line 69:
```suggestion
* When selecting a cloud region for your additional backup, you need to use the same cloud provider that your service uses.
```

docs/platform/concepts/backup-to-another-region.rst
```diff
@@ -0,0 +1,79 @@
...
+* When you want to :ref:`restore your service from an additional backup <fork-and-restore>` and you use a point in time to specify the scope of data to be restored, you need to set up the time to no later than the time of the latest backup.
+* To restore a service from an additional (not default) backup, you need to create a fork of the original service in the region where the additional backup resides.
```
caydinavn, line 71:
```suggestion
* Secondary backups can only be restored in the region where they were stored.
```

docs/platform/concepts/backup-to-another-region.rst
```diff
@@ -0,0 +1,79 @@
...
+* Secondary backup is available only after a primary backup is there and there might be a data replication lag between the primary region and the secondary region.
```
caydinavn, line 72:
```suggestion
* Secondary backup is generated only after a primary backup is complete and there might be a data replication lag between the primary region and the secondary region.
```

images/platform/howto/add-advanced-parameter.png
caydinavn, line 1: This looks out of date, will need new screenshots from UX (and probably need to review action sequences accordingly)

caydinavn (dismissed):
Made some edit suggestions, also I think we'll need to review final UX and update the action sequences accordingly.

caydinavn:
> Hi @caydinavn, I have a question about forking to the secondary location. If we have a disaster scenario and the user forks the service to the secondary location, which is still up and running. Is it true that the new forked service will have the same backup locations as the source. So the primary location will be the one that currently is down?

The primary backup location of the new service should also be the new region, @ali-gholami-aiven could you double check this?

images/platform/howto/add-advanced-parameter.png
wojcik-dorota: Removed it. It shouldn't have been there in the first place - we refrain from using images as much as possible to restrict potential maintenance effort.

wojcik-dorota:
> Made some edit suggestions, also I think we'll need to review final UX and update the action sequences accordingly.

The steps have been updated according to the new UX and checked with @UX-Irene .

wojcik-dorota:
> > > Hi @caydinavn, I have a question about forking to the secondary location. If we have a disaster scenario and the user forks the service to the secondary location, which is still up and running. Is it true that the new forked service will have the same backup locations as the source. So the primary location will be the one that currently is down?
> > 
> > 
> > The primary backup location of the new service should also be the new region, @ali-gholami-aiven could you double check this?
> 
> That's correct. I double checked it. primary backup location of the new (forked) service will be the new region (without additional backup region configured)

Thanks @ali-gholami-aiven ! Does it only refer to forking or also to migrating? I mean, Is that still correct to say:
```
When you migrate your service, locations of service backups, both primary and secondary ones, do not change.
```

wojcik-dorota:
PR transferred to https://github.com/aiven/aiven-docs/pull/6 due to docs migration to a new platform.

PR #1810, opened 2023-03-14
Fix formatting of numbering in Migrate public service to VPC article

staceysalamon-aiven:
This PR just fixes the RST formatting to show the step numbers in the heading instead of #.


staceysalamon-aiven:
The heading steps have now been rewritten to fit with the sentence case enforced by Vale because it wouldn't allow the format "Step X. Sentence case heading" for some reason only in the first three instances (despite working in every other step and in other articles). For now, I'd like to get at least this merged so that the article no longer has "#" in the headings.

wojcik-dorota:
@staceysalamon-aiven , thanks for handling that! How about we get rid of the numbers totally and just provide section names (using the sentence capitalization)? The order of the steps is clearly stated in the Overview section so no fear that the user starts taking those steps randomly. The current change - starting a section name with a lower case after a number - doesn't seem right.

PR #1821, opened 2023-03-17
Fix instructions for moving projects

staceysalamon-aiven:
Instructions currently state it's possible to move projects between organizations, but moving between units within an org is the only action that is supported.

### DIFF
--- a/docs/platform/howto/manage-project.rst
+++ b/docs/platform/howto/manage-project.rst
@@ -39,9 +39,9 @@ Renaming a project is possible **only** when all the services in the project are
 Move a project
 ---------------
 
-To move a project from one organization or organizational unit to another:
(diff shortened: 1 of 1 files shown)


docs/platform/howto/manage-project.rst
```diff
@@ -39,17 +39,20 @@ Renaming a project is possible **only** when all the services in the project are
 Move a project
 ---------------
 
-To move a project from one organization or organizational unit to another:
+To move a project from one  organizational unit to another:
```
m-novikov, line 42:
```suggestion
To move a project from one organizational unit to another:
```
Extra space

m-novikov (changes requested):
Small comment, otherwise LGTM

PR #1823, opened 2023-03-20
Stop preserving user in anonymous tracking and use the same state storage strategy

niklas-lumio-aiven:
- Previous configuration stored the user for anonymous users, which is not inline with anonymous session
- Every page view had different session id, so the session was not stored correctly
-> We needed to ensure the tracker and the anonymous tracking uses the same storage strategy
-> https://docs.snowplow.io/docs/collecting-data/collecting-from-own-applications/javascript-trackers/javascript-tracker/javascript-tracker-v3/tracker-setup/additional-options/#disable-anonymous-tracking


carabaestlein (approved):
LGTM!

PR #1843, opened 2023-03-30
Add end of availability dates for Kafka

tvainika:

Add Availability end on Aiven Platform dates of Kafka and reformat the table same as other services.


aiven-anton (approved):
Thanks for this fix! 👍

PR #1849, opened 2023-04-06
Slightly clearer wording on the marketplace migration process

kevinbowman-aiven:

Feedback from customers using this process has led to needing slightly clearer wording on which IDs they need to send us.

### DIFF
--- a/docs/platform/howto/move-to-gcp-marketplace-billing.rst
+++ b/docs/platform/howto/move-to-gcp-marketplace-billing.rst
@@ -17,9 +17,9 @@ Aiven will need some information from both your existing and new subscriptions i
 
 * The name of the project (or projects) that contain the services you wish to move. 
 
-**From your new GCP Marketplace subscription:**
+**From your new Aiven account with the GCP marketplace subscription:**
 
-* Your Account ID (in the format ``E-1234-1234-1234-1234``), as shown in the Aiven GCP console **Settings** screen.
+* Your Account ID (in the format ``E-1234-1234-1234-1234``), as shown at the top of the `Aiven GCP console <https://console.gcp.aiven.io/>`_.
 
 Send the request to Aiven
 -------------------------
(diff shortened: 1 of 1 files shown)


docs/platform/howto/move-to-gcp-marketplace-billing.rst
```diff
@@ -6,7 +6,7 @@ Aiven makes its services available through the Google Cloud Platform Marketplace
 Create a new account using a Google Cloud Platform Marketplace subscription
 ---------------------------------------------------------------------------
 
-Follow the steps to :doc:`set up Google Cloud Platform Marketplace for Aiven Services <billing-google-cloud-platform-marketplace-subscription>`.  This will create a new empty project, which is where your services will be moved to.
+Follow the steps to :doc:`set up Google Cloud Platform Marketplace for Aiven Services <billing-google-cloud-platform-marketplace-subscription>`.  This will create a new account, which is where your project will be moved to.
```
staceysalamon-aiven, line 9: Should this read, "This will create a new organization" since we use that term now instead of "accounts"?

docs/platform/howto/move-to-gcp-marketplace-billing.rst
```diff
@@ -17,9 +17,9 @@ Aiven will need some information from both your existing and new subscriptions i
 
 * The name of the project (or projects) that contain the services you wish to move. 
 
-**From your new GCP Marketplace subscription:**
+**From your new Aiven account with the GCP marketplace subscription:**
```
staceysalamon-aiven, line 20: Similar to the comment above, I'm not sure if this is the right term to use.

staceysalamon-aiven (changes requested):
Thanks for this! We have moved away from using the term "account" to avoid confusion with the recent changes from accounts to organizations/organizational units. I'm not certain how this looks in the GCP Console, so I have added a couple of questions in the comments. Thanks!

kevinbowman-aiven:
Thanks @staceysalamon-aiven , I've pushed a change to update to use "organization" terminology.

staceysalamon-aiven (approved):
Thanks! LGTM

PR #1850, opened 2023-04-06
Udpated content for OS optimal number of shards

harshini-rangaswamy:

I have updated the content on this topic based on information in Slack. 
Ref: https://aiven.atlassian.net/browse/DOC-280

### DIFF
--- a/docs/products/opensearch/concepts/shards-number.rst
+++ b/docs/products/opensearch/concepts/shards-number.rst
@@ -29,6 +29,8 @@ Consider the following recommendations for selecting the appropriate number of s
 
 These suggestions are only indicative, and optimal values depend on your usage patterns and anticipated data growth in OpenSearch®. Monitoring disk and CPU usage and upgrading when necessary to ensure optimal performance is essential.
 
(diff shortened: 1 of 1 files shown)


docs/products/opensearch/concepts/shards-number.rst
```diff
@@ -1,34 +1,39 @@
...
 
--  If you estimate you will have terabytes of data, increase the shard size compared to the previous example. 
 
-   For example, for 1TB index 50 shards could be a relevant suggestion.
+Adjusting shard count
+----------------------
 
-These suggestions are only indicative - optimal values depend heavily on
-your usage pattern and anticipated growth of data in OpenSearch.
+You can change the number of shards without losing your data, but this process requires a brief downtime while the index is rewritten.
```
fingon, line 36: As opensearch is frequently used for e.g. daily logs, another option is just adding more shards per index to new indexes so following days' indexes would have more shards.

harshini-rangaswamy: Updated content to include this information.

PR #1853, opened 2023-04-06
SUP-553: Updating datasource integration

bcrisologo:
# What's wrong?

There is no longer a **"datasource"** integration option available that the Datasource integration document is mentioning with the below currently available options:

image

Instead, what appears is being described is the **IP Allow-List** option

image

I believe this was changed recently so my plan is to update the content with the following:

> There are two types of datasource integrations you can use with Aiven for PostgreSQL®: Aiven for Grafana®, and another Aiven for PostgreSQL® service. If you are connecting two PostgreSQL® services together, perhaps to query across them, but still want to have a restricted IP allow-list, then you will need to use the ~Datasource~ **IP Allow-List** service integration.

> As the new IP address cannot be predicted, if you want to maintain a connection between two PostgreSQL services your choices are either to have a very broad IP allow-list (which might be acceptable in the private IP-range of a project VPC) or to use the ~Datasource~ **IP Allow-List** service integration to dynamically create an IP allow-list entry for the other PostgreSQL service.

>  1. On the service overview page for your PostgreSQL service, go to Manage Integrations and choose the ~datasource~ **IP Allow-List** option.

The above should be up to date with our current console options.

https://docs.aiven.io/docs/products/postgresql/howto/datasource-integration

https://docs.aiven.io/docs/products/postgresql/howto/datasource-integration#integrate-two-postgresql-services

### DIFF
--- a/docs/products/postgresql/howto/datasource-integration.rst
+++ b/docs/products/postgresql/howto/datasource-integration.rst
@@ -1,14 +1,14 @@
 Connect two PostgreSQL® services via datasource integration
 ===========================================================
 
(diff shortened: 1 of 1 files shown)


harshini-rangaswamy (approved):
LGTM!

PR #1854, opened 2023-04-07
postgresql: add information on delta base backups

wojcik-dorota:
https://aiven.atlassian.net/browse/DOC-159


docs/products/postgresql/concepts/pg-backups.rst
```diff
@@ -55,7 +62,13 @@ Differences between logical and full backups
...
 
-.. Tip::
-    The size of backups and the Aiven backup size shown on the Aiven web console differ, in some cases significantly. The backup sizes shown in the Aiven web console are for daily backups, before encryption and compression.
+Delta base backups
+------------------
+
+Aiven for PostgreSQL uses delta base backups, which allows to store data files that have been changed since the last backup and leave out the unchanged files. It's particularly beneficial for databases that include considerable portions of static data. Compared to regular base backups, delta base backups are more efficient, bringing improved performance and speeding up backup and restore operations.
+
+Since delta base backups don't take all the data files, they are faster and easier to perform on large databases with huge volumes of data. Because performing a delta base backup doesn't last long, the frequency of taking backups can be increased if needed. With the increased backup frequency, service restoration and node replacement are faster for highly updated services because fewer WAL files need to be restored since the last backup (WAL restoration in PostgreSQL is single-threaded and, therefore, slow).
```
alexole, line 70: `can be increased if needed`
I think we should clarify here `by OPS only`, it's not available as a user setting.

alexole: And I don't think even that should be done for every plan/customer, so I would just mention technical possibility or something. It sounds to me like customer can go and increase a backup frequency.
In reality the request should be analyzed first. We should check how static the data is in a specific case and then with some approval process do it.

alexole: In ideal world it should be done automatically, but it's not there yet :)

alexole (approved):
LGTM

PR #1865, opened 2023-04-14
Add instructions for upgrading TF to v4

staceysalamon-aiven:
For customers who have not upgraded from v2 to v3 some instructions are needed to move to version 4.

### DIFF
--- a/docs/tools/terraform/howto/upgrade-provider-v3-v4.rst
+++ b/docs/tools/terraform/howto/upgrade-provider-v3-v4.rst
@@ -68,7 +68,7 @@ In this example, the ``aiven_database`` field is updated to the service-specific
 
 2. View a list of all resources in the state file::
 
-    terraform state list | grep azure
+    terraform state list
 
 3. Remove the resource from the control of Terraform::
 
@@ -79,7 +79,7 @@ In this example, the ``aiven_database`` field is updated to the service-specific
 
 4. Add the resource back to Terraform by importing it as a new resource::
 
-    terraform import aiven_pg_database project_name/ ???
+    terraform import aiven_pg_database project_name/service_name/db_name
 
 5. Check that the import is going to run as you expect::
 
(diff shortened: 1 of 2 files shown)


Serpentiel (approved):
lgtm

staceysalamon-aiven:
@Serpentiel Thanks for reviewing! I just fixed up step 4 to include the same path as in the import resources instructions. Could you make sure that looks ok and then I'll merge! 🙏

docs/tools/terraform/howto/upgrade-provider-v3-v4.rst
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+
+    + resource "aiven_pg_database" "mydatabase" {
+        project       = aiven_project.myproject.project
+        service_name  = aiven_pg.mypg.service_name
+        database_name = "<DATABASE_NAME>"
+    }
+
+2. View a list of all resources in the state file::
```
Serpentiel, line 69: is it okay that we have double semicolon at the end here? also, if it is, please check step 1 as it doesn't have it that way

Serpentiel (changes requested):

docs/tools/terraform/howto/upgrade-provider-v3-v4.rst
staceysalamon-aiven: I'm pretty sure step 1 doesn't have it because it has the code block explicitly as `.. code::` (I just reused the content from the other article). I can change all of these to be consistent, though.

Serpentiel: ahh I see, well you're welcome if you want to, but it's good as it is already

staceysalamon-aiven: Ok, cool. Thank you!

Serpentiel (approved):
lgtm

PR #1882, opened 2023-05-04
Query statistics view and fetch instructions

cybermax-008:

Based on recent customer support tickets, where they asked to how to fetch query statistics from system.query_log table - it seems like we do not allow due to security reasons. But, we do have API to fetch this. (recently added to our API docs)

This KB contains instructions to fetch the query stats via console and API.


cybermax-008:
Will add more content and create again.

PR #1891, opened 2023-05-10
Removed preview tag for HTTP and Snowflake connectors

harshini-rangaswamy:

Removed the preview tag for HTTP and SnowFlake connectors, as these connectors are stable now,


staceysalamon-aiven (approved):
LGTM

PR #1893, opened 2023-05-10
Sup 590 kb update email alerts help article with new information found

cybermax-008:
Found some information on what happens if the technical emails are not setup. Seems like good info to add as some customers opened support ticket asking why admin members are spammed with email alerts.
Discussion : https://aiven-io.slack.com/archives/C01UBB57XFS/p1683664871912459


cybermax-008:
Made a mistake¡ will create a new one

PR #1895, opened 2023-05-10
updating tech email help article with additional info

cybermax-008:
Added some information on what if the tech email is not configured by the customer. From discussion: https://aiven-io.slack.com/archives/C01UBB57XFS/p1683700726178779?thread_ts=1683664871.912459&cid=C01UBB57XFS

### DIFF
--- a/docs/platform/howto/technical-emails.rst
+++ b/docs/platform/howto/technical-emails.rst
@@ -1,9 +1,9 @@
 Get technical notifications
 ============================
 
-To stay up to date with the latest news about the services in a project, you can set up notifications for technical contacts for the project. 
+To stay up to date with the latest news about the services in a project, you can set up notifications for technical contacts for the project. If you don't add a technical email for a project, email notifications will still be sent to users with admin and operator roles.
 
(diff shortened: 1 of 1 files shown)


docs/platform/howto/technical-emails.rst
```diff
@@ -14,6 +14,9 @@ Set up email notifications
 
 #. Click **Save changes**. 
 
+.. note::
```
staceysalamon-aiven, line 17: I think this would be better as part of the introduction (the first paragraph) and slightly reworded as: "If you don't add a technical email for a project, email notifications will still be sent to users with admin and operator roles."

staceysalamon-aiven (changes requested):
Thanks for adding this extra information! I've made a suggestion to move and reword this.

docs/platform/howto/technical-emails.rst
cybermax-008: Thanks for the review and suggestion. I have made the changes to the new commit.

staceysalamon-aiven (approved):
Looks good - thanks!

PR #1900, opened 2023-05-12
Added a note about forking OS service with SSO configured

harshini-rangaswamy:

Added a note that Single Sign-On (SSO) must be reconfigured on the forked OS services.


hoang-vo (approved):
Comments given during 1-1. Other than that, approved this PR

PR #1909, opened 2023-05-17
Added parameter option for Debezium CDC source

ftisiot:

The ``must be superuser to create FOR ALL TABLES publication`` has a solution which is more elegant than forcing people to create the publication beforehand. This PR adds it


harshini-rangaswamy (approved):
Thanks for adding this! Looks good.

PR #1911, opened 2023-05-19
clickhouse: update the instruction for integrating ClickHouse with PostgreSQL based on UI changes

wojcik-dorota:
https://aiven.atlassian.net/browse/DOC-342

### DIFF
--- a/docs/products/clickhouse/howto/data-service-integration.rst
+++ b/docs/products/clickhouse/howto/data-service-integration.rst
@@ -33,6 +33,8 @@ Prerequisites
 * Aiven account
 * Access to `Aiven web console <https://console.aiven.io/>`_
 
+.. _create-data-service-integration:
+
 Create data service integrations
 --------------------------------
 
--- a/docs/products/clickhouse/howto/integrate-kafka.rst
+++ b/docs/products/clickhouse/howto/integrate-kafka.rst
@@ -43,16 +43,7 @@ Variable                         Description
 Create an integration
 ----------------------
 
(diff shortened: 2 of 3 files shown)


kmichel-aiven (changes requested):
We do need to change the documentation (this one, but also the one for ClickHouse-Kafka), but the new version does not describe the right thing.

The "Store Metrics" integration is a way to send metrics about ClickHouse itself and store them in PostgreSQL, to monitor the ClickHouse service, but it's not the integration to connect ClickHouse and PostgreSQL together.

To connect ClickHouse and PostgreSQL with an Integration, it's not in the Integrations tab.

Instead, you need to click on "Get Started" in the Overview tab, and in the "Data Service Integrations" check the "Aiven for PostgreSQL" box at the top, and then select the service (or the endpoint).

wojcik-dorota:
> We do need to change the documentation (this one, but also the one for ClickHouse-Kafka), but the new version does not describe the right thing.
> 
> The "Store Metrics" integration is a way to send metrics about ClickHouse itself and store them in PostgreSQL, to monitor the ClickHouse service, but it's not the integration to connect ClickHouse and PostgreSQL together.
> 
> To connect ClickHouse and PostgreSQL with an Integration, it's not in the Integrations tab.
> 
> Instead, you need to click on "Get Started" in the Overview tab, and in the "Data Service Integrations" check the "Aiven for PostgreSQL" box at the top, and then select the service (or the endpoint).

@kmichel-aiven , thanks! I discovered we have this covered already so just added the redirection to the relevant section.

kmichel-aiven (approved):

PR #1916, opened 2023-05-23
Link to billing group article to change billing contact

staceysalamon-aiven:
Updated the instructions for changing the billing contact to reflect the fact that payment cards are added from the billing section by linking to that article. This article will be updated again later when the billing addresses function is released.


rsalevsky:
Billing Contacts is managed as part of the Billing Group. 
Screenshot from 2023-05-30 14-58-40

PR #1917, opened 2023-05-23
Key considerations and system adaptation 

harshini-rangaswamy:
Added content for key considerations and system adaptation for OpenSearch Security management.


hoang-vo (approved):
It looks good to me. However, once we get to test the official OpenSearch Terraform to see if it works with our OpenSearch, we can include that information here

PR #1920, opened 2023-05-26
Fixed Debezium publication name in Flink CDC docs

dbeech:

Debezium expects the use the name `dbz_publication` when creating the publication and it doesn't seem like there is any way to change that in the Flink connector configuration, so I'm updating the example code to use the right value.


juha-aiven (approved):
LGTM

PR #1923, opened 2023-05-26
Add time synchronization

celestehorgan:
Per Janne's comments in slack


jasamkos:
To add context: This is one of our commitments from ISO 27017 where we say that we provide information about time synchronization.

harshini-rangaswamy (approved):
LGTM!

PR #1925, opened 2023-05-26
Updated maintenance-window & service-power-cycle to include that

richard-joerger-aiven:
maintenance updates will be auto applied.

Added a sentence on the Service Power Cycle page to indicate that maintenance updates will automatically be applied when a service is powered on. 

Similarly, I added a sentence to the maintenance window page about the same piece of information as it pertains to upgrades. 

This matters as customers should have an idea of what occurs to their services during power up and upgrades. This may also help deflect cases from support.


docs/platform/concepts/maintenance-window.rst
```diff
@@ -22,4 +22,4 @@ In case of **Apache Kafka®** and **OpenSearch®** the service DNS address resol
...
\ No newline at end of file
+Security updates, platform updates that affect reliability or stability of the service nodes, and quarterly patch updates are always mandatory. Other updates are initially optional. Advance notice is given for all updates. After optional updates have been available for six months, they become mandatory and are applied on the next week's maintenance window at the earliest. This means you have at least 7 days advance notice with exception of critical security updates. These critical updates are applied in the maintenance window of the current week. During service upgrades, maintenance updates are automatically applied and do not require intervention from the user to be applied.
```
staceysalamon-aiven, line 25:
```suggestion
Security updates, platform updates that affect reliability or stability of the service nodes, and quarterly patch updates are always mandatory. Other updates are initially optional. Advance notice is given for all updates. After optional updates have been available for six months, they become mandatory and are applied on the next week's maintenance window at the earliest. This means you have at least 7 days advance notice with exception of critical security updates. These critical updates are applied in the maintenance window of the current week. During service upgrades, maintenance updates are automatically applied and do not require any action from you. 
```

staceysalamon-aiven: Reworded slightly to use plain and more direct language.

staceysalamon-aiven (changes requested):
Thanks for adding this important point to the docs! I've made one minor style suggestion.

staceysalamon-aiven (approved):
Looks great - thanks again!

Issue #1724, opened 2023-01-25
bug: 404 page is indexed by search


celestehorgan:
# What's wrong?

When using the search box on Aiven's devportal, the 404 Page not Found page is indexed and returns as a result: 

![image](https://user-images.githubusercontent.com/17999517/214515813-fe0b1071-d1b5-442d-9eac-3f2603197b40.png)

# Steps to remediation 

The general remediation for these is to [add a `noindex` tag to the 404 page](https://developers.google.com/search/docs/crawling-indexing/block-indexing).

The 404 page in Sphinx is [generated via a python package](https://sphinx-notfound-page.readthedocs.io/) (commit 4384416a8bc7fc8232ca0bdf1f99b77df55fbffe), so we don't have direct access to the page as we would in other static site generators.

According to `sphinx-notfound-page`'s [configuration docs](https://sphinx-notfound-page.readthedocs.io/en/latest/configuration.html), the 404 by default inherits its template from `page.html`. Since we can't add `noindex` to page.html (as that would un-index all our pages) we need to create a template specific to the notfound page to un-index.

The other half of this might be convincing the website's OpenSearch implementation to not index a specific page? I don't know nearly enough about OpenSearch to make that call though.

PR #1927, opened 2023-05-29
Preprocess sitemap exclude html extension

angelinekwan:
To resolve issue [#1691](https://github.com/aiven/devportal/issues/1691), the sitemap `<url>` should not have `.html` or `index.html `as those link are redirected to the non html url. This creates a redirect loop that prevents search engine from crawling the pages effectively.

This also resolved the issue [#1724](https://github.com/aiven/devportal/issues/1724) removing 404 from sitemap.

# Demo
https://preprocess-sitemap-exclude-h.devportal.pages.dev/sitemap.xml -> notice no more .html vs the [live sitemap](https://docs.aiven.io/sitemap.xml) which is an issue


conf.py
```diff
@@ -96,8 +96,8 @@
 html_favicon = './_static/images/favicon.ico'
 html_theme = 'furo'
 html_theme_options = {
-    "light_logo": "images/logoLightPride.png",
-    "dark_logo": "images/logoDarkPride.png",
+    "light_logo": "images/logoLight.png",
+    "dark_logo": "images/logoDark.png",
```
TibsAtWork, line 100: Not sure if this is meant - when do we undo the pride crab?

scripts/postprocess_sitemap.py
```diff
@@ -0,0 +1,25 @@
...
+        continue
+    if 'genindex.html' in text:
+        loc.string = text[:-5]  # removes the ".html"
+    elif text.endswith('index.html'):
+        loc.string = text[:-10]  # removes the "index.html"
+    elif text.endswith('.html'):
+        loc.string = text[:-5]  # removes the ".html"
+
+with open('./_build/html/sitemap.xml', 'w') as f:
+    f.write(str(soup))
```
TibsAtWork, line 25: The unhappy looking "stop" symbol at the end of the file probably means it's lacking a newline character on the last line. It would be better to add one.

scripts/postprocess_sitemap.py
```diff
@@ -0,0 +1,25 @@
+from bs4 import BeautifulSoup
+
+with open('./_build/html/sitemap.xml', 'r') as f:
+    contents = f.read()
+
+soup = BeautifulSoup(contents, 'lxml')
```
TibsAtWork, line 6: When I run this locally, the first time I make this call it says:
> It looks like you're parsing an XML document using an HTML parser. If this really is an HTML document (maybe it's XHTML?), you can ignore or filter this warning. If it's XML, you should know that using an XML parser will be more reliable. To parse this document as XML, make sure you have the lxml package installed, and pass the keyword argument `features="xml"` into the BeautifulSoup constructor.

If I pass in `'xml'` instead of `'lxml'` it doesn't give the warning:
```suggestion
soup = BeautifulSoup(contents, 'xml')
```
so it's probably worth making the change.

scripts/postprocess_sitemap.py
```diff
@@ -0,0 +1,25 @@
...
+urls = soup.find_all('url')
+
+for url in urls:
+    loc = url.find('loc')
+    text = loc.string
+    # Remove the 'gen' and '404' pages
+    if '404' in text:
+        url.decompose()
+        continue
+    if 'genindex.html' in text:
```
TibsAtWork, line 17: It would probably be better to be more specific that `genindex.html` needs to be at the end, since it then goes to strip off the last few characters (in *practice* I'm sure we'd never get `genindex.html` in the middle of a string, but it's probably better to be safer).
```suggestion
    if text.endswith('genindex.html'):
```

(but a good catch that this needs to be special-cased so we don't end up with `gen`)

TibsAtWork (commented):
This seems like a decent pragmatic approach.
It's a bit sad we can't just tell sphinx to do what we want, but I don't know how to do that either.

Once the comments are addressed, I think this will be good to go.

scripts/postprocess_sitemap.py
angelinekwan: Which command gives that warning? For me,`make html` or `make livehtml` doesn't give me any warning
Screenshot 2023-07-06 at 13 16 41

But we can update to `xml` since is for `sitemap.xml `and the XML is well-formed and doesn't have any peculiarities that might confuse an XML parser.

scripts/postprocess_sitemap.py
angelinekwan: I add `f.write('\n')` or just extra line in 26 but still seeing this "stop" symbol. Not an issue locally in my visual studio code. Any idea?

conf.py
angelinekwan: Was an accident commit :) Is now all good since we reverted back from Pride logo.

TibsAtWork (approved):
Thanks for sorting this one out!

scripts/postprocess_sitemap.py
angelinekwan: Works now :) was not earlier for some reason ;)

PR #1940, opened 2023-06-07
Update instructions for authentication

staceysalamon-aiven:
Updated instructions for authentication article to reflect the new navigation to the user profile.


rsalevsky (approved):

staceysalamon-aiven:
@rsalevsky Updated two files to reflect the most recent change in the menu (it now says Authentication instead of Authentication methods). This is ok to be published, correct?

PR #1941, opened 2023-06-07
Planned for end of June - Add instructions for managing org users

staceysalamon-aiven:
In an upcoming release it will be possible to add users directly to an organization to control their access to its units and projects. These instructions cover inviting and removing users from orgs.

### DIFF
--- a/docs/platform/howto/manage-org-users.rst
+++ b/docs/platform/howto/manage-org-users.rst
@@ -1,0 +1,59 @@
+Manage users in an organization
+================================
+
+.. important::
+    Organization users are available as a feature preview and must be :doc:`enabled in the user profile </docs/platform/howto/feature-preview>`.
+
+Adding users to your organization lets you give them access to specific organizational units, projects, and services within that organization. 
+
+Invite users to an organization
+---------------------------------
+
+To add users to your organization, send them an invite:
+
+#. Click **Admin**.
+
+#. Click **Users**.
+
+#. Click **Invite users**.
+
+#. Enter the email addresses of the people you want to invite. 
+
+#. Click **Invite users**.
+
+The users receive an email with instructions to sign up (for new users) and accept the invite.
+
+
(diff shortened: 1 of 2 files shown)


docs/platform/howto/manage-org-users.rst
```diff
@@ -0,0 +1,59 @@
+Manage users in an organization
+================================
+
+.. important::
+    Organization users are available as a :doc:`feature preview </docs/platform/howto/feature-preview>`.
```
rsalevsky, line 5: Can we add a sentence that the feature preview has enabled? Just to avoid support tickets that the menu item is not visible.

docs/platform/howto/manage-org-users.rst
```diff
@@ -0,0 +1,59 @@
...
+
+#. Click **Admin**.
+
+#. Click **Users**.
+
+#. Click **Invite users**.
+
+#. Enter the email addresses of the people you want to invite. 
+
+#. To automatically add these users to user groups, switch the toggle to **Yes** and select the groups.
```
rsalevsky, line 22: This part doesn't exist yet.

docs/platform/howto/manage-org-users.rst
```diff
@@ -0,0 +1,59 @@
...
+
+#. Click **Admin**.
+
+#. Click **Users**.
+
+#. Find the user that you want to remove and click the **Actions** menu. 
+
+#. Select **Remove**.
+
+#. Confirm you want to remove the user by clicking **Remove user**.
```
rsalevsky, line 42: Let's add a sentence that the user will be removed from all teams and projects and will have no longer access to any resource in the Org.

rsalevsky (changes requested):

rsalevsky (approved):

harshini-rangaswamy (approved):
LGTM!

PR #1951, opened 2023-06-13
Add instructions for managing domains

staceysalamon-aiven:
New article with instructions for adding and removing domains for the release.

### DIFF
--- a/docs/platform/howto/manage-domains.rst
+++ b/docs/platform/howto/manage-domains.rst
@@ -1,0 +1,95 @@
+Manage domains
+===============
+
+Verified domains let you manage users in your organization and set up security policies.
+
+There are two ways you can verify a domain:
+
+* by adding a DNS TXT record to the domain (recommended)
+* by uploading an HTML file to your website
+
+.. note::
+    After adding a domain, your organization users will automatically become managed users. 
+
+Add a domain using a DNS TXT record
+-------------------------------------
+
+1. In the organization you want to add a domain to, click **Admin**.
+
+2. Click **Domains**.
+
+3. Click **Add domain**.
+
+4. Enter a **Domain name**.
+
+5. In the **Verification method**, select **Add a DNS TXT record to your domain host**.
+
+6. Click **Add domain**.
+
+7. In the **Verification method** column, click **DNS TXT record**.
+
+8. Copy the TXT record value.
+
+9. In another browser tab or window, log in to your domain hosting provider.
+
+10. Go to the DNS settings.
+
+11. In the DNS settings for your domain provider, create a TXT record with the following:
+
+.. list-table::
+      :header-rows: 1
+      :align: left
+
+      * - Field name
+        - Value
+      * - Name
+        - ``_aiven-challenge.{your domain}``
+      * - Record value 
+        - The TXT record value you copied in the format ``token=<random_string>,expiry=never``
+      * - Type
+        - ``TXT``
+
+12. In the Aiven Console, open the **Actions** menu and click **Verify**.
+
+It can take up to 72 hours for your DNS records to update the domain to be verified. If the domain is still not verified after that time, you can retry it by repeating the last step.
+
+
+Add a domain using an HTML file 
+--------------------------------
+
+#. In the organization that you want to add a domain to, click **Admin**.
+
+#. Click **Domains**.
+
+#. Click **Add domain**.
+
+#. Enter a **Domain name**.
+
+#. In the **Verification method**, select **Upload an HTML file to your website**.
+
+#. Click **Add domain**.
+
+#. In the **Verification method** column, click **HTML file upload**.
+
+#. Download the HTML file.
+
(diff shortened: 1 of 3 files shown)


docs/platform/howto/manage-domains.rst
```diff
@@ -0,0 +1,87 @@
...
+.. list-table::
+      :header-rows: 1
+      :align: left
+
+      * - Field name
+        - Value
+      * - Name
+        - ``_aiven-challenge.{your domain}``
+      * - Value 
+        - The TXT record value shown in Aiven Console.
```
italomg, line 44: I like refering to it as "record value" as it is probably to most correct, but it is possible that even technical folks might not be super familiar with it, I wonder if it would make sense to provide the general format of it like `token=<random_string>,expiry=never`. I don't think that is necessary though, just an idea.

docs/platform/howto/manage-domains.rst
```diff
@@ -0,0 +1,87 @@
...
+      * - Type
+        - ``TXT``
+
+#. In the Aiven Console, click **Add domain**.
```
italomg, line 48: After the domain is setup the customer will have to click the "three dots" in the row of the domain they want to verify and there will be a button with text "Verify", that's the one to be pressed to trigger the verification itself.

docs/platform/howto/manage-domains.rst
```diff
@@ -0,0 +1,87 @@
...
+
+It can take up to 72 hours for your DNS records to update the domain to be verified.
```
italomg, line 50: this is true, but we will not retry automatically so the customer must come back at a later time to do this.

docs/platform/howto/manage-domains.rst
```diff
@@ -0,0 +1,87 @@
...
+
+
+Add a domain using an HTML file 
+--------------------------------
+
+#. In the organization that you want to remove a domain from, click **Admin**.
```
italomg, line 56: I believe this would be as in the previous section
```suggestion
#. In the organization you want to add a domain to, click **Admin**.
```

docs/platform/howto/manage-domains.rst
```diff
@@ -0,0 +1,87 @@
...
+
+#. Enter a **Domain name**.
+
+#. In the **Verification method**, select **Upload an HTML file to your website**.
+
+#. Download the HTML file.
+
+#. In another browser tab or window, upload the HTML file to your website's root directory.
+
+#. In the Aiven Console, click **Add domain**.
```
italomg, line 70: same as for the DNS instructions, this should probably go a little up in the list and have a click "Verify" here

docs/platform/howto/manage-domains.rst
```diff
@@ -0,0 +1,87 @@
...
+
+#. Click **Add domain**.
+
+#. Enter a **Domain name**.
+
+#. In the **Verification method**, select **Upload an HTML file to your website**.
+
+#. Download the HTML file.
+
+#. In another browser tab or window, upload the HTML file to your website's root directory.
```
italomg, line 68: In this case we probably don't have to mention "In another browser or window", depending on the customer size and their technical processes this flow will be drastically different, but I think most commonly it will include some kind of version control system like git for example.
Also, it would be useful to have the actual path we want it to be, so something like "upload the HTML file to your website in the path <domain>/.well-known/aiven/<html_file>".

docs/platform/howto/manage-domains.rst
```diff
@@ -0,0 +1,95 @@
...
+
+8. Copy the TXT record value.
+
+9. In another browser tab or window, log in to your domain hosting provider.
+
+10. Go to the DNS settings.
+
+11. In the DNS settings for your domain provider, create a TXT record with the following:
+
+.. list-table::
```
italomg, line 39: this is great!

docs/platform/howto/manage-domains.rst
```diff
@@ -0,0 +1,95 @@
...
+
+#. In the **Verification method**, select **Upload an HTML file to your website**.
+
+#. Click **Add domain**.
+
+#. In the **Verification method** column, click **HTML file upload**.
+
+#. Download the HTML file.
+
+#. Upload the HTML file to your website in the path ``/.well-known/aiven``.
```
italomg, line 76: Actually the file name here is important(I should probably check it in the console as well). The file name should have the format `aiven-code-<challenge_token_prefix>`
Where `challenge_token_prefix` is the first 6 digits of the challenge token. When the customer download the file from the console it should be downloaded with the proper name, so having it here is probably not necessary but it wouldn't hurt either.

italomg (approved):
I left one last comment, but it is more of a nicety/convenience for the customer, not really a blocker, since the customer will probably always have the HTML file properly named anyway.

rsalevsky (changes requested):
Can we please add information that this is currently in preview.

staceysalamon-aiven:
New PR #2339 opened for this.

PR #1963, opened 2023-06-20
Added note to inform users to check ACLs so MirrorMaker will work.

richard-joerger-aiven:

I added information about making sure that the ACLs for Kafka & MM are correct so that customers know that they have to provide access to these clusters to each user. This will help customers be more self service and deflect cases from support.


harshini-rangaswamy (approved):
Thanks for adding this note.

PR #1970, opened 2023-06-21
SUP-675: PG nodes when a plan change happens

bcrisologo:
Under the [PostgreSQL Controlled switchover during upgrades or migrations](https://docs.aiven.io/docs/products/postgresql/concepts/upgrade-failover#controlled-switchover-during-upgrades-or-migrations), we can add a **note** indicating what happens with the following plan changes where node count changes:
- Business plan (2 nodes) => Startup plan (1 node)
- Startup plan (1 node) => Business plan (2 nodes)

We had a customer asking to confirm what happens and it would be good to have a public documentation explaining what happens to the nodes during the plan change where the number of nodes do change.


docs/products/postgresql/concepts/upgrade-failover.rst
```diff
@@ -58,6 +58,9 @@ During maintenance updates, cloud migrations, or plan changes, the below procedu
 .. Note::
     The old primary server is kept alive for a short period of time (minimum 60 seconds) with a TCP forwarding setup pointing to the new primary server allowing clients to connect before learning the new IP address.
 
+.. Note::
+    If the service plan is changed from a business plan that has two (2) nodes to a startup plan which only has one (1) node, the standby node will be removed while the primary node will be retained and connections to the primary are not affected by the downgrade.  Similarly, upgrading the service plan from startup to business will simply add a standby node to the service cluster and connections to the primary node are unaffected.
```
wojcik-dorota, line 62: changing from the future tense to the present tense since we're talking about a general truth here, something that is always right + a few minor changes:

 If the service plan is changed from a business plan that has two nodes to a startup plan which only has one (1) node, the standby node is removed while the primary node is retained, and connections to the primary are not affected by the downgrade.  Similarly, upgrading the service plan from a startup one to a business one simply adds a standby node to the service cluster, and connections to the primary node are unaffected.

wojcik-dorota (changes requested):

bcrisologo:
I need to make slight changes to this - the above is only true when the business plan and startup plan are the same (e.g. business-8 => startup-8).  If they select a different plan grade, it will cause a node replacement because of the difference in hardware.  Making additional changes for review.

docs/products/postgresql/concepts/upgrade-failover.rst
```diff
@@ -58,6 +58,9 @@ During maintenance updates, cloud migrations, or plan changes, the below procedu
 .. Note::
     The old primary server is kept alive for a short period of time (minimum 60 seconds) with a TCP forwarding setup pointing to the new primary server allowing clients to connect before learning the new IP address.
 
+.. Note::
+    If the service plan is changed from a business plan that has two nodes to a startup plan which only has one (1) node of the same tier (e.g. business-8 to startup-8), the standby node is removed while the primary node is retained, and connections to the primary are not affected by the downgrade. Similarly, upgrading the service plan from a startup one to a business one simply adds a standby node to the service cluster, and connections to the primary node are unaffected.
```
wojcik-dorota, line 62: A few final tiny things:
1. Remove ```(1)```
2. Replace ```e.g.``` with ```for example,```
3. Remove ```simply```

wojcik-dorota (changes requested):

wojcik-dorota (approved):

PR #1994, opened 2023-07-11
Update Kafka custom datadog metrics example

jsnowfreedman:
I'll be honest, I'm not sure if this is a bug in the CLI, or just mis-documented here. 
The issue arrises in the fact that in the CLI, array value parameters are processed as Python literals [cli.py#L77](https://github.com/aiven/aiven-client/blob/0366b6b0eb78cab430168d7dcbc4e28faf3406bb/aiven/client/cli.py#L77)

This means that either the code there needs to be update, or the documentation needs to be updated. If you try to run these examples as is, it always seems to give a very unobviouse parsing error.

I assumed it would be easiest to fix the documentation (and have less side effects to others who may have already figured this out), however if it would be better to raise a PR in the client to fix the layout, please let me know and I would be happy to do so.

### DIFF
--- a/docs/products/kafka/howto/datadog-customised-metrics.rst
+++ b/docs/products/kafka/howto/datadog-customised-metrics.rst
@@ -35,7 +35,7 @@ Variable               Description
 Customise Apache Kafka® metrics sent to Datadog
 -----------------------------------------------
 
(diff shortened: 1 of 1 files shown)


harshini-rangaswamy (approved):
@jsnowfreedman Thank you for bringing up this issue and for your contribution to the documentation. I appreciate your effort in trying out the CLI and identifying this discrepancy. I have also confirmed with our development team that when we specify ``include_topics`` as part of a user config it is always an array.

PR #2010, opened 2023-07-19
Fixing EOL for OpenSearch 1.x

jkowall:
Per upstream which I have fixed, 1.x EOL is not announced and not planned until 3.x comes out.


harshini-rangaswamy (approved):
LGTM

PR #2012, opened 2023-07-20
postgresql: add info on logical replication restrictions

wojcik-dorota:
https://aiven.atlassian.net/browse/DOC-438


gordonca (changes requested):
LGTM, just it might be better to use `current` instead of `15` for the version in the url, so we don't need to change in the future.

wojcik-dorota:
> LGTM, just it might be better to use `current` instead of `15` for the version in the url, so we don't need to change in the future.

Thanks @gordonca , it's done :-)

gordonca (approved):

PR #2013, opened 2023-07-20
Fix list in the create service topic

harshini-rangaswamy:

Fixed the list in the create service topic. 
image


wojcik-dorota (changes requested):
I propose to keep the ordered list here coz it does matter in what order the user takes those steps.

harshini-rangaswamy:
> I propose to keep the ordered list here coz it does matter in what order the user takes those steps.

In the current topic page, the options in the Overview page are added as a sub-list, and numbered list within a numbered list, in my opinion, doesn't look good (if you see it in the screenshot attached to the PR).

PR #2037, opened 2023-07-26
Update feature preview note to new template

staceysalamon-aiven:
Update language in the feature preview note in the org users article to match the new standardized wording.


harshini-rangaswamy (approved):
LGTM!

PR #2059, opened 2023-08-01
Added a warning that slow query log will not work on a read-only repl…

josh-martin-aiven:
…ica. Reference: https://aiven.slab.com/posts/00048962-tqpunb8kX

Added this warning to the slow query log doc. If this is enabled on read only replicas, mysql will throw an error when trying to write to the mysql.slow_log table.

There is currently an idea submitted on the idea forum requesting this functionality, but in the meantime, we should inform customers of this limitation. 
https://ideas.aiven.io/forums/951280-operational-databases/suggestions/46266829-allow-mysql-slow-query-log-on-read-only-replicas

Let me know if you have any questions or recommendations. :)


docs/products/mysql/howto/enable-slow-queries.rst
```diff
@@ -24,6 +25,11 @@ Follow these steps to enable your slow queries in your Aiven for MySQL service v
 
 Your Aiven for MySQL service can now log slow queries. If you want to simulate slow queries to check this feature, check the next section.
 
+.. warning::
+
+    Aiven managed MySQL services use table output for the slow query log.
+    This means that the slow query log will not function on a read-only replica.
```
wojcik-dorota, line 31: Let's simplify it a little:
The  slow query logging is not supported on read-only replicas because Aiven for MySQL services use table output for the slow query log.

wojcik-dorota:
@josh-martin-aiven , many thanks for adding this! I suggest moving the warning upwards just after the 1st paragraph in the article.

josh-martin-aiven:
Thanks for the review @wojcik-dorota. I've made the recommended changes. Let me know your thoughts. :)

docs/products/mysql/howto/enable-slow-queries.rst
```diff
@@ -3,6 +3,10 @@ Enable slow query logging
 
 You can identify inefficient or time-consuming queries by enabling `slow query log <https://dev.mysql.com/doc/refman/5.7/en/slow-query-log.html>`_ in your Aiven for MySQL® service. In this article, you'll find out how to enable slow queries in your Aiven for MySQL service.
 
+.. warning::
+
+    The slow query log is not supported on read-only replicas because Aiven for MySQL services use table output for the slow query log.
```
wojcik-dorota, line 8: Thanks, @josh-martin-aiven ! Apologies for digging into this but would it be still possible to make it more clear for users who are not familiar with data management and MySQL: why does this table output prevent the slow query log from working on read-only replicas? Please help me understand. Is it that the table output is not supported on read-only replicas or what is the actual reason?

josh-martin-aiven: I'm happy to expound further! These are the case notes from the support ticket which spawned this change. 
https://aiven.slab.com/posts/00048962-tqpunb8k  


Excerpt below for reference. 
The slow query log can have one of two output destinations. Log file, and the mysql.slow_log table on the server itself. Docs for reference
https://dev.mysql.com/doc/refman/8.0/en/log-destinations.html
Log Output:
This currently cannot be configured for Aiven managed Mysql services. This is because customers do not have access to the underlying filesystem for service nodes.
Table Output:
This is the output destination used by Aiven hosted Mysql services, for the same reasoning above.
This means that when an Aiven managed Mysql service enables the slow query log through advanced config, the output destination will be the mysql.slow_log table.
https://docs.aiven.io/docs/products/mysql/reference/advanced-params#slow-query-log

This is a problem for read replicas, because the mysql.slow_log table cannot be written to on read-only hosts. The following error is produced in such cases on the read replica server when the slow query log advanced config is enabled. 

"[Server] Failed to write to mysql.slow_log: The MySQL server is running with the --read-only option so it cannot execute this statement (cannot open table for slow log)"

For these reasons, the slow query log cannot be successfully be enabled on a read-only replica. 

The resolution on this is two-fold.
1. Feature request to add functionality. 
https://ideas.aiven.io/forums/951280-operational-databases/suggestions/46266829-allow-mysql-slow-query-log-on-read-only-replicas%C2%A0
2. Docs change to inform customers. (This PR) 

Initially, I had the impression that this behavior was a bug, but my colleagues in support advised that this would be better viewed as a limitation/feature request. Let me know your thoughts. :)

wojcik-dorota: Thanks so much! this clarifies it all :-) How about we reword this warning into the following:

```
Since the output of the slow query log is written  to the ``mysql.slow_log`` table on a particular server, the slow query logging is not supported on read-only replicas, which don't allow any writes.
```

josh-martin-aiven: Sounds good. I just updated the wording. Let me know if you have any other thoughts/questions, etc. :)

josh-martin-aiven: I think we have a false positive for the Build HTML check. It's flagging the table name for stylization purposes. 
[Aiven.common_replacements] Use 'MySQL' instead of 'mysql'.

Not sure if there's anything I can do to help that, but let me know if there is. :)

wojcik-dorota: Thanks, @josh-martin-aiven ! to fix that, please use double quotes (\``mysql.slow_log\``).

josh-martin-aiven: It looks like it's still failing the check with the double quotes. Any ideas on how to handle that?

wojcik-dorota: Please compare \''mysql.slow_log\'' to \``mysql.slow_log\``.
Please try to use the latter ones.

josh-martin-aiven: Ah got it. Looks like the double ticks worked.

wojcik-dorota (approved):
Thank you :-)

PR #2080, opened 2023-08-08
Add Kafka 3.5 dates & fix Kafka 3.4 dates

tvainika:

Add Kafka 3.5 dates & fix Kafka 3.4 dates

Kafka 3.4 dates were a bit off (start had wrong year) compared to our internal configuration.


harshini-rangaswamy (approved):
Thanks for adding this!

PR #2089, opened 2023-08-14
adding DR based read-replica details

cybermax-008:
While we do not support native HA plan with PostgreSQL with multi region setup, we can still have an HA plan among available AZs plus a remote read replica in a separate region/cloud which might be suitable for customers to guarantee reliability.
Added the above info to our current PG read-replica KB.


docs/products/postgresql/howto/create-read-replica.rst
```diff
@@ -57,3 +57,7 @@ If the above command returns ``TRUE`` if you are connected to the replica, and `
 .. Warning::
 
     Aiven for PostgreSQL uses asynchronous replication and so a small lag is expected. When running an ``INSERT`` operation on the primary node, a minimal delay (usually less than a second) can be expected for the change to be propagated to the replica and to be visible there.
+
+Read-replica for disaster recovery
+----------------------------------
+In case of high-availability, distribution occurs across the available availability zones within a single region. While a default multi-region service, with node allocation spanning multiple regions, is not provided, this capability can still be attained. This can be accomplished by initially establishing a high-availability PostgreSQL service within a single region. Subsequently, a remote read replica can be configured in a different region or even on an alternate cloud platform.This approach introduces an additional node in the distinct region/cloud, although it does not function as a hot standby node. It can, however, be manually promoted to a "primary" role, operating as an independent standalone service.
```
wojcik-dorota, line 63: A few minor changes proposed:

High availability enables data distribution across availability zones within a single region. To accomplish that without a default multi-region service with node allocation spanning multiple regions, you can take the following steps:

1. Establish a high-availability Aiven for PostgreSQL service within a single region.
2. Configure a remote read-only  replica in a different region or even on an alternate cloud platform.

As a result, you introduce an additional node in the distinct region/cloud. Since this node does not work as a hot standby node, you might want to promote it manually to the primary role, which makes it operate as an independent standalone service.

wojcik-dorota (changes requested):

docs/products/postgresql/howto/create-read-replica.rst
cybermax-008: That looks neat and better. Thanks @wojcik-dorota . I have made the changes.

wojcik-dorota (approved):
Thanks :-)

PR #2096, opened 2023-08-16
Updated startup-2 plan reference in the article

harshini-rangaswamy:

As per the feedback received,  the alert was thought to only fire for the startup plan, but it will actually fire for any plans with a CPU less than or equal to 2. It can also occur in the business plan, but this is a very rare case. 

Update the article to remove reference to the Statup-2 plan and make it more generic such as any service plans with a CPU less than or equal to 2.


sebinsunny (approved):
LGTM

PR #2106, opened 2023-08-22
Add instructions for changing to a new IdP

staceysalamon-aiven:
Add instructions on changing to a new IdP so it's clear that org users must be linked to the new provider.


cmuir363 (approved):
LGTM

PR #2109, opened 2023-08-24
kafka: fix the article title to the correct MQTT => Kafka connector orientation

wojcik-dorota:
Fixing the title


staceysalamon-aiven (approved):
Thanks for doing this!

PR #2144, opened 2023-09-13
clickhouse: remove "data upload from S3" from the clickhouse limitations doc

wojcik-dorota:
Removing the limitation since it's supported now

https://aiven.atlassian.net/browse/DOC-518


harshini-rangaswamy (approved):
Looks good!

PR #2148, opened 2023-09-15
Add documentation for super admin 

staceysalamon-aiven:
# What changed, and why it matters
Super admin will replace account owners. Added a brief explanation of the super admin role and instructions on how to make/revoke super admin privileges.

PR #2145, opened 2023-09-13
Add info on automated migration from teams to groups

staceysalamon-aiven:
# What changed, and why it matters
Added information about the migration from teams to groups and what to expect.

PR #2157, opened 2023-09-26
Add instructions to manually migrate teams to groups

staceysalamon-aiven:

Relies on #2148.
 
Added instructions for how customers can migrate their existing teams to the new groups feature as [discussed in another PR](https://github.com/aiven/devportal/pull/2145#discussion_r1337167541).


docs/tools/aiven-console/howto/create-manage-teams.rst
```diff
@@ -54,3 +56,24 @@ To add projects and roles to a team:
...
+------------------------
+
+To get started using organization groups, replace your existing teams with groups:
+
+#. In the organization, click **Admin** 
+
+#. Click **Organization** and on the **Teams** tab view each team to make a note of: 
+
+* which users are members of the team
+* which projects the team is assigned to
```
rsalevsky, line 72: This and the point below are more naturally done after creating the group. Therefore I suggest to move this after "Enter the name of one of the teams and assign the same users to this group. Do this for each of your organization's teams." and add an extra step which says navigate the the project.

rsalevsky (changes requested):

docs/tools/aiven-console/howto/create-manage-teams.rst
staceysalamon-aiven: I think this process is a bit more efficient, though?  It's easier to get all of the information about a team from Admin > Org > Teams  instead of navigating to each project individually. Also, when they create the group they can add the team members right away instead of doing it in another step.

staceysalamon-aiven: It also slightly duplicates the instructions on the linked page if we tell them to go to the project.

rsalevsky (approved):

PR #2160, opened 2023-09-27
platform: how to get connection info for services using multiple privatelink connections

wojcik-dorota:
https://aiven.atlassian.net/browse/DOC-541

### DIFF
--- a/docs/platform/howto/use-aws-privatelinks.rst
+++ b/docs/platform/howto/use-aws-privatelinks.rst
@@ -177,7 +177,7 @@ To acquire connection information for your service component using AWS PrivateLi
   * SERVICE_NAME for Aiven for Apache Kafka®, for example, can be ``kafka-12a3b4c5``.
   * PRIVATELINK_CONNECTION_ID can be ``plc39413abcdef``.
 
-* For connection information for your service component using AWS PrivateLink with SASL enabled, run the following command:
+* For SASL connection information for Aiven for Apache Kafka® service components using AWS PrivateLink, run the following command:
 
 .. code-block:: bash
 
--- a/docs/platform/howto/use-aws-privatelinks.rst
+++ b/docs/platform/howto/use-aws-privatelinks.rst
@@ -150,7 +150,7 @@ Acquire connection information
 One AWS PrivateLink connection
 ''''''''''''''''''''''''''''''
 
-If you have one private endpoint connected to your Aiven service, you can preview the connection information (URI, hostname, or port required to access the service through the private endpoint) in `Aiven Console <https://console.aiven.io/>`_ > the service's **Overview** page > the **Connection information** section, where you'll also find the toggle switch for the ``privatelink`` access route. AWS PrivateLink values for ``host`` and ``port`` differ from those for the default ``dynamic`` access route used to connect to the service.
+If you have one private endpoint connected to your Aiven service, you can preview the connection information (URI, hostname, or port required to access the service through the private endpoint) in `Aiven Console <https://console.aiven.io/>`_ > the service's **Overview** page > the **Connection information** section, where you'll also find the switch for the ``privatelink`` access route. ``privatelink``-access-route values for ``host`` and ``port`` differ from those for the ``dynamic`` access route used by default to connect to the service.
 
 .. note::
 
--- a/docs/platform/howto/use-aws-privatelinks.rst
+++ b/docs/platform/howto/use-aws-privatelinks.rst
@@ -172,7 +172,7 @@ Each endpoint (connection) has PRIVATELINK_CONNECTION_ID, which you can check us
 
    avn service connection-info UTILITY_NAME SERVICE_NAME -p PRIVATELINK_CONNECTION_ID
 
-Where
+.. topic:: Where
 
   * UTILITY_NAME is ``kcat``, for example
   * SERVICE_NAME is ``kafka-12a3b4c5``, for example
@@ -184,7 +184,7 @@ Where
 
    avn service connection-info UTILITY_NAME SERVICE_NAME -p PRIVATELINK_CONNECTION_ID -a sasl
 
-Where
+.. topic:: Where
 
   * UTILITY_NAME is ``kcat``, for example
   * SERVICE_NAME is ``kafka-12a3b4c5``, for example
--- a/docs/platform/howto/use-aws-privatelinks.rst
+++ b/docs/platform/howto/use-aws-privatelinks.rst
@@ -142,13 +142,15 @@ currently support AWS PrivateLink.
    successful heartbeats before they transition from the ``initial``
    state to ``healthy`` and are included in the active forwarding rules of the load balancer.
 
-   | **Note:** Currently, you can only create one VPC endpoint for each
-     Aiven service.
-
 .. _h_b6605132ff:
 
-Connection information
-----------------------
+Acquire connection information
+------------------------------
+
+.. _one-connection:
+
+One AWS PrivateLink connection
+''''''''''''''''''''''''''''''
 
 Once you have enabled PrivateLink access for a service component, a
 switch for the ``privatelink`` access route appears under **Connection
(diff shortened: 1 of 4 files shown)


docs/platform/howto/use-azure-privatelink.rst
```diff
@@ -161,22 +161,69 @@ To enable Private Link access in `Aiven Console <https://console.aiven.io/>`_:
...
+
+One Azure Private Link connection
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+If you have Azure Private Link enabled for one service component, you can preview its connection information (URI, hostname, or port required to access the service through the private endpoint) on the service's **Overview** page in `Aiven Console <https://console.aiven.io/>`_.
+
+Multiple Azure Private Link connections
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+If you have more than one Azure Private Link connection, you can get connection information for the first endpoint as described in :ref:`One Azure Private Link connection <one-private-service-connection>` (from `Aiven Console <https://console.aiven.io>`__). For connection information for the remaining endpoints, you need to use CLI.
```
mlatief, line 183: should we use cloud provider terminology or ours? for instance "Azure private endpoint" rather than "Azure Private Link connection". 

Also ```:ref:`One Azure Private Link connection <one-private-service-connection>` ``` links to Google PSC.

docs/platform/howto/use-azure-privatelink.rst
```diff
@@ -161,22 +161,69 @@ To enable Private Link access in `Aiven Console <https://console.aiven.io/>`_:
...
+
+Each endpoint (connection) has PRIVATELINK_CONNECTION_ID, which you can check using the :doc:`avn service privatelink azure connection list SERVICE_NAME </docs/tools/cli/service/privatelink>` command.
+
+To acquire connection information for your service component using Azure Private Link, run the :doc:`avn service connection-info </docs/tools/cli/service/connection-info>` command.
+
+* For SSL connection information for your service component using Azure Private Link, run the following command:
+
+.. code-block:: bash
+
+   avn service connection-info UTILITY_NAME SERVICE_NAME -p PRIVATELINK_CONNECTION_ID
```
mlatief, line 193: FYI: `-p` is a shorthand for `--privatelink-connection-id` if there is a preference in our docs.

docs/platform/howto/use-aws-privatelinks.rst
```diff
@@ -142,25 +142,63 @@ currently support AWS PrivateLink.
...
+  * PRIVATELINK_CONNECTION_ID is ``plc39413abcdef``, for example
+
+.. note::
 
-Once you have enabled PrivateLink access for a service component, a
-switch for the ``privatelink`` access route appears under **Connection
-information** on the **Overview** page in `Aiven Console <https://console.aiven.io>`__. The ``host`` -
-and for some service components such as Kafka, ``port`` - values differ
-from the default ``dynamic`` access route that is used to connect to the
-service. You can use the same credentials with any access route.
```
mlatief, line 158: > The ``host`` - and for some service components such as Kafka, ``port`` - values differ from the default ``dynamic`` access route that is used to connect to the service. You can use the same credentials with any access route.

I would not omit this part, that is already more informative than my suggestion in the ticket.

docs/platform/howto/use-aws-privatelinks.rst
```diff
@@ -142,25 +142,63 @@ currently support AWS PrivateLink.
...
+.. code-block:: bash
+
+   avn service connection-info UTILITY_NAME SERVICE_NAME -p PRIVATELINK_CONNECTION_ID -a sasl
+
+.. topic:: Where
+
+  * UTILITY_NAME is ``kcat``, for example
+  * SERVICE_NAME is ``kafka-12a3b4c5``, for example
+  * PRIVATELINK_CONNECTION_ID is ``plc39413abcdef``, for example
+
```
mlatief, line 193: These are Kafka specific, may be have them rather as an example for Kafka? they are pretty useful though so I won't omit them either.

docs/platform/howto/use-aws-privatelinks.rst
```diff
@@ -142,25 +142,63 @@ currently support AWS PrivateLink.
...
+Each endpoint (connection) has PRIVATELINK_CONNECTION_ID, which you can check using the :doc:`avn service privatelink aws connection list SERVICE_NAME </docs/tools/cli/service/privatelink>` command.
+
+To acquire connection information for your service component using AWS PrivateLink, run the :doc:`avn service connection-info </docs/tools/cli/service/connection-info>` command.
+
+* For SSL connection information for your service component using AWS PrivateLink, run the following command:
+
+.. code-block:: bash
+
+   avn service connection-info UTILITY_NAME SERVICE_NAME -p PRIVATELINK_CONNECTION_ID
```
mlatief, line 174: FYI: `-p` is a shorthand for `--privatelink-connection-id` if that is more preferred in docs.

docs/platform/howto/use-aws-privatelinks.rst
```diff
@@ -142,25 +142,63 @@ currently support AWS PrivateLink.
...
+.. _one-connection:
+
+One AWS PrivateLink connection
+''''''''''''''''''''''''''''''
+
+Once you have enabled AWS PrivateLink access for a service component, the toggle switch for the ``privatelink`` access route is available in the **Connection information** section on the **Overview** page in `Aiven Console <https://console.aiven.io>`__. Values for ``host`` and ``port`` (for some service components, such as Kafka) differ from those for the default ``dynamic`` access route used to connect to the service.
+
+.. note::
+
+   You can use the same credentials with any access route.
```
mlatief, line 159: This is also valid for multiple connections/endpoints.

docs/platform/howto/use-azure-privatelink.rst
```diff
@@ -161,22 +161,69 @@ To enable Private Link access in `Aiven Console <https://console.aiven.io/>`_:
...
 
+Acquire connection information
+------------------------------
+
+.. _one-privatelink-connection:
+
+One Azure Private Link connection
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+If you have Azure Private Link enabled for one service component, you can preview its connection information (URI, hostname, or port required to access the service through the private endpoint) on the service's **Overview** page in `Aiven Console <https://console.aiven.io/>`_.
```
mlatief, line 178: > If you have Azure Private Link enabled for one service component
I would rather say something like: "If you have one private endpoint connected to Aiven service". It doesn't depend on how many components but rather how many endpoints.

docs/platform/howto/use-google-private-service-connect.rst
```diff
@@ -176,8 +176,52 @@ To enable Private Link access in `Aiven Console <https://console.aiven.io/>`_, t
...
 
+Acquire connection information
+------------------------------
+
+.. _one-private-service-connection:
+
+One Private Service Connect connection
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+If you have Private Service Connect enabled for one service component, you can preview its connection information (URI, hostname, or port required to access the service through the private endpoint) on the service's **Overview** page in `Aiven Console <https://console.aiven.io/>`_.
```
mlatief, line 187: same comment about one service component as in Azure.

docs/platform/howto/use-aws-privatelinks.rst
```diff
@@ -142,25 +142,63 @@ currently support AWS PrivateLink.
...
+
+Multiple AWS PrivateLink connections
+''''''''''''''''''''''''''''''''''''
+
+If you have more than one AWS PrivateLink connection, you can get connection information for the first endpoint as described in :ref:`One AWS PrivateLink connection <one-connection>` (from `Aiven Console <https://console.aiven.io>`__). For connection information for the remaining endpoints, you need to use CLI.
```
mlatief, line 164: I would omit the console reference for first endpoint, the CLI can be used to retrieve all connection info even for one endpoint. Showing all connection info on Console will be added at some point.

mlatief (commented):
This looks great, there are just few comments and meanwhile I will test these instructions for the three clouds and at least both Kafka and PG services.

docs/platform/howto/use-aws-privatelinks.rst
wojcik-dorota: It's not omitted :-) It's actually still there - I just reworded it a little bit, which is why GitHub shows this part as removed. But if you look just below, you'll see there's a new part added saying exactly the same using slightly different wording.

docs/platform/howto/use-azure-privatelink.rst
wojcik-dorota: We can use both. I believe the users are aware with the concept of Azure Private Link at this point in this article so introducing term "Azure private endpoint" and using both in parallel should not confuse them.

mlatief (approved):
Looks great

PR #2235, opened 2023-11-07
Kafka tiered storage: minor improvements

juha-aiven:
Change Google Cloud to Google Cloud Storage.

Make naming Kafka brokers consistent, always call them Kafka brokers.

Clarify the encryption keys are on aiven management machines and on the Kafka brokers.


harshini-rangaswamy (approved):
Thanks for these changes! Looks good.

PR #2236, opened 2023-11-07
Update free plan docs - Aiven for MySQL does not have connection pooling

carabaestlein:

Currently, the docs state that we do not offer connection pooling on free MySQL plans. While this is technically correct, it is misleading given we do not offer it on paid plans either. Therefore it should be removed and we should only state that we do not offer connection pooling for PostgreSQL on free plans, i.e. pgBouncer is not included.


staceysalamon-aiven (approved):
Thank you for doing this! LGTM

PR #2241, opened 2023-11-13
platform: update upcloud vpc peering how-to  - add a requirement to renew a dhcp lease

wojcik-dorota:
https://aiven.atlassian.net/browse/DOC-614

### DIFF
--- a/docs/platform/howto/vpc-peering-upcloud.rst
+++ b/docs/platform/howto/vpc-peering-upcloud.rst
@@ -146,17 +146,19 @@ You only need to take this step if any of your VMs has been created before setti
   
   A peering connection between an Aiven VPC and VMs created before the peering setup won't work unless you refresh the DHCP lease for a relevant network interface.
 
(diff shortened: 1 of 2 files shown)


docs/platform/howto/vpc-peering-upcloud.rst
```diff
@@ -137,6 +137,27 @@ HTTP status       Error code               Description
...
+
+.. code-block:: bash
+
+    dhclient -r NETWORK_INTERFACE_NAME
+
+or
+
+.. code-block:: bash
+
+    dhclient NETWORK_INTERFACE_NAME
```
giacomo-alzetta-aiven, line 159: They actually need to do both. The first comment "deletes" the old dhcp lease, and the second command forces a renew.

giacomo-alzetta-aiven (changes requested):
Only change needed is that customers need to run both `dhclient` commands to properly renew the lease.

giacomo-alzetta-aiven (approved):
LGTM!

PR #2259, opened 2023-11-17
turn `::` into `.. code::`

ArthurFlageul:

Using explicit code tag.

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -197,7 +197,7 @@ Howto article template
 
 Title template: Start with a verb (e.g. *Connect with Go*, *Install or upgrade an extension*).
 
-::
+.. code::
 
     Article title
     #############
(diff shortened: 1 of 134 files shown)


wojcik-dorota (commented):
For some reason, IIRC we agreed to use `.. code-block::`, which was requested by DevRels actually back then. Not sure why not  `.. code::``.  As a result, I've been using `.. code-block::` consistently for a few months now.

harshini-rangaswamy:
> For some reason, IIRC we agreed to use `.. code-block::`, which was requested by DevRels actually back then. Not sure why not `.. code::`. As a result, I've been using `.. code-block::` consistently for a few months now.

I often ran into issues using `..code-block::` and resorted to using just `..code::`.

ArthurFlageul:
I've checked the output for these pages, there's no problem as far as I can tell. The main thing is to not use `::` which breaks the conversion to MD 🙈

PR #2285, opened 2023-11-23
customer facing recommendation to prevent KStream data write issue

cybermax-008:
This article is targeted to the customers who might be using KStream 3.6.0 version. It contains details of the issue and solution for the same. 
More details: https://aiven-io.slack.com/archives/C066M97D679/p1700762457816329

### DIFF
--- a/_toc.yml
+++ b/_toc.yml
@@ -433,6 +433,8 @@ entries:
             title: Advanced parameters
           - file: docs/products/kafka/reference/kafka-metrics-prometheus
             title: Metrics available via Prometheus
+          - file: docs/products/kafka/reference/kstream-data-write-issue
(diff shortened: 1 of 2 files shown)


.github/vale/dicts/aiven.dic
```diff
@@ -143,6 +143,7 @@ keystores
 Kibana
 Klaw
 Kpow
+KafkaStreams
```
aiven-amartin, line 146: Seems the existing standard is to write `Apache Kafka® Streams`

cybermax-008: Thanks for finding this out. Replaced KafkaStreams with  Apache Kafka® Streams

aiven-amartin (approved):

PR #2288, opened 2023-11-24
fix pipes

ArthurFlageul:

Remove unwanted pipe characters


harshini-rangaswamy (approved):
Looks good!

PR #2294, opened 2023-11-27
mysql: add prerequisite for ca.pem permissions while connecting to mysql with php

wojcik-dorota:
https://aiven.atlassian.net/browse/DOC-650

### DIFF
--- a/code/products/mysql/connect.php
+++ b/code/products/mysql/connect.php
@@ -9,7 +9,7 @@ $conn = "mysql:";
 $conn .= "host=" . $fields["host"];
 $conn .= ";port=" . $fields["port"];;
 $conn .= ";dbname=defaultdb";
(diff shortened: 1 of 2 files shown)


EddyNhara:
Hi @wojcik-dorota , I think the line after checking for read/write permissions should read something like, "If you have permissions". This would mean you have permissions to the file but cannot locate the file hence the need for the absolute path. Hope this makes sense

wojcik-dorota:
> Hi @wojcik-dorota , I think the line after checking for read/write permissions should read something like, "If you have permissions". This would mean you have permissions to the file but cannot locate the file hence the need for the absolute path. Hope this makes sense

Thanks for the comment :-) Not sure what changes are expected though. Please propose the wording for this prerequisite.

PR #2311, opened 2023-11-29
clickhouse: add limitation for scaling down with sharding

wojcik-dorota:
https://aiven.atlassian.net/browse/DOC-656

### DIFF
--- a/docs/products/clickhouse/reference/limitations.rst
+++ b/docs/products/clickhouse/reference/limitations.rst
@@ -52,9 +52,9 @@ From the information about restrictions on using Aiven for ClickHouse, you can e
    * - Creating a database using SQL
      - You cannot create a database directly using SQL, for example, if you'd like to add a non-default database.
      - Use the Aiven's public API.
-   * - Scaling down a sharded cluster
-     - You only can scale up between shards. Scaling down is not supported with the sharding strategy. You cannot downgrade to a service plan with fewer VMs if your ClickHouse cluster is sharded.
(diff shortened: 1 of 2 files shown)


docs/products/clickhouse/reference/limitations.rst
```diff
@@ -52,6 +52,9 @@ From the information about restrictions on using Aiven for ClickHouse, you can e
    * - Creating a database using SQL
      - You cannot create a database directly using SQL, for example, if you'd like to add a non-default database.
      - Use the Aiven's public API.
+   * - Scaling down a sharded cluster
+     - You only can scale up between shards. Scaling down is not supported with the sharding strategy. You cannot downgrade to a service plan with fewer VMs if your ClickHouse cluster is sharded.
```
alanfranz, line 56: You can "scale up" or "scale down" what, exactly?
I just checked, and it _is_possible to scale down storage, cpu and RAM, in a sharded cluster, if disk space allows that.

The only thing that is unsupported is scaling down the number of nodes. I don't know if this is supported for unshared clusters by the way, because I wasn't able to find plans with more than 3 nodes and just 1 shard, so the point is kind of "undecidable" right now.

alanfranz (dismissed):
After a chat: https://aiven-io.slack.com/archives/C052RDBD1TP/p1702303328664239

I'd say things in a simpler fashion: scaling down the amount of nodes is not supported.

So customers should pick their initial plan, especially the number of shards, with a bit of care.

alanfranz (approved):

PR #2312, opened 2023-11-29
clickhouse: add ref doc for supported protocols and drivers

wojcik-dorota:
https://aiven.atlassian.net/browse/DOC-655

### DIFF
--- a/docs/products/clickhouse/reference/supported-interfaces-drivers.rst
+++ b/docs/products/clickhouse/reference/supported-interfaces-drivers.rst
@@ -1,0 +1,33 @@
+Interfaces and drivers supported in Aiven for ClickHouse®
+=====================================================================
+
+Find out what technologies and tools you can use to interact with Aiven for ClickHouse®.
+
+.. _clickhouse-interfaces:
+
+Interfaces (protocols)
+----------------------
+
+Aiven for ClickHouse® supports the following fundamental underlying interfaces (protocols):
+
+* ``HTTPS``
+* ``Native TCP``
+* ``MySQL Interface``
+
+.. topic:: Full list of interfaces (protocols) supported in Aiven for ClickHouse
+   
+   Check out `Drivers and Interfaces <https://clickhouse.com/docs/en/interfaces/overview>`_ for all the details.
+
+Drivers (libraries)
+-------------------
+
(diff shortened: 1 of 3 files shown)


docs/products/clickhouse/reference/supported-interfaces-drivers.rst
```diff
@@ -0,0 +1,33 @@
...
+=====================================================================
+
+Find out what technologies and tools you can use to interact with Aiven for ClickHouse®.
+
+.. _clickhouse-interfaces:
+
+Interfaces (protocols)
+----------------------
+
+Aiven for ClickHouse® supports the following fundamental underlying interfaces (protocols):
```
joelynch, line 11: Maybe we could add something here about how Aiven does not support connecting to ClickHouse without TLS (ie plaintext) for security reasons (this is a deliberate choice to protect the customer from doing something dangerous and unsecure).

joelynch: And also we do not support gRPC right now, might be worth explicitly saying this.  I believe it is good to be more upfront about unsupported Clickhouse things.

docs/products/clickhouse/reference/supported-interfaces-drivers.rst
```diff
@@ -0,0 +1,33 @@
...
+* ``MySQL Interface``
+
+.. topic:: Full list of interfaces (protocols) supported in Aiven for ClickHouse
+   
+   Check out `Drivers and Interfaces <https://clickhouse.com/docs/en/interfaces/overview>`_ for all the details.
+
+Drivers (libraries)
+-------------------
+
+There are a number of drivers (libraries) that use one of :ref:`the fundamental underlying interfaces supported in Aiven for ClickHouse <clickhouse-interfaces>` under the hood. It's up to you to pick up a driver (library) of your choice and use it for connecting to your Aiven for ClickHouse service.
```
joelynch, line 24: We have a bunch of guides for using different drivers here https://docs.aiven.io/docs/products/clickhouse/howto/list-connect-to-service.  Might be worth linking to this for information about drivers and libraries, especially from an Aiven perspective.  We can then link to the ClickHouse docs and say something like "for the full list of known drivers and libraries for connecting to ClickHouse" or something.

We can also say that any driver that uses one of the supported protocols, and uses TLS, can be used.

joelynch (changes requested):
Looking pretty good, few comments to improve things a bit.

joelynch (approved):
Looks good

PR #2338, opened 2023-12-08
Add redirect to aiven.io/community

FloorD:
Adding a direct link to our community page on aiven.io/community, instead of maintaining multiple pages.


_redirects
```diff
@@ -13,6 +13,7 @@
 /api                /docs/tools/api
 /cli                /docs/tools/cli.html
 /terraform          /docs/tools/terraform
+/community          https://aiven.io/community/
```
ArthurFlageul, line 16: Hi Floor 👋 

If we're gonna have a redirect, let's delete the page altogether that way we do some winter cleaning 😊  Could you delete the page and update the links to the old location to the new one if you don't mind?

ArthurFlageul (changes requested):

ArthurFlageul (approved):

PR #2344, opened 2023-12-11
platform: add CIDR guidelines to the BYOC docs

wojcik-dorota:
https://aiven.atlassian.net/browse/DOC-671

### DIFF
--- a/docs/platform/howto/byoc/create-custom-cloud.rst
+++ b/docs/platform/howto/byoc/create-custom-cloud.rst
@@ -472,7 +472,7 @@ In the **Create custom cloud** wizard, proceed as follows:
    * Region
    * CIDR
 
-     **CIDR** represents an IP address range of the BYOC VPC. It needs to be specified in the IaC template so that a `CIDR block of the VPC <https://docs.aws.amazon.com/vpc/latest/userguide/vpc-cidr-blocks.html>`_ can be created in your AWS account while deploying the template. Check `the AWS documentation on VPC CIDR blocks <https://docs.aws.amazon.com/vpc/latest/userguide/vpc-cidr-blocks.html>`_ for more information.
+     **CIDR** represents an IP address range of the BYOC VPC. It needs to be specified in the IaC template so that a CIDR block of the VPC can be created in your AWS account while deploying the template.
 
      In the **CIDR** field, specify an IP address range for the BYOC VPC using a CIDR block notation, for example: ``10.0.0.0/16``, ``172.31.0.0/16``, or ``192.168.0.0/20``.
      
--- a/docs/platform/howto/byoc/create-custom-cloud.rst
+++ b/docs/platform/howto/byoc/create-custom-cloud.rst
@@ -478,7 +478,7 @@ In the **Create custom cloud** wizard, proceed as follows:
      
      Make sure an IP address range you use meets the following requirements:
      * IP address range is within the private IP address ranges allowed in `RFC 1918 <https://datatracker.ietf.org/doc/html/rfc1918>`_.
-     * CIDR block size doesn't excced the ``/24`` netmask.
+     * CIDR block size doesn't exceed the ``/24`` netmask.
 
      .. note::
 
--- a/docs/platform/howto/byoc/create-custom-cloud.rst
+++ b/docs/platform/howto/byoc/create-custom-cloud.rst
@@ -484,6 +484,9 @@ In the **Create custom cloud** wizard, proceed as follows:
      .. note::
 
         * CIDR block needs be large enough so that, after splitting it into per-region subnets, each subnet has enough addresses to fit required services.
+
+          For example, using the maximum ``/24`` prefix length, you should have up to 16 IP addresses available, which might be enough for a few services, but can pose challenges during node replacements or maintenance upgrades, when you may need extra addresses.
+
         * Make sure the CIDR block of your BYOC VCP doesn't overlap with the CIDR blocks of VPCs you plan to peer your BYOC VPC with. You cannot change the BYOC VPC CIDR block after your custom cloud is created.
 
 2. Select **Next**.
--- a/docs/platform/howto/byoc/create-custom-cloud.rst
+++ b/docs/platform/howto/byoc/create-custom-cloud.rst
@@ -472,22 +472,19 @@ In the **Create custom cloud** wizard, proceed as follows:
    * Region
    * CIDR
 
-     The **CIDR** block defines the IP address range of the VPC that Aiven creates in your AWS account. Any Aiven service created in your Aiven organization will be placed in the VPC and will get an IP address within this address range.
+     The **CIDR** block defines the IP address range of the VPC that Aiven creates in your AWS account. Any Aiven service created in the custom cloud will be placed in the VPC and will get an IP address within this address range.
 
      In the **CIDR** field, specify an IP address range for the BYOC VPC using a CIDR block notation, for example: ``10.0.0.0/16``, ``172.31.0.0/16``, or ``192.168.0.0/20``.
      
-     Make sure an IP address range you use meets the following requirements:
+     Make sure that an IP address range you use meets the following requirements:
 
      * IP address range is within the private IP address ranges allowed in `RFC 1918 <https://datatracker.ietf.org/doc/html/rfc1918>`_.
      * CIDR block size is between ``/16`` (65536 IP addresses) and ``/24`` (256 IP addresses).
+     * CIDR block is large enough to host the desired number of services after splitting it into per-availability-zone subnets.
 
-     .. note::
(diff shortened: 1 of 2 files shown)


docs/platform/howto/byoc/create-custom-cloud.rst
```diff
@@ -472,11 +472,19 @@ In the **Create custom cloud** wizard, proceed as follows:
...
 
-     * Specify inbound rules with the CIDR block notation, for example: 200.1.2.3/32 (allowing 200.1.2.3 as a single address), 0.0.0.0/0 (allowing traffic from anywhere), or 100.1.0.0/16 (allowing traffic from 100.1..).
-     * To create VPC peerings with that VPC, choose a CIDR block that doesn't overlap with CIDR blocks of peer VPCs.
-     * Keep in mind that CIDR block needs be large enough so that, after splitting it into per-region subnets, each subnet has enough addresses to fit required services.
+     In the **CIDR** field, specify an IP address range for the BYOC VPC using a CIDR block notation, for example: ``10.0.0.0/16``, ``172.31.0.0/16``, or ``192.168.0.0/20``.
+     
+     Make sure an IP address range you use meets the following requirements:
+
+     * IP address range is within the private IP address ranges allowed in `RFC 1918 <https://datatracker.ietf.org/doc/html/rfc1918>`_.
+     * CIDR block size doesn't exceed the ``/24`` netmask.
```
heikju, line 482: This could be written in a more clear way I think. In that network notation the `/24` part is called a "prefix length", i.e. the number of leading set bits in the netmask, and the larger the prefix length, the smaller the network.

docs/platform/howto/byoc/create-custom-cloud.rst
```diff
@@ -472,11 +472,19 @@ In the **Create custom cloud** wizard, proceed as follows:
...
+
+     .. note::
+
+        * CIDR block needs be large enough so that, after splitting it into per-region subnets, each subnet has enough addresses to fit required services.
```
heikju, line 486: I think we could try to give the reader something a bit more concrete here. If we go with the maximum prefix length of 24, then that leaves 8 bits for the network addresses. When this space is split between bastion and workload networks, and then between availability zones (iI think there's at least one AWS with 6 AZs, so we can take 1/8 for each in that case), eventually a single workload az subnet isn't guaranteed more than 1/16 of the total, i.e. 16 addresses with the max. prefix lenght. This is enough for several services, but could run out. And during node replacements or maintenance upgrades we could need some extra addresses.

Do you know if our public docs contain any advice on sizing project VPCs? I didn't find any with a brief search, but if there's some, then that could be applicable here too with small adjustments.

docs/platform/howto/byoc/create-custom-cloud.rst
```diff
@@ -472,11 +472,19 @@ In the **Create custom cloud** wizard, proceed as follows:
    * Region
    * CIDR
 
-     Aiven needs CIDR for the `CIDR block of the VPC <https://docs.aws.amazon.com/vpc/latest/userguide/vpc-cidr-blocks.html>`_ that will be created in your AWS account.
+     **CIDR** represents an IP address range of the BYOC VPC. It needs to be specified in the IaC template so that a CIDR block of the VPC can be created in your AWS account while deploying the template.
```
heikju, line 475: I think this could be phrased better: user gives us a single CIDR block via our API, and then we parametrize the template to create a VPC with that CIDR block, and then subnets in that VPC. "CIDR block of the VPC can be created" is not that clear to me.

docs/platform/howto/byoc/create-custom-cloud.rst
```diff
@@ -472,11 +472,19 @@ In the **Create custom cloud** wizard, proceed as follows:
...
+     
+     Make sure an IP address range you use meets the following requirements:
+
+     * IP address range is within the private IP address ranges allowed in `RFC 1918 <https://datatracker.ietf.org/doc/html/rfc1918>`_.
+     * CIDR block size doesn't exceed the ``/24`` netmask.
+
+     .. note::
+
+        * CIDR block needs be large enough so that, after splitting it into per-region subnets, each subnet has enough addresses to fit required services.
+        * If creating VPC peerings, make sure a CIDR block of your VCP doesn't overlap with CIDR blocks of peer VPCs.
```
heikju, line 487: It could be worthwhile to emphasize that the CIDR block can't be changed after a BYOC cloud has been provisioned. AWS doesn't support resizing VPCs in a simple way.

heikju (changes requested):

docs/platform/howto/byoc/create-custom-cloud.rst
kaikontio, line 475:
```suggestion
     This CIDR block defines the IP address range of the VPC Aiven creates in your AWS account. Any Aiven services created in your account will be placed in the VPC and get an IP address within this address range.
```

docs/platform/howto/byoc/create-custom-cloud.rst
kaikontio, line 482:
```suggestion
     * The CIDR block size is between ``/16`` (65536 IP addresses) and ``/24`` (256 IP addresses).
```

docs/platform/howto/byoc/create-custom-cloud.rst
kaikontio: Agree that this is a bit detached from the reality of running services. We could give an example of how many services the smallest /24 block would accommodate, for example.

docs/platform/howto/byoc/create-custom-cloud.rst
kaikontio, line 487:
```suggestion
        * Make sure the CIDR block of your BYOC VCP doesn't overlap with the VPCs you plan to peer the BYOC VPC with. The BYOC VPC CIDR block cannot be changed after creation.
```

docs/platform/howto/byoc/create-custom-cloud.rst
```diff
@@ -472,11 +472,22 @@ In the **Create custom cloud** wizard, proceed as follows:
    * Region
    * CIDR
 
-     Aiven needs CIDR for the `CIDR block of the VPC <https://docs.aws.amazon.com/vpc/latest/userguide/vpc-cidr-blocks.html>`_ that will be created in your AWS account.
+     The **CIDR** block defines the IP address range of the VPC that Aiven creates in your AWS account. Any Aiven service created in your Aiven organization will be placed in the VPC and will get an IP address within this address range.
```
kaikontio, line 475: This applies only to the custom cloud in question. On the org level other deployment targets can be used as well.

```suggestion
     The **CIDR** block defines the IP address range of the VPC that Aiven creates in your AWS account. Any Aiven service created in the custom cloud will be placed in the VPC and will get an IP address within this address range.
```

docs/platform/howto/byoc/create-custom-cloud.rst
```diff
@@ -472,11 +472,22 @@ In the **Create custom cloud** wizard, proceed as follows:
...
+     Make sure an IP address range you use meets the following requirements:
+
+     * IP address range is within the private IP address ranges allowed in `RFC 1918 <https://datatracker.ietf.org/doc/html/rfc1918>`_.
+     * CIDR block size is between ``/16`` (65536 IP addresses) and ``/24`` (256 IP addresses).
+
+     .. note::
+
+        * CIDR block needs be large enough so that, after splitting it into per-region subnets, each subnet has enough addresses to fit required services.
+
+          For example, using the maximum ``/24`` prefix length, you should have up to 16 IP addresses available, which might be enough for a few services, but can pose challenges during node replacements or maintenance upgrades, when you may need extra addresses.
```
kaikontio, line 488:
```suggestion
          For example, the smallest ``/24`` CIDR block might be enough for a few services, but can pose challenges during node replacements or maintenance upgrades, if running low on available free IP addresses.
```

docs/platform/howto/byoc/create-custom-cloud.rst
```diff
@@ -472,11 +472,22 @@ In the **Create custom cloud** wizard, proceed as follows:
...
+     In the **CIDR** field, specify an IP address range for the BYOC VPC using a CIDR block notation, for example: ``10.0.0.0/16``, ``172.31.0.0/16``, or ``192.168.0.0/20``.
+     
+     Make sure an IP address range you use meets the following requirements:
+
+     * IP address range is within the private IP address ranges allowed in `RFC 1918 <https://datatracker.ietf.org/doc/html/rfc1918>`_.
+     * CIDR block size is between ``/16`` (65536 IP addresses) and ``/24`` (256 IP addresses).
+
+     .. note::
+
+        * CIDR block needs be large enough so that, after splitting it into per-region subnets, each subnet has enough addresses to fit required services.
```
kaikontio, line 486:
```suggestion
        * CIDR block is large enough to host the desired number of services, after splitting it into per availability zone subnets.
```

kaikontio (approved):
Content-wise LGTM.

PR #2383, opened 2023-12-19
platform: vpc peering connection time interval for implementing user changes

wojcik-dorota:
https://aiven.atlassian.net/browse/DOC-706

### DIFF
--- a/docs/platform/howto/manage-vpc-peering.rst
+++ b/docs/platform/howto/manage-vpc-peering.rst
@@ -85,7 +85,7 @@ Troubleshoot VPC connection issues
 
 Any network changes to VPC peered hosts external from Aiven can cause issues with routing to your Aiven services hosted in a VPC. To troubleshoot such issues, :ref:`refresh VPC connections <refresh-vcp-connections>`.
 
-Changes to your VPCs (such as adding a new subent) can take up to 24 hours to take effect. There's no need to troubleshoot issues with your changes not applied unless it's been over 24 hours since you introduced the changes. You can always force immediate  application of your changes by :ref:`refreshing VPC connections <refresh-vcp-connections>`.
+Changes to your VPCs (such as adding a new subnet) can take up to 24 hours to take effect. There's no need to troubleshoot issues with your changes not applied unless it's been over 24 hours since you introduced the changes. You can always request the application of your changes by :ref:`refreshing VPC connections <refresh-vcp-connections>`.
 
 .. _refresh-vcp-connections:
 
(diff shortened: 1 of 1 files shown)


docs/platform/howto/manage-vpc-peering.rst
```diff
@@ -81,9 +81,15 @@ IP filtering (the **Allowed IP addresses** list on the service's **Overview** pa
 Also note that safelisting applies to both internal and external traffic. If you safelist an external IP address and want to keep traffic flowing with the internal (peered) connections, make sure that you safelist the CIDR blocks of the peered networks as well to avoid disruptions to the service.
 
 Troubleshoot VPC connection issues
--------------------------------------
+----------------------------------
 
-Any network changes to VPC peered hosts external from Aiven can cause issues with routing to your Aiven services hosted in a VPC. To troubleshoot such issues, take the following steps:
+Any network changes to VPC peered hosts external from Aiven can cause issues with routing to your Aiven services hosted in a VPC. To troubleshoot such issues, :ref:`refresh VPC connections <refresh-vcp-connections>`.
+
+Changes to your VPCs (such as adding a new subent) can take up to 24 hours to take effect. There's no need to troubleshoot issues with your changes not applied unless it's been over 24 hours since you introduced the changes. You can always force immediate  application of your changes by :ref:`refreshing VPC connections <refresh-vcp-connections>`.
```
ali-gholami-aiven, line 88:
```suggestion
Changes to your VPCs (such as adding a new subnet) can take up to 24 hours to take effect. There's no need to troubleshoot issues with your changes not applied unless it's been over 24 hours since you introduced the changes. You can always force immediate  application of your changes by :ref:`refreshing VPC connections <refresh-vcp-connections>`.
```

ali-gholami-aiven (commented):
Looks good, there's just a tiny typo.

docs/platform/howto/manage-vpc-peering.rst
giacomo-alzetta-aiven, line 88: I would word it like this:

```suggestion
Changes to your VPCs (such as adding a new subent) can take up to 24 hours to take effect. There's no need to troubleshoot issues with your changes not applied unless it's been over 24 hours since you introduced the changes. You can always manually request application of your changes by :ref:`refreshing VPC connections <refresh-vcp-connections>`.
```

The request button schedules a request to refresh the VPC connection, but a few minutes may pass before this is picked up.

Also, as far as I know, there are still some limits on how often the refresh is carried out. So if the presses the button, for example, every minute for an hour, it won't necessarily translate in refreshing the actual state and cloud resources 60 times... likely only a few times.

giacomo-alzetta-aiven: I wouldn't go too much into the specifics on exactly when/if the connections are refreshed, because it is actually complicated and IMHO provide little value and only increases the chances that the documentation stays behind compared to the actual implementation which is going to evolve at some point.

giacomo-alzetta-aiven (dismissed):

giacomo-alzetta-aiven (approved):
LGTM!

PR #2405, opened 2023-12-29
Fix Datadog integration update example

juha-aiven:
The values need to in quotes.


harshini-rangaswamy (approved):
Thanks for fixing this.

PR #2410, opened 2024-01-03
platfrom: byoc standard architecture update to show one vpc

wojcik-dorota:
https://aiven.atlassian.net/browse/DOC-712


docs/platform/concepts/byoc.rst
```diff
@@ -54,16 +54,15 @@ You receive two separate monthly invoices, one from Aiven for their managed serv
...
-Aiven services you deploy. As the user of these services (for example, Aiven for Apache Kafka®), you are able to utilize them through standard VPC peering techniques. Although the bastion host and the service nodes reside in your managed VPC, they are not accessible (for example, SSH) to anyone outside Aiven.
+The standard BYOC deployment requires you to create a Virtual Private Cloud (VPC) dedicated to Aiven services within a cloud region you want to operate in. Aiven accesses this VPC via a static IP address and routes traffic through a proxy for additional security. To accomplish this, Aiven utilizes a bastion host logically separated from the Aiven services you deploy. You are able to run your services using standard VPC peering techniques. Although the bastion host and the service nodes reside in the VPC under your management (**Aiven service VPC**), they are not accessible (for example, via SSH) to anyone outside Aiven.
```
akudiyar, line 65: > Aiven accesses this VPC via a static IP address

via -> from?

> Aiven utilizes a bastion host logically separated

As it is a VM in s different subnet, can we talk here about it being _physically_ separated?

> You are able to run your services

Maybe "you are able to integrate your services". The old phrasing here seems to be more accurate 

Should we write here also that the service nodes connect to the Aiven services and so they need an internet access via NAT, while the bastion service utilizes an internet gateway?

wojcik-dorota: @akudiyar , thank you :-) Implemented all bit for the last one. Since we decided to make it a simplified view and removed the outbound from the diagram (both NAT GW and Internet GW not visible), I'd say we better leave them out from the description for now. We can cover that as soon as we have move detailed tech architecture in place. Hope that makes sense?

bugant (commented):
This looks good to me.  I'd anyway wait for reviews from @heikju and @akudiyar.

bugant (approved):

heikju (approved):

ArthurFlageul (approved):

